PR #690, opened 2026-05-20
feat: Phase 5 — SyncToken-gated UPSERT + CDC watermark + payload drift

jfrench9:
## Summary

Ships Phase 5 of `quickbooks-adapter.md` §4.3 — SyncToken capture + freshness-gated UPSERT, `last_cdc_watermark` column + advance logic, plus the CDC API client surface. Closes the last big in-spec QB backend item. The adapter is now robust to the failure modes webhook-style delivery would expose us to (replay, out-of-order, drift).

## Scope vs spec §4.3.x

- **§4.3.1 SyncToken capture** — flatten functions stamp `sync_token` from QB's `SyncToken` field; dbt staging + marts surface it; loader persists to `Event.metadata_['qb_sync_token']` + `Agent.metadata_['qb_sync_token']`.
- **§4.3.2 SyncToken-gated UPSERT** — `_compare_sync_tokens()` returns `fresh / same / stale / no_info`; gate runs before the status branches so stale + same-version incoming rows skip the entire UPSERT path. Committed events advance the live SyncToken even without business-payload drift (bookkeeping).
- **§4.3.3 payload_drift** — already shipped as Wave 1 G4 (reused, drift comparison updated to exclude `qb_sync_token`).
- **§4.3.4 `last_cdc_watermark` column + advance logic** — platform migration `d528e24fb069`; `Connection.advance_cdc_watermark()` helper; `_run_qb_load` captures `sync_started_at` at load start, advances on success.
- **§4.3.4 CDC client surface** — `QBClient.cdc(changed_since, entities)` returns `(entities_by_type, watermark_too_old)`; tight 30-day-cap detection (only the specific changedSince fault flips watermark fallback; all other 400s propagate).
- **§4.3.5 Idempotent re-ingest harness** — 4 scenarios covering identical re-sync no-op, out-of-order stale skip, committed-event drift with immutable live payload, and the full insert→resync snapshot stability.

## Deferred (NOT shipped here, documented in spec)

**The CDC endpoint is wired but not yet called during sync.** `extract.py` still uses the per-entity fetch + JournalReport pipeline. The blocker is architectural: the current GL pipeline depends on JournalReport, which can't be CDC-fetched (it's a report, not entities); switching to per-entity `Line[]` arrays is the clean path but pairs with §4.6.5 `stg_qb_journal_entries` retire. That deserves its own focused PR. Spec §4.3.4 now carries a status note explaining the gap.

Also deferred: handler dispatch for QB-side deletions (CDC returns `status='Deleted'` records). The `cdc()` docstring flags this; until extract.py is wired, it doesn't matter operationally.

## Live verification

Verified end-to-end against the Harbinger production graph (`kg19e3d7cc55c95c0edbcc`, 7-year / 1,570-event QB tenant). Two consecutive 60-day incremental syncs:

- `last_cdc_watermark`: NULL → `2026-05-20 04:35:38` → `2026-05-20 04:40:51` (advances on each successful load)
- Agent SyncToken coverage: **86 / 86** (100%)
- Header-bearing event coverage: **8 / 8** (Bills + BillPayments; JournalReport-only rows stay NULL per documented limitation)
- One drift flagged on `Bill_1615` from prior QB sandbox edit; flag stable across the idempotent second sync (no false re-flag)
- Per-event sync_token values **byte-for-byte identical** across both syncs — live proof the gate short-circuits "same" decisions before UPSERT

## Test plan

- [x] `just test-code` clean (ruff + ruff format + basedpyright + cf-lint)
- [x] 299 tests pass across QB adapter + extensions + Connection model
- [x] 16 new tests added: 3 sync_token agent capture; 3 sync_token event capture; 6 gate decisions + helper; 3 `Connection.advance_cdc_watermark`; 6 CDC client (URL routing × 2, parsing, 30-day cap, unrelated-400 propagates, auth headers); 4 idempotent re-ingest scenarios; 1 `qb_load` asset coverage
- [x] Migration `d528e24fb069` applied locally (`just migrate-up`)
- [x] Two consecutive live syncs against Harbinger graph confirm gate + watermark behavior in production data

## Breaking changes

- **Platform DB migration `d528e24fb069`** must be applied before deployment (`alembic upgrade head` against platform DB). The column is nullable so existing connections are unaffected until their next sync advances the watermark.


jfrench9:
@claude please review this PR

PR #707, opened 2026-05-29
Canonical RDF ontology, graph-native v2 bundles & SHACL validation

jfrench9:
## Summary

This PR introduces a canonical RDF ontology for the taxonomy, migrates bundles to a graph-native v2 serialization format, and adds opt-in SHACL validation at publish time. It also consolidates validation tooling across all demo examples and adds statement-level reconciliation capabilities.

### Canonical RDF Ontology (`frameworks/ontology/v1/`)
- **`ontology.ttl`** — Formal OWL/RDF ontology that reifies XBRL arcs and `xbrli` concept attributes as first-class RDF properties and classes.
- **`shapes.ttl`** — SHACL shapes graph for validating bundle conformance against the ontology.
- **`context.jsonld`** — Shared JSON-LD context enabling compact, human-readable serialization aligned with the ontology.

### Graph-Native v2 Bundle Serialization
- XBRL dimensional context (entity, period, units, explicit/typed dimensions) is collapsed directly onto fact nodes, eliminating the separate context-to-fact indirection and producing a cleaner RDF graph.
- Bundle ontology version stamp corrected to **v1** (was incorrectly set to v2).
- All taxonomy packages (FAC, RS-GAAP, bridges) updated to use reified arc representations and the new JSON-LD context, resulting in large but mechanical diffs across ~20 taxonomy files.

### SHACL Validation at Publish
- New opt-in SHACL validation step integrated into the bundle publish pipeline; results are captured on the `Report` object.
- Dedicated test suites validate SHACL conformance of sample bundles and verify the publish-time validation hook.

### Consolidated Validation Tooling
- New shared `examples/_common/validate.py` module provides container-free SHACL and Arelle (XBRL 2.1) validation, replacing the per-demo `xbrl_validate.py` script.
- All three demos (Roboledger, Seattle Method, World Online) updated to use the common validator and now emit both SHACL and XBRL validation reports as sample outputs.

### Statement-Level Reconciliation (World Online)
- New `statement_reconcile.py` performs line-item reconciliation of generated financial statements against Charlie's reference data.
- Reconciliation output captured as a new sample artifact (`world-online-statement-reconciliation.md`).

## Breaking Changes

- **Bundle JSON-LD structure changed**: Consumers parsing the `.jsonld` output will see a different graph shape — dimensional context fields now appear inline on fact nodes rather than as separate context objects. Any downstream tooling that relies on the v1 context/fact indirection will need to be updated.
[...]

PR #708, opened 2026-05-30
Clean up XBRL serialization output and add bundle support

jfrench9:
## Summary

Follow-on to #707, addressing the XBRL-validity issues Charlie Hoffman flagged on the first shipped serialization (validated against XBRL Cloud + Instar). All fixes are framework-independent — rs-gaap stays the report framework. The canonical RDF ontology is the interop layer; XBRL 2.1 is the bridge (mini-as-reporting-framework is explicitly out of scope).

## Changes

All four changed source/test files are **modifications** to existing modules (no new module or runtime dependency):

- **`xbrl/xbrl_21.py`** — fact dedup; label linkbase (`report-lab.xml`); conditional emission of every linkbase file (presentation/calc/def/label all gated on having content); `NS_XML` constant.
- **`serialization/bundle.py`** — `_source_calculation_arcs`: sources calc arcs onto the rendered Network's ELR and emits a subtotal only when every child's stored weight sign is XBRL-legal for the endpoints' balance types (§5.1.1.2 Table 6).
- **`rdf/jsonld.py`** — scope the Association IRI by link group so a presentation arc and a sourced calc arc at the same index under one Structure don't collapse onto one node (which failed `AssociationShape`).
- **`tests/`** — `test_xbrl_emitter.py` (dedup, labels, conditional linkbases) + new `test_calc_sourcing.py` (the balance-legality heuristic).

## What it fixes (Charlie's review)

- **Duplicate facts** — dedupe the instance on the full `(concept, contextRef, unitRef, value, decimals)` tuple. The same value legitimately flows through several FactSets (NetIncomeLoss via IS/CF/SE); deduping on the full tuple removes the redundant copies while still surfacing a genuine inconsistency (same aspects, different value) to the validator.
- **No labels** — emit `report-lab.xml` from the element display name (standard label role, `xml:lang=en`); concepts whose only text echoes the QName local part stay unlabelled.
- **No calculation relations** — calc arcs lived on separate rs-gaap-calculation Structures and never reached the bundle. The legality filter drops the cash-flow rollups (indirect-method children mix debit/credit under a cash subtotal — the canonical case real filers omit) and keeps the well-behaved BS/IS summations.
- **Empty linkbase noise** — suppress empty calc/def linkbase files (the linkbaseRefs were already gated).

## Validation

End-to-end on all three demos (lemonade, World Online, roboledger): **0 duplicate facts, 62 calculation arcs, 75 labels, SHACL conforms, Arelle valid XBRL 2.1**. `sample_output/` refreshed. Full serialization suite (122 tests) green; ruff/format/basedpyright clean.

## Out of scope (deliberate)

- Mini-as-reporting-framework output — would regress the CoA→rs-gaap projection story.
- Luca/DFRNT Excel/JSON import format — proprietary vendor ingestion shape; XBRL 2.1 is already the interop layer.


jfrench9:
@claude please review this PR

PR #720, opened 2026-06-03
chore(billing): raise Standard tier to $149 and drop LadybugDB from tier names

jfrench9:
## Summary

Two billing changes, isolated to config + presentation layers:

1. **Standard tier price: $99 → $149/mo.**
2. **Drop the "LadybugDB" brand from customer-facing tier names** — display names become `Standard` / `Large` / `XLarge`. The tenant is branded as a "graph"; LadybugDB stays as engine attribution on landing/marketing only, never the leading subscription name.

## Changes

- `config/billing/core.py` — Standard `base_price_cents` 9900 → 14900; tier `display_name`s drop "LadybugDB"; Standard `description` drops "LadybugDB"; `get_plan_display_name` docstring example updated.
- `config/graph_tier.py` — tier `display_names` map drops "LadybugDB" (the SEC "Shared Repository" label is unchanged).
- `models/api/graphs/subgraphs.py` — field description now reads "requires Large/XLarge tier".
- `dagster/jobs/invoice_billing.py` + `operations/graph/provisioning_service.py` — invoice line-item descriptions resolve the human display name with a "Graph" context (e.g. `Graph subscription - Standard`) instead of rendering the raw internal id (`ladybug-standard`); falls back to the plan name if no config match.
- `tests/config/billing/test_core.py` — price assertion updated to 14900.

**Internal tier IDs (`ladybug-standard`, etc.) and the `GraphTier` enum are unchanged — this is verbiage-only.** The frontend subscription UI renders `display_name`/`plan_display_name` from the API, so it updates with no frontend change.

## Pricing behavior — not a breaking change for existing customers

- **Existing subscribers are grandfathered automatically.** Invoicing reads `subscription.base_price_cents`, which is snapshotted at subscription-creation time — so existing Standard subscribers keep their current rate. Only **new** subscriptions snapshot the $149 config price.
- **No Stripe action required.** There is no graph product/price in Stripe today; the existing graph customer is billed via the platform's manual-invoice path (not a Stripe subscription). A new $149 Stripe price would only be created the first time a graph checkout ever runs through Stripe.

## Testing

- `just test-all`: **10406 passed**.
- Billing config, subscriptions router (incl. display-name mocks), and invoice tests pass; CodeQL clean.

## Follow-ups (non-blocking)

- Unit test for the new invoice-description format + the unknown-plan fallback.
- Consolidate the tier display-name mapping that currently lives in both `core.py` and `graph_tier.py` (single source of truth).

### DIFF
--- a/robosystems/config/billing/core.py
+++ b/robosystems/config/billing/core.py
@@ -203,7 +203,7 @@ class BillingConfig:
   ) -> str:
     """Resolve a plan's internal name to its human-readable display name.
 
-    For graphs: "ladybug-standard" -> "LadybugDB Standard"
+    For graphs: "ladybug-standard" -> "Standard"
     For repos:  "sec-advanced"     -> "SEC EDGAR Filings - Pro"
 
     Falls back to the raw plan name when no display mapping is found.
--- a/robosystems/config/billing/core.py
+++ b/robosystems/config/billing/core.py
@@ -33,7 +33,7 @@ DEFAULT_GRAPH_BILLING_PLANS: list[dict[str, Any]] = [
     "name": "ladybug-standard",
     "display_name": "LadybugDB Standard",
     "description": "Dedicated m7g.large LadybugDB infrastructure with subgraph support",
-    "base_price_cents": 9900,  # $99/month
+    "base_price_cents": 14900,  # $149/month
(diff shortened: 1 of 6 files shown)


jfrench9:
@claude please review this PR

PR #731, opened 2026-06-06
feat(taxonomy): fac bridge coverage + CF/recurrence trait axes

jfrench9:
## Summary

Completes a set of pre-launch fixes from a taxonomy framework review, done in the pre-tenant window where the seeded library is still freely rebuildable. It hardens fac-trait coverage on the tenant-shipped rs-gaap working set, adds a new analytical trait axis, and completes `fac` as a cross-framework anchor over the MVP working set. All changes are either tenant-additive (recoverable via resync) or library-only.

## Changes

**Trait consistency — `rs-gaap-traits`**
- Bind `indirectCashFlowReconcilingItem` to the 13 kept `IncreaseDecreaseIn*` working-capital concepts (consistency with their excluded siblings; classification-only — verified it is not consumed by the renderer, so no behavior change).
- Bind `use/remeasurement` to the 3 kept impairment-bearing concepts (matches the existing impairment pattern, e.g. `TangibleAssetImpairmentCharges`).

**New `recurrence` trait axis — `fac-traits` + model + migration**
- Add a fac-traits axis `recurrence/nonrecurring` (single-value flag) for normalized-earnings / NOPAT classification, bound to the 9 kept operating special-items (restructuring x4, impairment, disposal gains/losses x3, debt extinguishment). Orthogonal to `use/remeasurement` so both "NOPAT ex-remeasurements" and "NOPAT ex-all-special-items" cuts are expressible.
- Widen `check_trait_category` in `robosystems/models/extensions/trait.py` and in `migrations/extensions/versions/0002_taxonomy_library.py`. `0002` both creates the constraint and runs the content-driven seed that inserts the member, so the widen lives there; consolidated into `0002` rather than a separate migration since no DB has run it yet pre-launch.

**fac->rs-gaap bridge — `fac-to-rs-gaap`**
- Add 96 `owl:equivalentClass` arcs so every concrete tenant-kept rs-gaap concept reaches a fac anchor (**137/137, was 41/137**), making `fac` a complete cross-framework substrate over the MVP working set. Most arcs were derived from each concept's nearest fac-anchored calc ancestor; the rest assigned by EFS/liquidity/CF-section. Library-only (the bridge is `tenant_copy:false`); the kept set and `tenant-exclude/v1.json` artifact are unchanged (verified by recomputing `keep_critical`).

**Test fix — `tests/taxonomy/test_framework_bridge_seed.py`**
- Update a stale `resolve_pin` assertion. The `fac` package became `tenant_copy:false` in the earlier tenant-copy curation, so it is omitted from the resolved copy-pin; assert on `fac-traits` (a copied `depends_on` package) instead. This file skips when the extensions DB is absent, so the staleness was hidden in CI — surfaced here only because the change set was verified against a live seeded library.

## Testing

- **Library reseed verified** (`just reset-local` -> recovered via `just migrate-up extensions`): migration chain applies cleanly; recurrence trait seeds (**26 categories / 100 members**); **137/137** concrete kept concepts reach a fac anchor (0 without); tenant-exclude artifact still in exact sync (2012/2012, zero diff); coherence scan all four checks OK; all 3 reporting styles render non-empty BS/IS/CF/SE.
- **Seeded-library integration tests** (run locally with the extensions DB seeded): `test_suggest_liquidity_narrowing.py` + `test_framework_bridge_seed.py` -> **13 passed**.
- **Code quality**: `ruff check`, `ruff format --check`, `basedpyright` on changed Python -> clean (0 errors); pre-commit hooks passed on every commit.
- **`just test-all`** (full unit suite): **not run** this session.

## Notes / Follow-ups

- **Migration consolidated into `0002`** pre-launch (no prod DB has run it). Once `0002` runs in prod it is frozen, so any later trait-axis addition (e.g. an explicit `recurrence/recurring` value, or a `specialItem` axis) will need its own migration.
- **Prod rollout**: the first extensions migration run seeds everything (recurrence axis + bindings + the 96 bridge arcs) via the full chain — no special migration step. The bridge is library-only, so tenant schemas never receive it.
- **Deferred by choice** (from the review): ASC references for ~8 native concepts (additive-recoverable at filing/audit time); a CI lane that seeds the extensions library so the seeded-library integration tests run in CI (the stale `resolve_pin` test fixed here is the concrete consequence of that gap); a deep content audit of `disclosure-mechanics` / `reporting-checklist` / `rollup-rules` / `disclosures-to-textblocks`.
- A design-doc refresh (`local/docs/ref/taxonomy.md`, gitignored) was made locally to match (fac = 4 packages, 26 trait categories); not part of this diff.


jfrench9:
@claude can you review

PR #737, opened 2026-06-12
fix(worker): publish queue depth from a background thread

jfrench9:
## Summary

Worker autoscaling is driven by a `QueueDepth` CloudWatch metric. It was published from inside the consume loop (`QueueDepthReporter.maybe_publish`), called once per iteration *before* `await _process_task(...)`. While a worker processed a task the loop never returned to publish — and several task types run far longer than the 60s publish interval (`extensions_materialize` 1800s, `dagster_job_monitor` 3600s, `graph_tier_upgrade` 3600s, `operator` 600s). At `WORKER_MIN_COUNT=1` a single busy worker went blind for the whole task duration, so the scale-out signal stalled exactly when the queue was backing up behind it.

This is prerequisite work for turning worker autoscaling on — `WORKER_AUTOSCALING_ENABLED_PROD` stays `false`; this PR only makes the signal reliable.

## Changes

- **`robosystems/worker/metrics.py`** — replace the async, leader-elected `QueueDepthReporter` with `QueueDepthPublisher`, a daemon thread that owns its own **sync** Valkey client and publishes `QueueDepth` to the `RoboSystems/Worker/{environment}` namespace every 60s. Running off the consume loop in an OS thread makes it immune to event-loop starvation from the handlers' unguarded blocking calls (sync boto3/SQLAlchemy). Dev still skips the CloudWatch publish.
- **Drop leader election** — every worker publishes the same global `llen("worker:tasks")`. This is correct under the alarms' `Statistic: Maximum` (scale-out on the peak reading, scale-in only when all readings are 0) and removes the 90s leader-failover gap. ≤5 `PutMetricData`/min at `WORKER_MAX_COUNT=5`; the worker IAM role already scopes `cloudwatch:PutMetricData` to this namespace.
- **`robosystems/worker/consumer.py`** — start the publisher at boot, remove the per-loop `maybe_publish()` call, `stop()`/join in `finally`.
- **`robosystems/admin/commands/worker.py`** — drop the now-dead `Metrics leader` line from `worker status`.
- **`cloudformation/worker.yaml`** — flip the scale-out alarm to `TreatMissingData: breaching` (fail open, so a publishing gap errs toward adding capacity); scale-in stays `notBreaching` so a gap never wrongly scales to zero. No effect until autoscaling is enabled (alarms are `Condition: AutoscalingEnabled`).
- **`tests/worker/test_metrics.py`** — new unit tests: publish path (namespace + metric), dev-skip, clean start/stop.

## Testing

- `just test worker` → 27 passed (3 new).
- `just test-code` (ruff + format + basedpyright + cfn-lint) → clean.
- `cfn-lint cloudformation/worker.yaml` → clean.

- Flip `WORKER_AUTOSCALING_ENABLED_PROD=true` once the metric is verified in prod (requires a worker stack redeploy — it gates a CloudFormation `Condition`, not a runtime flag).
- Scale-in / long-task robustness: with autoscaling on, scale-in terminates a worker the same as a deploy — SIGKILL at `StopTimeout=119s` mid-task, after which the reaper only requeues once the full task timeout elapses (crash-tuned), stalling long tasks ~20–30 min. Worth addressing (graceful requeue-on-SIGTERM, scale-in protection, or moving long-running task types off the autoscaled pool) before enabling autoscaling.

### DIFF
--- a/cloudformation/worker.yaml
+++ b/cloudformation/worker.yaml
@@ -517,7 +517,10 @@ Resources:
       EvaluationPeriods: 1
       Threshold: 2
       ComparisonOperator: GreaterThanThreshold
-      TreatMissingData: notBreaching
+      # Fail open: a gap in published depth errs toward adding capacity rather
+      # than silently not scaling. Scale-in stays notBreaching so a gap never
+      # wrongly scales to zero.
+      TreatMissingData: breaching
       AlarmActions:
         - !Ref WorkerScaleOutPolicy
 
--- a/robosystems/admin/commands/worker.py
+++ b/robosystems/admin/commands/worker.py
@@ -152,9 +152,6 @@ def worker_status(client):
     inflight_keys = list(queue.scan_iter(match="worker:inflight:*", count=100))
     inflight_total = sum(queue.llen(k) for k in inflight_keys)
 
-    # Check leader
-    leader = queue.get("worker:metrics:leader")
-
     console.print()
     console.print("[bold cyan]WORKER STATUS[/bold cyan]")
     console.print("=" * 40)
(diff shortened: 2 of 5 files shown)


jfrench9:
@claude can you review

PR #738, opened 2026-06-12
fix(worker): harden ECS task scale-in protection

jfrench9:
## Summary

Follow-up to the ECS task scale-in protection added in `5bca20f8`. An adversarial review of that commit surfaced one blocker and several should-fixes; this hardens the protection path so it can't degrade live task processing. The protection code runs in prod ECS regardless of whether autoscaling is enabled, so these matter before the next worker deploy — not only once autoscaling is on.

## Changes

- **`robosystems/worker/task_protection.py`**
  - **Bound the ECS API call (blocker):** the boto3 `ecs` client is now built with `Config(connect_timeout=2, read_timeout=5, retries={"max_attempts": 2})`. `protect()` is `await`ed *before* the handler runs, so botocore's default 60s×5 retry behavior (~5 min) on a slow/throttled control plane would have stalled the very task it's meant to protect. Worst case is now ~14s.
  - **No permanent self-disable on a transient blip:** a failed metadata fetch increments a counter and only disables protection after `MAX_METADATA_FAILURES` (10) consecutive misses, resetting on success. Previously a single 2s-timeout hiccup on the first read disabled protection for the entire worker-process lifetime.
- **`robosystems/worker/consumer.py`**
  - **Only protect long tasks:** protect/unprotect now fire only when the task's timeout `>= PROTECT_MIN_TIMEOUT_SECONDS` (120s, the worker's `StopTimeout=119` SIGTERM grace). Short tasks (60s types) drain before SIGKILL on scale-in anyway, so protecting them was 2 needless ECS control-plane calls per task — a throttling risk on a busy worker.
- **Tests (`tests/worker/`)**
  - `protect()` swallows an ECS API error without raising (the best-effort contract).
  - protect/unprotect actually fire around a processed task.
  - Invariant guard: `PROTECTION_EXPIRES_MINUTES * 60 > max(TASK_TIMEOUTS)` — fails CI if a task timeout is later bumped past the 90-minute protection lease.
  - Updated the metadata-failure test to assert the new retry-then-disable behavior.

## Testing

- `just test worker` → **37 passed** (4 new/updated worker tests).
- `just test-code` (ruff + format + basedpyright + cfn-lint) → clean.

## Notes / Follow-ups

- The `5bca20f8` parent landed directly on `main` (no PR); this is its review-driven follow-up.
- Not done here (deferred, separate concern): switching from the boto3 control-plane API to the in-container task-protection endpoint (`${ECS_CONTAINER_METADATA_URI_V4}/task-protection/v1/state`), which would drop the `ecs:UpdateTaskProtection` IAM grant and the SDK round-trip entirely. The boto3 path is hardened and well-understood; the endpoint swap is an optimization for later.
- Worker autoscaling itself stays disabled (`WORKER_AUTOSCALING_ENABLED_*=false`); enabling it remains a staging-first ops step.

### DIFF
--- a/robosystems/worker/consumer.py
+++ b/robosystems/worker/consumer.py
@@ -30,7 +30,10 @@ from robosystems.middleware.sse.operation_manager import (
 from robosystems.worker.cleanup import cleanup_connections
 from robosystems.worker.constants import DEFAULT_TASK_TIMEOUT, TASK_TIMEOUTS
 from robosystems.worker.metrics import QueueDepthPublisher
-from robosystems.worker.task_protection import TaskProtectionManager
+from robosystems.worker.task_protection import (
+  PROTECT_MIN_TIMEOUT_SECONDS,
+  TaskProtectionManager,
+)
 from robosystems.worker.tasks import get_task_handler
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


jfrench9:
@claude please review this PR

PR #752, opened 2026-06-17
fix(rate-limits): authed-tier identification + dedicated extensions categories

jfrench9:
## Summary

While turning on the v1.5.0 extensions surface in production, the GraphQL and connection endpoints started returning 429s during normal use. Two root causes: (1) a JWT-decode bug pinned **every authenticated request to the anonymous `base` tier**, and (2) the extensions OLTP surface was bucketed into the **LadybugDB graph rate-limit categories**, so its limits couldn't be tuned independently and were far too low. This PR fixes both.

## Changes

**Authenticated-tier identification (`middleware/rate_limits/rate_limiting.py`)**
- `_verify_jwt_for_rate_limiting` decoded the session JWT with no `audience`/`issuer` args. Tokens are minted with `aud`/`iss` claims (`middleware/auth/jwt.py`), so PyJWT raised `InvalidAudienceError`, the bare `except` swallowed it, and `get_user_from_request` returned `None` → requests fell back to the `base` tier (e.g. `GRAPH_READ` 30/min, `GRAPH_SYNC` 3/min) instead of `ladybug-standard`.
- Now decodes with `verify_aud`/`verify_iss` disabled — the signature is still enforced; rate-limit identification only needs `user_id`.

**Dedicated extensions categories (`config/rate_limits.py`)**
- New `EndpointCategory.EXTENSIONS_GRAPHQL` and `EXTENSIONS_WRITE`. The extensions surface hits the shared extensions PostgreSQL (RDS) — a different backend/cost profile than LadybugDB — so it now has its own buckets instead of borrowing `GRAPH_READ` / `GRAPH_WRITE`.
- `get_endpoint_category` routes `/extensions/{g}/graphql` → `EXTENSIONS_GRAPHQL` and `/extensions/{d}/{g}/operations/*` (plus the report-download fallthrough) → `EXTENSIONS_WRITE`.
- Per-tier limits added to all four tiers: `base` 60/30, `ladybug-standard|large|xlarge` **600/min reads, 300/min writes** (sized for RDS). LadybugDB `GRAPH_READ`/`GRAPH_QUERY`/`GRAPH_ANALYTICS` are now fully independent.

**Tests**
- `tests/middleware/rate_limits/test_rate_limiting.py`: real-decode (un-mocked) regression tests for `_verify_jwt_for_rate_limiting` — token with `aud`/`iss` resolves the user, token without resolves, bad signature returns `None`. (Existing tests mocked the function entirely, which is how the bug shipped.)
- `tests/config/test_rate_limits.py`: categorization cases for the GraphQL endpoint, an operations write, an analytical view, and the report download.

## Testing

- `just test-code` (ruff + format + basedpyright + cf-lint): **passing**.
- `just test` / unit suite: **not run locally** — a prod SSM DB tunnel occupied `localhost:5432` this session, so pytest's DB fixtures hit prod RDS instead of the local test DB. The new tests are unit-level and will run in CI.

## Notes / Follow-ups

- Analytical view operations (`build-fact-grid`, `live-financial-statement`, `financial-statement-analysis`) ride `EXTENSIONS_WRITE` for now because they share the `operations/{name}` path shape; a precise view-name split (they're actually LadybugDB-backed reads) is a follow-up.
- Surfaced by the same review but **out of scope** here (tracked separately): SSM-tunable rate limits (remove the deploy-lock), the unused tier multiplier / all authenticated users flattened to `ladybug-standard`, the QuickBooks `redirect_uri` drift (authorize reads from the request, token-exchange from env), and a reaper for abandoned `pending_oauth` connections.
- Context: discovered during the production turn-on of the v1.5.0 extensions surface.

### DIFF
--- a/robosystems/config/rate_limits.py
+++ b/robosystems/config/rate_limits.py
@@ -51,6 +51,14 @@ class EndpointCategory(str, Enum):
   GRAPH_QUERY = "graph_query"  # Direct Cypher queries
   GRAPH_IMPORT = "graph_import"  # Bulk data imports
 
+  # Extensions surface (OLTP on shared RDS — distinct from LadybugDB categories)
+  EXTENSIONS_GRAPHQL = (
+    "extensions_graphql"  # Typed GraphQL reads: /extensions/{g}/graphql
+  )
+  EXTENSIONS_WRITE = (
+    "extensions_write"  # Command writes + views: /extensions/{d}/{g}/operations/*
+  )
+
   # Table operations (DuckDB staging tables)
   TABLE_QUERY = "table_query"  # SQL queries on staging tables
   TABLE_UPLOAD = "table_upload"  # File uploads to staging tables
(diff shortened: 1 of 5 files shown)


jfrench9:
@claude please review this PR

PR #2, opened 2025-10-06
Fix MCP tools query results to include units and improve dimensions

jfrench9:
## Summary
This PR addresses issues in the MCP tools query functionality by enhancing the data structure and field naming conventions used in query results. The changes focus on improving data completeness and clarity in the elements and facts tools.

## Key Changes
- **Enhanced data completeness**: Modified query results to include unit values alongside dimensional data, ensuring consumers have access to complete measurement information
- **Improved field naming**: Updated dimension field names to follow more consistent and descriptive naming conventions
- **Better data accuracy**: Refined the query result structure to provide more accurate and comprehensive information to downstream consumers

## Files Modified
- `elements_tool.py`: Updated query result structure to include unit information
- `facts_tool.py`: Refactored dimension field naming and enhanced query result completeness

## Breaking Changes
⚠️ **Potential breaking changes**: The dimension field name updates may impact existing integrations that rely on the previous field naming structure. Consumers of these tools should verify their field mappings are still valid.

## Testing Notes
- Verify that query results now include unit values where expected
- Confirm that dimension fields use the updated naming conventions
- Test downstream consumers to ensure compatibility with the new field structure
- Validate that the enhanced query results maintain backward compatibility where possible

## Infrastructure Considerations
- Monitor MCP tool consumers for any issues related to the field name changes
- Consider updating any documentation that references the old dimension field names
- Review logging and monitoring to ensure the enhanced query results are captured appropriately

---
🤖 Generated with [Claude Code](https://claude.ai/code)

**Branch Info:**
- Source: `bugfix/mcp-tools-query-fix`
- Target: `main`
- Type: bugfix

Co-Authored-By: Claude <noreply@anthropic.com>

PR #761, opened 2026-06-18
fix(mcp): make the close workflow operable from the MCP surface alone

jfrench9:
## Summary

Three MCP workflow-legibility fixes surfaced while dogfooding the RoboLedger month-end close through the MCP surface (Claude Cowork / raw MCP). Authoring the recurring schedules turned out to require *codebase awareness* — which is a launch problem, since "multi-company accountant runs closes via Claude" is the headline persona. This PR makes the close workflow operable from the tool surface alone, without reversing the generic Information Block dispatcher architecture.

The gap was localized (a few schema/description seams), not systemic. Each fix was validated by a **codebase-blind agent** given only the generated MCP tool catalog and forbidden from reading source.

## Changes

**Fix #1 — dead link** (`0dc28356`)
- `list-period-drafts` told callers to draft entries via a `create-closing-entry` tool that does not exist. It now names the real draft-producing paths: `promote-obligations` (period sweep), `create-event-block(event_type='schedule_entry_due')` (single schedule), `journal_entry_recorded` (manual).

**Fix #2 — typed payloads at the MCP boundary** (`0dc28356`)
- `derive_input_schema` flattened every discriminated-union `RootModel` body (create/update/delete-information-block) into an opaque `payload: {additionalProperties: true}`, so the typed schedule arm's real fields and its `json_schema_extra` examples never reached the MCP client.
- `_flatten_top_level_union` now correlates each arm to its discriminator value(s); when an arm carries a typed payload it exposes every arm's payload schema under a nested `payload.anyOf` (titled by `block_type`) and hoists the arms' examples to the envelope. Nested unions are safe — `Optional[...]` fields already emit nested `anyOf`. All-untyped unions keep the prior permissive shape, so existing behavior and its regression test are unchanged.
- New regression tests in `tests/middleware/mcp/test_registrar.py`.

**Fix #3 — `get-close-playbook` + element-id legibility** (`b612b386`)
- New `get-close-playbook` MCP tool (`robosystems/middleware/mcp/tools/playbook_tools.py`): version-locked, guidance-only, read-only-safe. Encodes the canonical close chain, the two modes (initiate / recurring), the bootstrap strategy (derive-from-history + interview), and the gotchas no single schema makes obvious — one schedule = one debit/credit element pair, amounts in cents, period-identifier shapes (`YYYY-MM` vs `YYYY-MM-DD`), which block types 501, and close blockers. Wired into `manager.py` (registration + listing + dispatch). This MCP surface has no protocol-level `instructions` field, so discovery is via prominent pointers from `get-fiscal-calendar` and `create-information-block`.
- Element-id bug the gate caught: the schedule machinery resolves debit/credit by CoA `Element.id`, but the `create-information-block` example used taxonomy qnames (`us-gaap:Depreciation`). Fixed the example (placeholder element ids + note) and the `EntryTemplateRequest` / `CreateScheduleRequest` field descriptions to say "CoA element id (from `get-unmapped-elements`), not a qname." The rollforward block type keeps its `*_qname` fields by design.
- New tests in `tests/middleware/mcp/tools/test_playbook_tools.py`.

**Disambiguation + stale-doc cleanup** (`44163621`)
- `create-taxonomy-block` accepts a structure `block_type='schedule'`, but that only writes a bare ontology row (no per-period facts, no `schedule_entry_due` obligations, no closing-entry generator). Its description now points callers to `create-information-block(block_type='schedule')` for a functional schedule.
- Scrubbed stale references to dedicated `create-schedule` / `update-schedule` / `delete-schedule` ops — those were consolidated into the generic Information Block dispatcher; no such ops exist today.

## Testing

- `just test-code` (ruff + format + basedpyright + cf-lint) — **passed** (run repeatedly via the pre-commit hook on each commit).
- `uv run pytest tests/middleware/mcp/` — **574 passed**.
- `uv run pytest tests/routers/extensions/roboledger/` — **passed**.
- `uv run pytest tests/ -k "information_block or schedule or close_period or fiscal"` — **486 passed**.
- Full `just test-all` (unit-test portion) not run in this session.

## Validation — codebase-blind gate runs

A `general-purpose` subagent was handed only the generated MCP tool catalog (JSON) and forbidden from reading source:

- **Gate 1 (after #1+#2):** running a close flipped to YES; the schedule-payload field-name-guessing blocker was resolved ("complete call, all from schema/examples, NOT guessed").
- **Gate 2 (after #3):** the agent called `get-close-playbook` first, derived the one-schedule-per-pair decomposition, used the element-discovery tools, and ran the full close. It also flagged the qname-vs-element-id ambiguity (now fixed) and wrongly worried that omitting `schedule_metadata` would create a failing `SumEquals` rule (verified harmless — the rule only generates when a cost basis is supplied).

## Notes / Follow-ups

- Deferred (demand-pulled, not blockers): per-tenant procedures-doc *seeding* at provision; an auto-extending schedule horizon; a `list-prior-entries-by-account` read tool (`read-graph-cypher` covers it today).
- Considered and deferred: dedicated `create-schedule`/etc. ops (option B). If SDK ergonomics or the constructible-block-type count demand it, the only version worth building is auto-generated *from the block-type registry* so it adds no per-type boilerplate and can't drift — schedule logic stays in `ScheduleService` either way.

### DIFF
--- a/robosystems/middleware/mcp/tools/fiscal_calendar_tools.py
+++ b/robosystems/middleware/mcp/tools/fiscal_calendar_tools.py
@@ -89,6 +89,8 @@ class GetFiscalCalendarTool:
       "name": "get-fiscal-calendar",
       "description": """Get the current fiscal calendar state — what period is next to close, what's blocking it.
 
+**NEW TO THE CLOSE? Call `get-close-playbook` first** — it lays out the full tool sequence, the schedule-setup decisions, and the gotchas this tool's output assumes you know.
+
 **WHEN TO USE:**
 - At the START of every close session — check what period you're working on
 - Before calling close-period — verify it will succeed (check `closeable_now`)
--- a/robosystems/middleware/mcp/tools/manager.py
+++ b/robosystems/middleware/mcp/tools/manager.py
@@ -294,6 +294,15 @@ class GraphMCPTools:
       self.get_information_block_tool = GetInformationBlockTool(graph_client)
       self.list_information_blocks_tool = ListInformationBlocksTool(graph_client)
 
+    # Workflow playbook (guidance only — version-locked to this build).
+    # Pure read with no DB access, so it stays available on read-only graphs:
+    # an operator can learn the close workflow before they have write access.
+    self.get_close_playbook_tool = None
+    if self._has_extension("roboledger") and env.ROBOLEDGER_ENABLED:
+      from .playbook_tools import GetClosePlaybookTool
+
+      self.get_close_playbook_tool = GetClosePlaybookTool(graph_client)
+
     # Agent reads (get-agent, list-agents, agent-activity)
     self.get_agent_tool = None
     self.list_agents_tool = None
@@ -554,6 +563,10 @@ class GraphMCPTools:
     surface through the generic information-block read tools.
     """
     tools = []
+    # Playbook first so it's the most prominent close-workflow entry in the
+    # listing — it tells the agent how the rest of these tools compose.
+    if self.get_close_playbook_tool is not None:
+      tools.append(self.get_close_playbook_tool.get_tool_definition())
     if self.get_period_close_status_tool is not None:
       tools.append(self.get_period_close_status_tool.get_tool_definition())
     if self.list_period_drafts_tool is not None:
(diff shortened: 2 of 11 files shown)


jfrench9:
@claude please review this PR — focus on the registrar schema-flattening change (`_flatten_top_level_union` in robosystems/middleware/mcp/tools/registrar.py): confirm nested `payload.anyOf` + hoisted `examples` are valid in an Anthropic tool `input_schema`, the all-untyped path is unchanged, and dispatch-time validation still enforces the real contract.

PR #766, opened 2026-06-19
fix(schedules): contra-aware roll-forward direction from QB AccountSubType

jfrench9:
## Summary

Prepaid amortization schedules rendered their asset balance **climbing toward cost** instead of **depleting to zero**, while depreciation/amortization schedules (which use a contra account) were correct. The fix gives the schedule generator chart-of-accounts role awareness so a directly-credited asset (a prepaid) draws down from cost while a contra-asset (accumulated depreciation/amortization) accumulates up — driven by data QuickBooks already emits.

Surfaced during the first production close (Harbinger FinLab).

## Root cause

QuickBooks files contra-assets (Accumulated Depreciation/Amortization) under an asset `AccountType` (`Fixed Asset`) with `Classification=Asset` — identical top-level classification to a prepaid (`Other Current Asset`). Our dbt `normal_balance` macro maps both to `balance_type='debit'`, so balance_type alone cannot distinguish them. The one field that *does* distinguish them — QB's `AccountSubType` (`AccumulatedDepreciation` vs `PrepaidExpenses`) — was being **dropped at load** (the `Element` table kept only `metadata.account_type`).

The earlier partial fix keyed the draw-down on `balance_type` (unreliable, as above) **and** required `original_amount` (absent on these schedules), so neither branch ever fired.

## Changes

- **dbt** (`adapters/quickbooks/dbt/models/ledger/elements.sql`) — persist `account_sub_type` into `Element.metadata` (JSONB; no migration).
- **loader** (`operations/extensions/loader.py`) — add `_QB_CONTRA_ASSET_SUB_TYPES` + `_is_qb_contra_asset_sub_type`; in `_apply_source_element_traits`, a contra sub-type now receives the FASB **`contraAsset`** `elementsOfFinancialStatements` trait instead of the plain `asset` its account_type implies. Robust contra identification *at element creation* — also correct for rollups and the auto-map agent, not just schedules.
- **schedule service** (`operations/roboledger/schedules/service.py`) — add `_element_is_contra_asset` (reads the `contraAsset` trait, so it's source-agnostic); drive roll-forward direction off it (`credit_draws_down = balance_type=='debit' and not is_contra`); and when `original_amount` is absent, derive the opening cost from the straight-line curve (`monthly x periods`) so a prepaid declines from cost to ~0 (this also correctly anchors the final-period rounding, NBV, and the SumEquals rule).
- **tests** — `tests/operations/extensions/test_loader.py`: contra sub-type maps to the `contraAsset` trait, plus the `_is_qb_contra_asset_sub_type` predicate (known sub-types, `Accumulated*` catch-all, direct assets, None/empty). `tests/operations/roboledger/schedules/test_service.py`: debit-balance contra accumulates; prepaid without `original_amount` derives the opening and draws down; updated the existing mock execute-ordering for the new contra-trait query.
- **`.env.example`** — correct `INTUIT_REDIRECT_URI` to the actual roboledger-app callback (`http://localhost:3001/connections/qb-callback`); the old `http://localhost:8000/auth/callback` was wrong.

## Testing

- `just test` (full unit suite) — **10537 passed, 15 skipped** (~7m).
- `just test-code` (ruff + format + basedpyright + cf-lint) — **clean** (also runs via pre-commit on each commit).
- **End-to-end on a live QuickBooks sandbox** (full-rebuild sync through the real pipeline): `metadata.account_sub_type` persisted; the contra got the `contraAsset` trait (prepaid got `asset`); `_element_is_contra_asset()` returned True/False correctly on the real synced elements. Same generator, $100/mo over 3 periods, no `original_amount`: prepaid **depletes** 300 -> 0, depreciation **accumulates** 0 -> 300.

## Notes / Follow-ups

- **No DB migration** — the contra signal rides in the existing JSONB `metadata`.
- **Existing graphs need a re-sync** to apply the `contraAsset` trait / `metadata.account_sub_type` to already-loaded elements; new syncs are correct from the start. Harbinger remediation: re-sync, then re-create the prepaid schedules to pick up the corrected (depleting) roll-forward. Depreciation/amortization schedules were already accumulating correctly.
- GL postings were never affected — only the roll-forward **balance tracking** for directly-credited assets.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -112,7 +112,7 @@ OPENSEARCH_URL=http://robosystems-opensearch:9200
 INTUIT_CLIENT_ID=your-intuit-client-id-here
 INTUIT_CLIENT_SECRET=your-intuit-client-secret-here
 INTUIT_ENVIRONMENT=sandbox
-INTUIT_REDIRECT_URI=http://localhost:8000/auth/callback
+INTUIT_REDIRECT_URI=http://localhost:3001/connections/qb-callback
 
 ## CORS — extra origins for dev (e.g., ngrok tunnels for OAuth callbacks).
 ## Comma-separated. Dev only — staging/prod use hardcoded lists.
--- a/robosystems/adapters/quickbooks/dbt/models/ledger/elements.sql
+++ b/robosystems/adapters/quickbooks/dbt/models/ledger/elements.sql
@@ -35,5 +35,8 @@ select
     'USD'                                       as currency,
     a.is_active,
     false                                       as is_placeholder,
-    json_object('account_type', a.account_type) as metadata
+    json_object(
+      'account_type', a.account_type,
+      'account_sub_type', a.account_sub_type
+    )                                            as metadata
 from accounts a
--- a/robosystems/operations/extensions/loader.py
+++ b/robosystems/operations/extensions/loader.py
@@ -143,16 +143,31 @@ _QB_CONTRA_ASSET_SUB_TYPES: frozenset[str] = frozenset(
 )
 
 
+# Accumulated-* sub-type families that are contra-ASSETS. Used as a
+# forward-compatible catch-all (QB periodically adds new variants, e.g.
+# AccumulatedDepreciationEquipment). Deliberately NOT a bare
+# ``startswith("Accumulated")`` — ``AccumulatedOtherComprehensiveIncome``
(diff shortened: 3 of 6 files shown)


jfrench9:
@claude please review this PR

PR #786, opened 2026-06-29
feat(qb): cut JournalReport over to Intuit v2 reporting service

jfrench9:
## Summary

Intuit is migrating the QuickBooks Reports APIs to its modernized ("v2") reporting service. After the **2026-08-31** cutover, all Reports responses are served by v2 unconditionally and the temporary `testing_migration` opt-in param becomes a no-op. `JournalReport` is our **live GL posting source** (`adapters/quickbooks/pipeline/extract.py`), so this PR cuts it over to v2 ahead of the deadline — under our control, with v1 still reachable as a runtime fallback until the cutover.

Validated non-destructively against a live sandbox tenant's full history before defaulting it on: v1 and v2 parse **byte-identically** (129 transactions / 339 lines / $81,064.53, balanced both ways), no nested rows, unchanged 8-column shape. The only structural difference is v2 dropping redundant per-transaction header rows, which `parse_journal_report` already skips. The migration is effectively a no-op for our parser.

## Changes

- **`config/env.py`** — new `INTUIT_REPORTS_TESTING_MIGRATION` flag, **default `true`**. Set `false` via SSM (`features/INTUIT_REPORTS_TESTING_MIGRATION`) to fall back to v1 at runtime with no redeploy (only meaningful until the 2026-08-31 cutover).
- **`adapters/quickbooks/client/api.py`** — `QBClient.get_transactions` takes a `testing_migration` arg (defaults to the config flag); when set, appends `testing_migration` to the JournalReport request so it routes through Intuit's v2 service.
- **`tests/adapters/qb/pipeline/test_utils.py`** — `test_v2_modernized_report_shape`: parser regression guard covering the documented v2 changes that can surface in a JournalReport (typed `Section`/`Data` rows, Header block with StartPeriod/EndPeriod, empty-string nulls, extra trailing columns) and asserting group boundaries still hold.
- **`tests/adapters/qb/client/test_api.py`** — two tests for the param (explicit arg + flag-driven); the three existing date-handling tests pass `testing_migration=False` to stay isolated from the now-default-on flag.

## Testing

- `uv run pytest tests/adapters/qb/` — **143 passed**.
- `ruff check`, `ruff format --check`, `basedpyright` on changed files — clean (also confirmed by the pre-commit hook).
- **Live validation:** ran a non-destructive v1-vs-v2 `JournalReport` diff against a connected sandbox tenant (full history) — parsed output byte-identical, balanced, no nesting.
- Full `just test-all` — not run in this session.

## Notes / Follow-ups

- **After 2026-08-31:** remove the flag and the `get_transactions` param (dead code once Intuit forces v2 server-side).
- If the QB Phase 6 work (JournalReport retirement → per-entity `Line[]` synthesis) lands first, this flag is moot — JournalReport is no longer called.
- Rollout posture: deploy with v2 on; if anything looks off in production, flip the SSM flag to `false` for an instant fallback to v1.

### DIFF
--- a/robosystems/adapters/quickbooks/client/api.py
+++ b/robosystems/adapters/quickbooks/client/api.py
@@ -507,17 +507,27 @@ class QBClient:
     return self._paginate(JournalEntry)
 
   @_QB_RETRY
-  def get_transactions(self, start_date=None, end_date=None):
-    """Fetch JournalReport — no longer the transactional source.
+  def get_transactions(self, start_date=None, end_date=None, testing_migration=None):
+    """Fetch JournalReport — still the live GL posting source (extract.py).
 
-    Kept for backward compat with any external callers; the QB pipeline now
-    pulls Invoice / Bill / Payment / JournalEntry per-entity instead.
+    When ``testing_migration`` is True — or, when left None, the
+    ``INTUIT_REPORTS_TESTING_MIGRATION`` config flag is set — the request is
+    routed through Intuit's modernized "v2" reporting service via the
+    ``testing_migration`` query param. This is a temporary Intuit-side
+    migration switch: after the 2026-08-31 cutover the v2 service serves all
+    Reports API responses unconditionally and the param becomes a no-op.
+    See https://medium.com/intuitdev/upcoming-changes-to-reports-apis-5083ec9aadce
     """
     params = {}
     if start_date:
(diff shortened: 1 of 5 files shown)


jfrench9:
@claude please review this PR

PR #789, opened 2026-07-01
fix(sec): key incremental quarter selection off Eastern time

jfrench9:
## Summary

The nightly incremental SEC download selected its quarter from the container's **UTC clock**, so the 21:00 ET run on a quarter's last day (already the next day in UTC) rolled forward to the next quarter — and, because it also ran a "previous-quarter overlap," fired a second run for the prior quarter. On 2026-06-30 this produced two runs: **Q3 (empty — not Q3 yet)** and **Q2 (the real data)**.

This keys quarter selection off **Eastern time** (the SEC filing calendar) and does a **hard cut-over**: exactly one quarter per run, no previous-quarter overlap.

- `2026-06-30 21:00 ET` → `["2026-Q2"]`
- `2026-07-01 21:00 ET` → `["2026-Q3"]`

## Changes

- **`adapters/sec/__init__.py`** — `get_current_quarter()` and `get_quarters_to_scan()` default to Eastern (`America/New_York`) instead of UTC; `get_quarters_to_scan()` now returns a single current quarter. Removed the overlap machinery (`is_in_quarter_overlap_window`, `get_previous_quarter`, `QUARTER_OVERLAP_DAYS`).
- **`adapters/sec/pipeline/sensors.py`** — the incremental download schedule keys the quarter off `context.scheduled_execution_time` (already tz-aware in `America/New_York`) and yields a single `RunRequest`; `_get_quarters_to_scan()` threads that time through.
- **`adapters/sec/processors/ingestion/staging.py`** — incremental DuckDB staging stages exactly the target quarter; removed its own UTC-based overlap block (same `datetime.now(UTC)` bug).
- **`tests/adapters/sec/pipeline/test_sensors.py`** — replaced the obsolete "two quarters at boundary" test; added ET regression tests locking in `6/30 21:00 ET → 2026-Q2`, `7/1 → 2026-Q3`, and single-quarter-always.

## Testing

- Import smoke exercised the real functions: `6/30 21:00 ET → ['2026-Q2']`, `7/1 21:00 ET → ['2026-Q3']`.
- Pre-commit gate green: `ruff` clean, `ruff format` clean, `basedpyright` 0 errors.
- Unit tests written but **not executed locally** — this test path's conftest requires Postgres and the local stack was down this session. They run in CI.

## Notes

- **Deliberate trade-off:** the removed overlap re-scanned the previous quarter for the first 5 days of a new quarter to catch filings SEC indexes 1–5 days late. Hard cut-over drops that — a filing filed near quarter-end but indexed after the cut-over could be missed. We trust the final batch of each quarter to capture its filings.
- Surfaced while dry-running the SEC shared-master off-hours parking: with the master parked, the spurious empty Q3 run failed loudly (`staging: Connection error: All connection attempts failed`) instead of silently doing nothing.

### DIFF
--- a/robosystems/adapters/sec/__init__.py
+++ b/robosystems/adapters/sec/__init__.py
@@ -1,12 +1,12 @@
 """SEC EDGAR adapter for XBRL financial data extraction."""
 
-from datetime import UTC, datetime
+from datetime import datetime
+from zoneinfo import ZoneInfo
 
-# Quarter overlap window for incremental updates.
-# SEC filings submitted near quarter-end may not appear in EFTS for 1-2 days
-# due to SEC indexing delays and UTC/EST timing differences. We scan the
-# previous quarter during the first N days of a new quarter to catch these.
-QUARTER_OVERLAP_DAYS = 5
+# SEC filings follow the US-market (Eastern) calendar. The nightly pipeline runs
+# at 21:00 ET; on a quarter's last day that is already the next day in UTC, so
+# quarter selection must key off Eastern time, not the container's UTC clock.
+EASTERN = ZoneInfo("America/New_York")
 
 _LAZY_IMPORTS = {
   "SEC_BASE_URL": "robosystems.adapters.sec.client",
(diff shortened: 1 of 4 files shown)


jfrench9:
@claude please review this PR

PR #799, opened 2026-07-02
fix(security): app-layer hardening from SOC 2 review

jfrench9:
## Summary

Three app-layer security fixes surfaced by a SOC 2 readiness review of the platform: broken de-provisioning, fail-open authentication rate limiting, and an operation-audit trail that silently dropped all of its fields. All three are code-level fixes with no schema or infra changes.

This is PR 1 of a 4-part security sprint (the pure-code, zero-ops batch).

## Changes

**De-provisioning — a deactivated user kept full API access (Critical)**
- `middleware/auth/utils.py` — `validate_api_key` and `validate_api_key_with_graph` now reject a key whose **owning user** is deactivated, on both the cache-hit and DB paths. `UserAPIKey.get_by_key` filters only on the key's own `is_active` flag, and API keys carry no `session_version`, so previously a deactivated user retained full programmatic access via any existing key.
- `models/core/user/user.py` — `User.deactivate()` now cascades via `_revoke_api_keys()` to deactivate all of the user's API keys and invalidate each key's validation-cache entry, alongside the existing `session_version` bump. Keys must be regenerated after reactivation (consistent with the method's existing "deactivation is a security boundary" behavior).

**Fail-closed auth rate limiting**
- `middleware/rate_limits/cache.py` — `check_rate_limit` gains a `fail_closed` parameter; when the limiter backend (Valkey) errors, it denies instead of allowing.
- `middleware/rate_limits/rate_limiting.py` — wired `fail_closed=True` into `auth_rate_limit_dependency`, which guards `login`, `register`, `password_reset`, and `email_verification`. General traffic still fails open, so a Valkey outage hardens the auth surface rather than taking down the whole API.

**Audit trail — operation-audit lines were content-free**
- `config/logging.py` — `StructuredFormatter` now serializes the `audit` payload (emitted via `extra={"audit": ...}` by `log_operation_audit`) plus `ip_address` / `success` from the security-logging middleware. The formatter's field allowlist previously omitted these, so the "who / what / which-tenant / result" audit trail was being dropped.

## Testing

- `just test-code` — **passed** (ruff + format + basedpyright: 0 errors, 0 warnings).
- `uv run pytest` on the four affected modules — **81 passed (7 new)**:
  - `tests/middleware/auth/test_utils.py` — inactive-user rejected on cache and DB paths (+ graph variant)
  - `tests/models/core/user/test_user_model.py` — DB-backed `deactivate` revokes all keys
  - `tests/config/test_logging.py` — formatter serializes `audit` / `ip_address` / `success`
  - `tests/middleware/rate_limits/test_rate_limit_cache.py` — fail-closed denies on backend error; default still fails open

## Notes / Follow-ups

- **Fail-closed scope is deliberate.** Only `auth_rate_limit_dependency` (login, register, password_reset, email_verification) fails closed. The other auth-adjacent limiters (sso, jwt-refresh, sensitive-auth, logout, auth-status) still fail open — extending them is a clean follow-up, deferred here to avoid broadening the outage surface during a limiter-backend incident.
- **Operational note:** prod runs a single Valkey node, so a cache outage now returns `429` on those four auth endpoints (no logins until Valkey recovers). This is the intended security tradeoff; the `logger.error("Rate limiter failing CLOSED…")` line is the hook to alert on so an incident reads as a cache outage rather than an unexplained login failure.
- **Reactivation requires new API keys** — deactivation now permanently revokes them.
- Remaining sprint PRs: **PR 2** MCP Cypher hardening (confirmed-exploitable `LOAD`/`ATTACH`/`INSTALL` bypass on the read path), **PR 3** infra/TLS/RDS + staging secret validation, **PR 4** scope down the CI/CD OIDC role.

### DIFF
--- a/robosystems/config/logging.py
+++ b/robosystems/config/logging.py
@@ -83,6 +83,19 @@ class StructuredFormatter(logging.Formatter):
     if hasattr(record, "request_id"):
       log_entry["request_id"] = record.request_id
 
+    # Structured operation-audit payload (who / what / which-tenant / result).
+    # Emitted via extra={"audit": {...}} by middleware.operations.log_operation_audit;
+    # without this the entire audit payload was silently dropped.
+    if hasattr(record, "audit"):
+      log_entry["audit"] = record.audit
+
+    # Security / auth event fields emitted by the security-logging middleware.
+    # Source IP and success flag are required for auth-anomaly detection.
+    if hasattr(record, "ip_address"):
+      log_entry["ip_address"] = record.ip_address
+    if hasattr(record, "success"):
+      log_entry["success"] = record.success
+
     return json.dumps(log_entry, default=str, separators=(",", ":"))
 
 
--- a/robosystems/middleware/auth/utils.py
+++ b/robosystems/middleware/auth/utils.py
@@ -72,6 +72,13 @@ def validate_api_key(api_key: str, db_session: Session | None = None) -> User | 
       user.email = user_data.get("email")
       user.is_active = user_data.get("is_active", True)
 
+      # Reject keys owned by a deactivated user, even when the key row and its
+      # cache entry are still marked active. Deactivation invalidates the cache
+      # entry, but this guard closes the window until that propagates.
+      if not user.is_active:
+        logger.debug(f"API key rejected: owning user inactive: {cache_key[:8]}...")
+        return None
+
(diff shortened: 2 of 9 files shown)


jfrench9:
@claude please review this PR

PR #800, opened 2026-07-02
fix(security): close read-graph-cypher LOAD/ATTACH/INSTALL bypass

jfrench9:
## Summary

Closes a confirmed-exploitable bypass in the MCP `read-graph-cypher` tool: its read-only validator missed the bulk and admin Cypher verbs, so a "read-only" tool could read arbitrary local files, reach the instance-metadata endpoint (SSRF), and `ATTACH` other tenants' databases — reachable even on the read-only shared repositories (`sec`, `sec_historical`).

This is PR 2 of a 4-part security sprint (following #799).

## The bug

`CypherTool._validate_read_only` validated with a hand-rolled keyword list that only blocked writes (`CREATE`/`SET`/`DELETE`/`REMOVE`/`MERGE`/`DROP`/`DETACH DELETE`/`CALL db.`/`CALL apoc.`). It did **not** block:

- **Bulk:** `LOAD`, `COPY`, `IMPORT` → `LOAD FROM '/etc/passwd' RETURN *` (arbitrary local file read)
- **Admin:** `EXPORT`, `INSTALL`, `ATTACH`, `USE` → `INSTALL httpfs; LOAD FROM 'http://169.254.169.254/...'` (SSRF to IAM credentials), `ATTACH '/data/lbug-dbs/<other_graph>'` (cross-tenant access)

LadybugDB executes these verbs. The REST `/query` endpoint already blocks them via `security.cypher_analyzer`; only the MCP read path diverged.

**Confirmed at runtime:** a `LOAD FROM …` query passed the old validator and reached the engine (HTTP 400 from LadybugDB, i.e. an engine-level error, not a validator rejection).

## Changes

- `middleware/mcp/tools/cypher_tool.py` — `_validate_read_only` now delegates the dangerous categories to the central `security.cypher_analyzer` (the same analyzer the REST `/query` endpoint uses): `is_bulk_operation`, `is_admin_operation`, `is_schema_ddl`. The existing write-keyword and `CALL` checks are preserved, so there is no read-path regression.

## Testing

- `uv run pytest` on the MCP suites — **698 passed (17 new)**, no regressions:
  - new `tests/middleware/mcp/tools/test_cypher_tool.py` covers: bulk/admin verbs rejected (incl. the `INSTALL httpfs; LOAD FROM http://169.254.169.254/…` SSRF payload), writes still blocked, and legit reads pass — including a bulk keyword quoted inside a string literal (`CONTAINS 'load'`), which must **not** false-positive.

## Notes / Follow-ups

- **Deeper defense-in-depth deferred.** The strongest fix is to open the query connection `read_only=True` at the Graph API layer so read tools can't execute privileged verbs regardless of keyword lists. That's a larger change into `graph_api`; this PR closes the exploit at the validator and brings MCP in line with REST.
- The false-positive/consistency profile now matches the REST path exactly (both use the shared analyzer), which also reduces the "three divergent validators" drift risk.
- Remaining sprint PRs: **PR 3** infra/TLS/RDS + staging secret validation, **PR 4** scope down the CI/CD OIDC role.

### DIFF
--- a/robosystems/middleware/mcp/tools/cypher_tool.py
+++ b/robosystems/middleware/mcp/tools/cypher_tool.py
@@ -185,9 +185,35 @@ RETURN DISTINCT labels(a)[0] AS from_type, type(r) AS rel_type, labels(b)[0] AS 
         query: Cypher query to validate
 
     Raises:
-        ValueError: If query contains write operations
+        ValueError: If query contains write, bulk, admin, or schema-DDL operations
     """
     import re
+
+    # Delegate the dangerous categories to the central security analyzer — the
+    # same one the REST /query endpoint uses — so this read path can't diverge
+    # from it. The hand-rolled keyword list below previously missed LOAD / COPY /
+    # ATTACH / INSTALL / EXPORT, letting a "read-only" tool read local files,
+    # reach instance metadata (SSRF), and ATTACH other tenants' databases.
+    from robosystems.security.cypher_analyzer import (
+      is_admin_operation,
+      is_bulk_operation,
+      is_schema_ddl,
+    )
+
+    if is_bulk_operation(query):
+      logger.warning("Blocked bulk operation (COPY/LOAD/IMPORT) in read-graph-cypher")
+      raise ValueError("Only read-only queries are allowed")
+
+    if is_admin_operation(query):
+      logger.warning(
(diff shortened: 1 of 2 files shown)


jfrench9:
@claude please review this PR

PR #801, opened 2026-07-02
fix(security): TLS policy, RDS deletion protection, staging secret validation

jfrench9:
## Summary

Infrastructure and configuration hardening from the SOC 2 review: modern TLS on the public ALB, RDS deletion protection + longer backup retention, and secret validation that now covers staging (not just prod).

This is PR 3 of a 4-part security sprint (following #799, #800). It mixes CloudFormation (verified by `cfn-lint`, needs a staging deploy to confirm) with pure-code config changes (unit-tested).

## ⚠️ Deploy gate — read before merging to staging

This PR makes staging **fail-fast on startup if `CONNECTION_CREDENTIALS_KEY` is not set** in the staging secret (`robosystems/staging`). That's the intended fix (staging previously booted unvalidated), but it means:

1. **Before this reaches staging**, confirm `CONNECTION_CREDENTIALS_KEY` exists in the staging secret.
2. If staging has been running on the JWT-derived fallback, set a real key there first — existing sandbox QuickBooks connections were encrypted under the old derived key and would need re-authorization.

Prod already requires this key (and boots), so prod is unaffected.

## Changes

**ALB TLS (`cloudformation/api.yaml`)**
- Pin `SslPolicy: ELBSecurityPolicy-TLS13-1-2-2021-06` on the public HTTPS listener. Without an explicit policy the ALB default (`ELBSecurityPolicy-2016-08`) still negotiates TLS 1.0/1.1.

**RDS (`cloudformation/postgres.yaml`)**
- Add `DeletionProtection` via a new `EnableDeletionProtection` parameter (default `true`) + condition — protects the prod/staging DB from accidental deletion. Overridable to `false` when an intentional teardown is needed.
- Raise `BackupRetentionPeriod` from 7 to 30 days.
- **Multi-AZ intentionally left off** (deliberate cost decision) — these two are the no-added-capacity-cost insurance items.

**Secret validation (`config/validation.py`, `main.py`)**
- Extend the required-secret checks (`DATABASE_URL`, `JWT_SECRET_KEY`, `VALKEY_URL`, `AWS_REGION`, `CONNECTION_CREDENTIALS_KEY`) and the startup fail-fast from `prod` to `prod`+`staging`. The Stripe `sk_test_` prod-only check is left as-is (staging legitimately uses test keys).

**Credential-key fallback (`models/core/connection/connection_credentials.py`)**
- `_get_encryption_key` now refuses the weak, unsalted JWT-derived fallback in prod/staging and raises; the fallback remains only for dev/test/local. Defense-in-depth behind the startup validation above.

## Testing

- `cfn-lint` — **PASS** on both `postgres.yaml` and `api.yaml`.
- `uv run pytest` on the affected modules — **53 passed (5 new)**:
  - `tests/config/test_validation.py` — staging now enforces critical vars (success + missing-`CONNECTION_CREDENTIALS_KEY` failure)
  - `tests/models/core/connection/test_connection_credentials.py` — prod/staging refuse the JWT fallback; dev still allows it
- CloudFormation changes are `cfn-lint`-validated only; **staging deploy required** to confirm the TLS policy and RDS settings apply cleanly.

## Notes / Follow-ups

- S3 versioning on the user-data/backup bucket (the source-of-truth Parquet recovery angle) was scoped out of this PR — candidate follow-up.
- Remaining sprint PR: **PR 4** scope down the CI/CD OIDC role (the account-takeover primitive) — deployed carefully on its own since it modifies the deploy role itself.

### DIFF
--- a/.github/workflows/deploy-postgres.yml
+++ b/.github/workflows/deploy-postgres.yml
@@ -64,6 +64,13 @@ on:
         type: string
         default: "false"
 
+      # Accidental-deletion protection (recommended on for prod/staging; no added cost)
+      deletion_protection_enabled:
+        description: "Block accidental deletion of the RDS instance (set false only for an intentional teardown)"
+        required: false
+        type: string
+        default: "true"
+
       # RDS Proxy Configuration (off by default)
       rds_proxy_enabled:
         description: "Enable RDS Proxy for connection pooling (costs ~$0.015/vCPU/hour when enabled)"
@@ -156,6 +163,7 @@ jobs:
                 ParameterKey=DBAllocatedStorage,ParameterValue=${{ inputs.allocated_storage }} \
                 ParameterKey=DBMaxAllocatedStorage,ParameterValue=${{ inputs.allocated_max_storage }} \
                 ParameterKey=EnableMultiAZ,ParameterValue=${{ inputs.multi_az_enabled }} \
+                ParameterKey=EnableDeletionProtection,ParameterValue=${{ inputs.deletion_protection_enabled }} \
                 ParameterKey=EnableRDSProxy,ParameterValue=${{ inputs.rds_proxy_enabled }} \
                 ParameterKey=RDSProxyMaxConnectionsPercent,ParameterValue=${{ inputs.rds_proxy_max_connections_percent }} \
                 ParameterKey=RDSProxyConnectionBorrowTimeout,ParameterValue=${{ inputs.rds_proxy_connection_borrow_timeout }} \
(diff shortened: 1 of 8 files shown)


jfrench9:
@claude please review this PR. Pay particular attention to: (1) the RDS DeletionProtection parameter/condition wiring in cloudformation/postgres.yaml and whether it could block legitimate stack operations, (2) whether extending required-secret validation + startup fail-fast to staging could break a staging deploy if CONNECTION_CREDENTIALS_KEY is unset, and (3) the connection_credentials.py fallback guard.

PR #803, opened 2026-07-02
fix(security): bound extensions GraphQL query depth, aliases, and tokens

jfrench9:
## Summary

Bounds query cost on the extensions GraphQL endpoint, which previously had no depth or complexity limiting. An authenticated user could send a deeply nested or alias-heavy query and exhaust the small extensions OLTP connection pool (each resolved field can open a session). Pure-code change, unit-tested.

Follow-up from the SOC 2 review (Inj F3) and the natural "PR 5" after the 4-PR sprint (#799–#802).

## Changes

- `graphql/schema.py` — add `QueryDepthLimiter`, `MaxAliasesLimiter`, and `MaxTokensLimiter` to the Strawberry schema, passed as **factory callables** (per Strawberry's per-request construction contract; avoids the deprecated instance-passing form).
- `config/env.py` — three limits, **SSM-tunable at runtime** via `get_tuning_int` under `/tuning/graphql/` (same pattern as the admission/queue/MCP limits), so a limit can be tightened on abuse or loosened for a false-positive block without a code change or redeploy:
  - `EXTENSIONS_GRAPHQL_MAX_DEPTH` → `graphql/MAX_DEPTH` (default **15**)
  - `EXTENSIONS_GRAPHQL_MAX_ALIASES` → `graphql/MAX_ALIASES` (default **30**)
  - `EXTENSIONS_GRAPHQL_MAX_TOKENS` → `graphql/MAX_TOKENS` (default **2000**)

The default SSM params have been created in prod and staging at these paths.

## Why introspection still works

The schema deliberately allows anonymous introspection for SDK codegen. Verified empirically that the depth limiter **does not count introspection fields**, so introspection passes even well below the configured depth — SDK codegen is unaffected.

## Testing

- `uv run pytest tests/graphql/ tests/middleware/mcp/tools/test_graphql_tool.py` — **113 passed** (5 new), no regressions.
  - New `tests/graphql/extensions/test_schema_limits.py`: over-deep query rejected at the configured limit, in-bounds query passes, alias fan-out rejected, and — against the **real** schema via the recursive `AccountTreeNode.children` path — a deep query is rejected with `"exceeds maximum operation depth of 15"`, proving the limiter is actually wired in.

## Notes

- Values are read at startup (env class-body), like the other `get_tuning_*` params, so changing an SSM value takes effect on a service refresh — no code change / full redeploy.
- Field-level cost/complexity limiting (beyond depth/alias/token) is a possible future refinement; depth is the primary pool-exhaustion vector and is covered here.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -269,11 +269,11 @@ SECURITY_AUDIT_ENABLED=false
 # TUNING_QUEUES_TIMEOUT=300
 
 ## Extensions GraphQL query bounds — guard the OLTP pool from deep / alias-heavy
-## / oversized queries (introspection unaffected). Raw env-var names (read by
-## get_tuning_int); SSM: /tuning/graphql/{MAX_DEPTH,MAX_ALIASES,MAX_TOKENS}.
-# EXTENSIONS_GRAPHQL_MAX_DEPTH=15
-# EXTENSIONS_GRAPHQL_MAX_ALIASES=30
-# EXTENSIONS_GRAPHQL_MAX_TOKENS=2000
+## / oversized queries (introspection unaffected).
+## SSM: /tuning/graphql/{MAX_DEPTH,MAX_ALIASES,MAX_TOKENS}.
+# TUNING_GRAPHQL_MAX_DEPTH=15
+# TUNING_GRAPHQL_MAX_ALIASES=30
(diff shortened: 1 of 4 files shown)


jfrench9:

PR #805, opened 2026-07-02
fix(config): honor the documented TUNING_ env-var convention for tunables

### LINKED ISSUE #804: Tuning-section env-var docs use a TUNING_ prefix that is never read
## Problem

The `.env.example` **TUNING PARAMETERS** section documents runtime tunables with a `TUNING_{CATEGORY}_{KEY}` env-var prefix (e.g. `TUNING_LBUG_ADMISSION_MEMORY_THRESHOLD`, `TUNING_CACHE_BALANCE_TTL`), and the section header states `Env var name: TUNING_{CATEGORY}_{KEY}`.

But the `config/env.py` `get_tuning_int` / `get_tuning_float` helpers read the **raw** `env_key` via `os.getenv(env_key)` — there is no `TUNING_` handling. So those documented env vars are **never read** as a local/dev override.

### Verified

```
TUNING_LBUG_ADMISSION_MEMORY_THRESHOLD=99  -> env.LBUG_ADMISSION_MEMORY_THRESHOLD == 85.0   (default; ignored)
LBUG_ADMISSION_MEMORY_THRESHOLD=99         -> env.LBUG_ADMISSION_MEMORY_THRESHOLD == 99.0   (works)
```

The SSM path (`/robosystems/{env}/tuning/{category}/{KEY}`) is a separate priority-2 source and is unaffected — this is only about the **env-var override name** the docs advertise.

## Scope

Affects every `get_tuning_*`-backed param documented in the `.env.example` tuning section: `LIMITS`, `CACHE`, `DATABASE`, `ADMISSION`, `LBUG_ADMISSION`, `LOAD_SHEDDING`, `QUEUES`, `CIRCUIT`, `TIMEOUTS`, etc.

## Fix (pick one)

- **A (docs):** rewrite the `TUNING_*` example entries to the raw `env_key` names each `get_tuning_*` call actually reads, and fix the section header. (The new `EXTENSIONS_GRAPHQL_*` block added in #803 already uses the correct raw names as the reference pattern.)
- **B (code):** make `get_tuning_int/float` *also* honor a `TUNING_{path}` env var, so the documented convention becomes real. Touches every tuning param.

Surfaced during the #803 review (adding GraphQL query-bounding limits).

jfrench9:
## Summary

Makes the documented `TUNING_{CATEGORY}_{KEY}` env-var override **real** for `get_tuning_int` / `get_tuning_float`. Until now those helpers only read the raw `env_key`, so the `TUNING_*` names advertised in `.env.example` (and implemented in `config/tuning.py`) were never honored as an env-var override for any `get_tuning_*` param.

Closes #804.

## The bug (verified)

```
TUNING_LBUG_ADMISSION_MEMORY_THRESHOLD=99  ->  85.0  (default — ignored)
LBUG_ADMISSION_MEMORY_THRESHOLD=99         ->  99.0  (works)
```

`get_tuning_int/float` called `os.getenv(env_key)` on the raw name only. But the `.env.example` **TUNING PARAMETERS** section, its header (`Env var name: TUNING_{CATEGORY}_{KEY}`), and `config/tuning.py` all use the `TUNING_` prefix — so the documented override silently did nothing for admission, lbug_admission, org-graphs, cache, queues, etc.

## Change

`get_tuning_int` and `get_tuning_float` now resolve in priority order:

1. the raw `env_key` — backward compat, anything relying on it keeps working
2. `TUNING_{CATEGORY}_{KEY}` derived from the `ssm_path` (new `_tuning_env_key` helper, mirroring `config/tuning.py`)
3. SSM (`/tuning/...`, prod/staging only)
4. default

No behavior change for existing raw-name overrides; the documented `TUNING_*` convention now works too, so every entry already in `.env.example` becomes functional with **zero doc rewrite**.

## Testing

- `uv run pytest tests/config/test_env.py tests/config/test_tuning.py` — **75 passed (6 new)**: both names resolve, raw `env_key` wins when both are set, default fallback, float variant, and the `_tuning_env_key` derivation.

## Notes

- Surfaced during the #803 review (adding GraphQL query-bounding limits).
- Pairs with **#803**, whose `.env.example` block now uses `TUNING_GRAPHQL_*`. **Merge this before #803** so those documented names resolve as env overrides (SSM + the code defaults work regardless of merge order).

### DIFF
--- a/robosystems/config/env.py
+++ b/robosystems/config/env.py
@@ -174,6 +174,17 @@ def get_str_env(key: str, default: str = "") -> str:
   return os.getenv(key, default)
 
 
+def _tuning_env_key(ssm_path: str) -> str:
+  """Env-var name for a tuning param following the documented convention.
+
+  ``TUNING_{CATEGORY}_{KEY}`` derived from the SSM path — e.g.
+  ``"graphql/MAX_DEPTH"`` -> ``"TUNING_GRAPHQL_MAX_DEPTH"``. Mirrors the naming
+  in ``config/tuning.py`` and ``.env.example`` so the documented override
+  convention is actually honored (the raw ``env_key`` still works too).
+  """
+  return "TUNING_" + ssm_path.upper().replace("/", "_")
+
+
 def get_tuning_float(env_key: str, ssm_path: str, default: float) -> float:
   """
   Get a float tuning parameter with layered fallback.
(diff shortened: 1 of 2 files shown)


jfrench9:

PR #809, opened 2026-07-03
feat(security): emit security metrics and wire live detective-control alarms

jfrench9:
## Summary

Makes the API's security detective controls actually work. Today `FailedAdminAuthAlarm` in `api.yaml` watches a metric that is **never emitted** (the admin-key failure path only `logger.warning`s) and has **no `AlarmActions`**, so it can never fire and would notify no one if it did. Auth failures, injection attempts, and privilege-escalation events are logged but never turned into metrics or alarms. This PR emits those signals as CloudWatch metrics from the single `SecurityAuditLogger` chokepoint and wires them to live, email-backed alarms.

This is **P1** of the audit & observability work — the metrics + alarming slice. Audit-stream retention (the 13-month Type-II trail) and SSM session command-logging are deliberately left as follow-ups.

## Changes

**Metric emission (`robosystems/security/audit_logger.py`)**
- Publish the compliance-relevant subset of security events to `RoboSystems/Security/{env}` from `log_security_event`: `AuthFailure`, `AuthorizationDenied`, `InjectionAttempt` (incl. path-traversal), `PrivilegeEscalationAttempt`. Operational events (`email_sent`, `auth_success`, `operation_timeout`, `rate_limit_exceeded`) are intentionally excluded to keep the stream low-volume and the alarms meaningful.
- Emission is **best-effort and prod/staging-only** — no-op where there's no CloudWatch/credentials, and it never raises into the caller (a metrics failure must not break an auth path).
- Add `log_admin_auth_failure`, which flags the event as admin so a dedicated `FailedAdminAuth` metric is emitted alongside the generic `AuthFailure`.
- Chose `put_metric_data` over EMF: `StructuredFormatter` wraps every log record, so a root-level EMF `_aws` block would not survive the formatter; security events are low-volume, so the per-event API call cost is negligible.

**Admin failure path (`robosystems/middleware/auth/admin.py`)**
- Invalid admin-key attempts now go through `SecurityAuditLogger.log_admin_auth_failure` (previously only a `logger.warning`), so they are both audit-logged and feed `FailedAdminAuth`. Also hardened the client-IP read against a missing `request.client`.

**CloudFormation (`cloudformation/api.yaml`)**
- Add `SNSAlertEmail` param + `HasAlertEmail` condition + `SecurityAlertTopic` (SNS email subscription, created only when an alert email is supplied).
- Repoint `FailedAdminAuthAlarm` from the dead `RoboSystems/Admin` namespace to the live `RoboSystems/Security/${Environment}` and wire its `AlarmActions` to the topic.
- Add companion alarms → same topic: `AuthFailureSpikeAlarm`, `InjectionAttemptAlarm` (>0), `PrivilegeEscalationAlarm` (>0).
- Widen the task role's `cloudwatch:PutMetricData` condition to include the Security namespace.
- Alarms always exist; `AlarmActions` are conditional on `HasAlertEmail` so the stack still deploys cleanly if no email is set.

**Deploy wiring (`.github/workflows/deploy-api.yml`)**
- Thread `SNSAlertEmail` into the api stack params. The `aws_sns_alert_email` input already reaches this deploy (staging/prod pass `vars.AWS_SNS_ALERT_EMAIL`), so no changes to `staging.yml`/`prod.yml` were needed.

**Tests (`tests/security/test_audit_metrics.py`)**
- 16 unit tests: event→metric mapping (incl. admin flag → `FailedAdminAuth`, operational events → no metric), the prod/staging guard, best-effort swallow-on-failure, and the admin middleware failure path both logging and raising 401.

## Testing

- `uv run pytest tests/security/test_audit_metrics.py` — 16 passed.
- `uv run pytest tests/security/ tests/middleware/auth/ tests/routers/admin/` — 721 passed (no regressions).
- `ruff check` / `ruff format --check` / `basedpyright` on changed files — clean. Pre-commit hooks (ruff + format + basedpyright over the repo) passed on commit.
- `cfn-lint -t cloudformation/api.yaml` — clean. (`aws cloudformation validate-template` not run — SSO token expired locally; CI runs it.)

## Notes / Follow-ups

- **P2 — audit-stream retention**: route the audit-event subset to a durable ≥13-month sink (dedicated log group + a compliance retention tier), keeping operational INFO short. This is the actual SOC 2 Type-II evidence trail.
- **SSM session command-logging** (RC-6 residual): cross-stack (account-level `SSM-SessionManagerRunShell` doc + instance-role IAM on the graph/bastion stacks) — deferred to its own change.
- **Behavioral note**: `SNSAlertEmail` is currently set to the same value as `EmailFromAddress` (both from `aws_sns_alert_email`), matching the existing deploy wiring. If a dedicated ops/security distribution list is preferred, that's a one-line workflow change later.

### DIFF
--- a/.github/workflows/deploy-api.yml
+++ b/.github/workflows/deploy-api.yml
@@ -333,6 +333,7 @@ jobs:
                 ParameterKey=FargateBase,ParameterValue=${{ inputs.fargate_base }} \
                 ParameterKey=PrometheusStackName,ParameterValue=\"${{ inputs.prometheus_stack_name }}\" \
                 ParameterKey=EmailFromAddress,ParameterValue=\"${{ inputs.aws_sns_alert_email }}\" \
+                ParameterKey=SNSAlertEmail,ParameterValue=\"${{ inputs.aws_sns_alert_email }}\" \
                 ParameterKey=SharedProcessedBucketArn,ParameterValue=\"${{ inputs.shared_processed_bucket_arn }}\" \
                 ParameterKey=DeploymentBucketArn,ParameterValue=\"${{ inputs.deployment_bucket_arn }}\" \
                 ParameterKey=UserDataBucketArn,ParameterValue=\"${{ inputs.user_data_bucket_arn }}\" \
--- a/cloudformation/api.yaml
+++ b/cloudformation/api.yaml
@@ -8,47 +8,36 @@ Parameters:
     AllowedValues:
       - prod
       - staging
-    Description: Environment name (prod or staging)
-    ConstraintDescription: Must be prod or staging
 
   # Infrastructure & Networking
   VpcId:
     Type: AWS::EC2::VPC::Id
-    Description: VPC ID for ECS tasks
   VpcCidr:
     Type: String
     Default: 10.0.0.0/16
-    Description: CIDR block for the VPC
     AllowedPattern: "^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\\/(1[6-9]|2[0-8]))$"
-    ConstraintDescription: Must be a valid CIDR block
   SubnetIds:
     Type: List<AWS::EC2::Subnet::Id>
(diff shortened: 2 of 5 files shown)


jfrench9:
@claude please review this PR

PR #810, opened 2026-07-03
feat(audit): retain security-audit logs to S3 + SSM session command logging

jfrench9:
## Summary

P2 of the audit-observability spec plus the RC-6 SSM residual. Compliance-relevant audit records currently live in the 7-day `/robosystems/{env}/api` log group, commingled with operational INFO — there's no long-lived, separate evidence trail. And admin bastion sessions record *who connected* (CloudTrail) but not *what ran*. This PR adds both, **infra-only and gated OFF by default** (no application code, nothing on the request path).

## Changes

**Audit-log retention — `cloudformation/audit.yaml` (new, modeled on `waf.yaml`)**
- A CloudWatch Logs **subscription filter** on `/robosystems/${Environment}/api` forwards ONLY audit records — filter pattern `?SECURITY_AUDIT ?"\"audit\":"` OR-matches the two audit streams (`SecurityAuditLogger`'s `SECURITY_AUDIT:` messages and the operation-audit records that carry a structured `"audit":` key). Operational INFO stays in the short-retention group.
- Filter → **Kinesis Firehose** (`DirectPut`, GZIP, 5 MB / 300 s buffering) → a dedicated **S3 audit bucket**: versioned, all-public-access-blocked, AES256, `INTELLIGENT_TIERING`, with a **13-month lifecycle** (`RetentionDays` param, default 400). Two IAM roles (Firehose→S3, CloudWatch-Logs→Firehose) and a short-retention Firehose error log group round it out.
- Chose this log-side forward over the spec's tentative "dedicated log group + in-app shipper": the ECS `awslogs` driver ships all stdout to one group, so that option would mean an in-process CloudWatch-Logs writer — new app-runtime surface and the event-loop risk the P1 review flagged. This keeps everything off the request path.

**Deploy wiring — gated `audit` job (no new workflow)**
- `deploy-api.yml`: three new optional inputs (`audit_enabled`, `audit_stack_name`, `audit_retention_days`) and a new `audit` job that mirrors the existing `waf` job (`needs: deploy` so the API log group exists before the filter attaches; create/update/wait with the standard failure diagnostics).
- `stacks.yml`: `audit` stack entries in both env blocks (`RoboSystemsAudit{Prod,Staging}`), auto-resolved by the generic `stack-config.sh`.
- `staging.yml` / `prod.yml`: `audit_stack` stack-config output + echo, and `audit_enabled`/`audit_stack_name`/`audit_retention_days` passed to the api job — identical pattern to WAF. Toggle: `AUDIT_ENABLED_{PROD,STAGING}` (default `false`).

**SSM session command logging — `cloudformation/cloudtrail.yaml`**
- Adds an account-global `SSM-SessionManagerRunShell` (`AWS::SSM::Document`, `DocumentType: Session`) that streams shell commands to a new **400-day `/robosystems/ssm-sessions`** CloudWatch log group. Placed in the single shared `RoboSystemsCloudTrail` stack because the document name is account/region-global (the per-env bastion stacks would collide on it). The bastion instance role already carries `logs:CreateLogStream/PutLogEvents`, so no IAM change. Both resources are gated on the existing `CreateCloudTrail` condition.

## Testing

- `cfn-lint -t cloudformation/*.yaml` — clean (CI parity).
- `pytest tests/infrastructure/test_cloudformation.py` — **20 passed** (the new `audit.yaml` is auto-covered by the parametrized size/validity checks; `api.yaml` is untouched, so its inline-limit test still passes).
- `stack-config.sh {prod,staging} audit` → `RoboSystemsAudit{Prod,Staging}`.
- All three workflows + `stacks.yml` parse as YAML.
- No Python changed, so no unit tests apply. `aws cloudformation validate-template` and live deploy were **not run** (no AWS session in this environment) — CI runs cfn-lint; the items below are first-deploy-verify.

## Notes / Follow-ups

- **Rollout is opt-in.** Audit retention stays off until `AUDIT_ENABLED_{PROD,STAGING}` is set to `true`, then a deploy creates the stack.
- **SSM logging deploys with the CloudTrail stack** on the next staging/prod run (`CLOUDTRAIL_ENABLED` is already `true`) — it is not behind a separate toggle. **First-deploy risk:** if a console-created `SSM-SessionManagerRunShell` already exists in the account, the CFN create collides — import or rename if so.
- **First-deploy-verify (data, easily adjusted):** the subscription-filter pattern (tune if it under/over-matches); and CWL→Firehose delivers gzipped CloudWatch-Logs-format records, so S3 objects are double-gzip — a Firehose processing Lambda to emit clean JSON is a deferred P3 nicety, along with S3 Object Lock/WORM.

### DIFF
--- a/.github/configs/stacks.yml
+++ b/.github/configs/stacks.yml
@@ -122,6 +122,10 @@ prod:
       web_acl_id: WebACLId
       web_acl_arn: WebACLArn
 
+  # Audit Trail Retention (compliance log forwarding)
+  audit:
+    stack_name: RoboSystemsAuditProd
+
   # GitHub Actions Runner
   gha_runner:
     stack_name: RoboSystemsGHARunner # Shared between environments
@@ -267,6 +271,10 @@ staging:
     outputs:
       web_acl_id: WebACLId
       web_acl_arn: WebACLArn
+
+  # Audit Trail Retention (compliance log forwarding)
+  audit:
+    stack_name: RoboSystemsAuditStaging
 
   # Shared Resources (reference same as prod)
   gha_runner:
--- a/.github/workflows/deploy-api.yml
+++ b/.github/workflows/deploy-api.yml
@@ -221,6 +221,23 @@ on:
         type: string
         default: "true"
 
+      # Audit Trail Retention (optional)
+      audit_enabled:
+        description: "Enable long-retention forwarding of audit logs to S3"
+        required: false
+        type: string
+        default: "false"
+      audit_stack_name:
+        description: "CloudFormation stack name for the audit stack"
+        required: false
+        type: string
+        default: ""
+      audit_retention_days:
+        description: "Days to retain audit records in S3 (~13 months = 400)"
+        required: false
+        type: string
+        default: "400"
+
       # Notification Configuration
       aws_sns_alert_email:
         description: "Email address for SNS alerts and notifications"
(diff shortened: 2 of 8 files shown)


jfrench9:
@claude please review this PR

PR #812, opened 2026-07-03
feat(security): add off-by-default account security baseline stack

jfrench9:
## Summary

Adds an **off-by-default** account-global security baseline stack — the AWS detective controls (GuardDuty, Security Hub, IAM Access Analyzer, optional AWS Config, and Amazon Inspector) that were the biggest remaining SOC 2 readiness gap. It's modeled on the existing CloudTrail pattern: one shared stack, deployed by a gated job in the VPC workflow, toggled by a GitHub variable. Nothing turns on until `SECURITY_ENABLED=true` — the code, the repo variables, and the workflow defaults are all off.

## Changes

**`cloudformation/security.yaml`** (new — shared, account-global stack `RoboSystemsSecurity`)
- **GuardDuty** detector, **Security Hub** + the FSBP standard (optional CIS), and **IAM Access Analyzer** (account scope) — enabled when the stack deploys.
- **AWS Config** recorder + dedicated S3 bucket, delivery role, and delivery channel — behind its own `EnableConfig` parameter (default off, `DAILY` recording) because it's the cost outlier.
- **Amazon Inspector v2** has no native single-account CloudFormation resource, so it's enabled by an idempotent `inspector2:enable` step in the deploy job rather than a stack resource.
- Every service is gated by its own `Enable*` parameter; the Config S3 bucket is `Retain`/versioned/encrypted/public-blocked.

**`.github/workflows/deploy-vpc.yml`**
- New `security` job — a sibling to the existing `cloudtrail` job (`needs: deploy`, `if: security_enabled == 'true'`), so the two toggle independently. Deploys the stack, monitors it, then runs the Inspector enable step.
- New workflow inputs: `security_enabled`, `security_stack_name`, `security_config_enabled`.

**`.github/workflows/staging.yml` / `prod.yml`**
- Wire the `security_stack` name through stack-config and feed `SECURITY_ENABLED` / `SECURITY_CONFIG_ENABLED` variables into the VPC deploy call.

**`.github/configs/stacks.yml`**
- `security: RoboSystemsSecurity` (shared between environments) in both the prod and staging blocks.

**`cloudformation/bootstrap-oidc.yaml`** (deploy-role `RoboSystemsDeploymentPolicy`)
- New service grants: `guardduty:*`, `securityhub:*`, `config:*`, `inspector2:*`, `access-analyzer:*`.
- Added `config.amazonaws.com` to the `iam:PassRole` allowlist (for the Config recorder role).
- These pre-empt the same first-deploy IAM failure the audit stack hit with Firehose (#811), so enabling the stack later can't fail on missing permissions. (Service-linked-role creation and role management were already covered.)

**`bin/setup/gha.sh` / `bin/setup/README.md`**
- Seed `SECURITY_ENABLED` and `SECURITY_CONFIG_ENABLED` (both `false`) and document them.

**`SECURITY.md`**
- Document the new security baseline under Compliance Infrastructure (parameters, log location, validation commands).
- Catch up prior undocumented work: the full security-alarm set + `RoboSystems/Security/{env}` metric namespace + SNS topic (#809), the audit-retention pipeline (#810 `audit.yaml`), and SSM Session Manager command logging (#810 `cloudtrail.yaml`).

**Follow-up commits (from self- and review passes)**
- `fix`: order the Config delivery channel after the recorder (avoids a "configuration recorder is not available" failure when Config is enabled).
- `docs`: tag the Security Hub Hub resource; document that the workflow wires only `Environment` + `EnableConfig`, and the account-global "already enabled out-of-band" precondition.

## Testing

- `just cf-lint security` and `just cf-lint bootstrap-oidc` — clean (no warnings/errors).
- `uv run pytest tests/infrastructure/test_cloudformation.py` — 21 passed (the size-limit guard globs all templates; `security.yaml` ~8 KB, `bootstrap-oidc.yaml` still well under the 51,200-byte inline ceiling).
- Workflow + config YAML parse-checked (`deploy-vpc.yml`, `staging.yml`, `prod.yml`, `stacks.yml`).
- `stack-config.sh` resolves `security` → `RoboSystemsSecurity` for both `production` and `staging`.
- No Python/application code changed, so the unit suite is not relevant here; `just test-all` was not run.

## Notes / Follow-ups

- **Off by default, end to end.** The `SECURITY_ENABLED` / `SECURITY_CONFIG_ENABLED` repo variables are set to `false`; the workflow reads `${{ vars.SECURITY_ENABLED || 'false' }}`.
- **Before enabling:** re-run `just bootstrap` so the live deploy role picks up the new grants (same step the audit stack needed).
- **Inspector asymmetry:** enabled automatically by the deploy job, but because it isn't a stack resource, deleting the stack won't disable it — that's a manual `aws inspector2 disable`. Left as an MVP tradeoff over wrapping it in a custom-resource Lambda.
- The Config S3 bucket uses `DeletionPolicy: Retain` (same teardown consideration as the CloudTrail/audit buckets). cfn-guard advisories on the Config bucket (SSL-only policy, access logging) match the existing `audit.yaml`/`cloudtrail.yaml` convention and are outside the enforced `cf-lint`.

### DIFF
--- a/.github/configs/stacks.yml
+++ b/.github/configs/stacks.yml
@@ -146,6 +146,10 @@ prod:
       bucket_name: CloudTrailBucketName
       enabled: CloudTrailEnabled
 
+  # Security Baseline (GuardDuty, Security Hub, Config, Access Analyzer)
+  security:
+    stack_name: RoboSystemsSecurity # Shared between environments (account-global)
+
   # Dagster Orchestration
   dagster:
     stack_name: RoboSystemsDagsterProd
@@ -294,6 +298,10 @@ staging:
       trail_arn: CloudTrailArn
       bucket_name: CloudTrailBucketName
       enabled: CloudTrailEnabled
+
+  # Security Baseline (GuardDuty, Security Hub, Config, Access Analyzer)
+  security:
+    stack_name: RoboSystemsSecurity # Shared between environments (account-global)
 
   # Dagster Orchestration
   dagster:
--- a/.github/workflows/deploy-vpc.yml
+++ b/.github/workflows/deploy-vpc.yml
@@ -88,6 +88,23 @@ on:
         type: string
         default: ""
 
+      # Security Baseline Configuration (account-global detective controls, SOC 2)
+      security_enabled:
+        description: "Enable the account-global security baseline (GuardDuty, Security Hub, Access Analyzer, optional Config + Inspector)"
+        required: false
+        type: string
+        default: "false"
+      security_stack_name:
+        description: "Security baseline CloudFormation stack name"
+        required: false
+        type: string
+        default: "RoboSystemsSecurity"
+      security_config_enabled:
+        description: "Enable the AWS Config recorder inside the security stack (cost scales with resource count)"
+        required: false
+        type: string
+        default: "false"
+
     outputs:
       vpc_id:
         description: "VPC ID"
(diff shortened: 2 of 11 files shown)


jfrench9:
@claude can you review

PR #820, opened 2026-07-04
chore(opensearch): parameterize cluster scaling and Multi-AZ as inert GH-var levers

jfrench9:
## Summary

Makes OpenSearch node count, zone-awareness, and dedicated masters GitHub-variable-driven so the single→Multi-AZ transition becomes a **variable flip** rather than a template change. Every new parameter defaults to today's exact single-node domain, so the resolved CloudFormation template is unchanged — an inert no-op today, with the levers live to control later. No application behavior changes; no Python touched.

## Context

The prod OpenSearch domain is single-node / single-AZ / 0-replica. Only *vertical* scaling (instance type, EBS, version) was GH-var-driven; node count and Multi-AZ were hardcoded in the template, so "flip a var to go Multi-AZ" was not true. This closes that gap **without changing the running domain** — prod stays 1 instance / no Multi-AZ. When Multi-AZ is later needed, it becomes a var flip + a `put_settings` + a 2nd RI, not a rebuild.

## Changes

**`cloudformation/opensearch.yaml`**
- `SubnetId` (single `AWS::EC2::Subnet::Id`) → `SubnetIds` (`List<AWS::EC2::Subnet::Id>`); `VPCOptions.SubnetIds` is now `!Ref SubnetIds`. Default resolves to the same single first subnet.
- **Lever 1 (capacity/AZ):** new `InstanceCount`, `ZoneAwarenessEnabled`, `AvailabilityZoneCount` params; `ClusterConfig` `InstanceCount` / `ZoneAwarenessEnabled` / `ZoneAwarenessConfig` are driven by them (via an `IsZoneAware` condition), defaulting to `1` / `false` / absent — identical to the prior hardcoded values.
- **Lever 2 (HA/quorum):** new `DedicatedMasterEnabled` (+ `DedicatedMasterType` / `DedicatedMasterCount`), gated behind a `HasDedicatedMaster` condition; renders the same hardcoded `DedicatedMasterEnabled: false` by default. Masters are a control-plane role (no data, no reads/writes), hence a separate/smaller instance type.
- `EBSVolumeSize` `MaxValue` `1000` → `2048`.

**`.github/workflows/deploy-opensearch.yml`**
- Renamed input `subnet_id` → `subnet_ids`; added `instance_count` / `zone_awareness_enabled` / `availability_zone_count` / `dedicated_master_*` inputs (single-node defaults).
- Subnet selection: single-AZ uses the first subnet (byte-identical to the prior `cut -f1`); the Multi-AZ branch passes the first N subnets with comma-escaping for the CFN list param.
- Threads the new params into the stack deploy call.

**`.github/workflows/{prod,staging}.yml`**
- Wired the new inputs to `OPENSEARCH_*_{PROD,STAGING}` vars with single-node fallbacks; renamed `subnet_id` → `subnet_ids` (now consistent with every other stack, which already used plural `subnet_ids`).

**`bin/setup/gha.sh`**
- Seeds only the minimal single-node switches (`OPENSEARCH_INSTANCE_COUNT=1`, `OPENSEARCH_ZONE_AWARENESS=false`) per env; the AZ-count and master knobs ride the workflow `||` defaults until rollout.

## Testing

- `cfn-lint cloudformation/opensearch.yaml` — **pass** (exit 0)
- `aws cloudformation validate-template` — **pass** (exit 0)
- `just test-code` (ruff + format + basedpyright + cf-lint) — **pass**; no Python changed.
- `just test` (unit suite) — **Not run** (no Python touched, no behavior change).

## Notes / Follow-ups

- **Inertness verification gate:** the "no-op" claim is verified statically (defaults render identical values), but the load-bearing confirmation is a CloudFormation **change-set showing an empty diff** against the live prod stack. That can't run without a write to prod, so it happens at the first prod deploy carrying this branch. If the diff isn't empty, the `SubnetId → SubnetIds` type change is the suspect.
- **Prod vars already set:** `OPENSEARCH_INSTANCE_COUNT_PROD=1`, `OPENSEARCH_ZONE_AWARENESS_PROD=false` (keeps prod exactly as today).
- **Replica count intentionally excluded** — it's a runtime `put_settings` on the live index at rollout, not app config (an env var would only apply at index creation and would need threading through both index-creators).
- **To go Multi-AZ later** (all var flips, no code): `OPENSEARCH_INSTANCE_COUNT_PROD=2` + `OPENSEARCH_ZONE_AWARENESS_PROD=true` + `OPENSEARCH_DEDICATED_MASTER_PROD=true` → deploy → `put_settings` replicas=1 → buy the 2nd RI.

### DIFF
--- a/.github/workflows/deploy-opensearch.yml
+++ b/.github/workflows/deploy-opensearch.yml
@@ -29,8 +29,8 @@ on:
         description: "VPC ID for OpenSearch"
         required: true
         type: string
-      subnet_id:
-        description: "Private subnet ID for OpenSearch (single-AZ)"
+      subnet_ids:
+        description: "Comma-separated private subnet IDs (first is used for single-AZ; first N for Multi-AZ)"
         required: true
         type: string
 
@@ -40,6 +40,36 @@ on:
         required: false
         type: string
         default: "t3.small.search"
+      instance_count:
+        description: "Number of data nodes (1 = single-AZ; >=2 for Multi-AZ)"
+        required: false
+        type: string
+        default: "1"
+      zone_awareness_enabled:
+        description: "Distribute data nodes across AZs (requires instance_count >= 2)"
+        required: false
+        type: string
+        default: "false"
+      availability_zone_count:
+        description: "Number of AZs to spread data nodes across when zone awareness enabled (2 or 3)"
+        required: false
+        type: string
+        default: "2"
+      dedicated_master_enabled:
+        description: "Enable dedicated master nodes (independent HA/quorum lever; recommended at Multi-AZ rollout)"
+        required: false
+        type: string
+        default: "false"
+      dedicated_master_type:
+        description: "Instance type for dedicated masters (used only when enabled; small is fine)"
+        required: false
+        type: string
+        default: "t3.small.search"
+      dedicated_master_count:
+        description: "Number of dedicated masters (used only when enabled; 3 recommended, odd)"
+        required: false
+        type: string
+        default: "3"
       ebs_volume_size:
         description: "EBS volume size in GB"
         required: false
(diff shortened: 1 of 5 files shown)


jfrench9:
@claude please review this PR

PR #824, opened 2026-07-05
feat(extensions): materialize is_live flag on ledger-spine nodes

jfrench9:
## Summary

The extensions materialization mirrors the **full** ledger spine (`Event`, `Transaction`, `Entry`, `LineItem`) into the graph — including cancelled/replaced rows — which is correct for audit history but leaves the **read side** exposed: any AI-generated or ad-hoc Cypher that counts/sums the spine without a status filter silently includes voided/reversed rows and inflates results. The correct per-node filter was non-uniform and easy to forget, and `Transaction` was worse — the `pending`-only projection dropped the full status, making a voided transaction indistinguishable at the node.

This collapses that to **one rule — `WHERE n.is_live`** — by materializing a computed `is_live` boolean onto every spine node, and flips the existing Tier 1 read guidance to lead with it. Additive and safe: no OLTP migration, the graph is a rebuildable projection, and existing `status` filters keep working unchanged.

## Changes

**Materializer** (`operations/extensions/materialize.py`) — computed `is_live` on all four spine tables:
- `Event` → `status NOT IN ('voided','superseded')`
- `Transaction` → `status <> 'void'` (the `pending` boolean is kept for back-compat)
- `Entry` → `status = 'posted'`
- `LineItem` → denormalized parent `Entry.is_live` via an **inner join to `entries`** (`entry_id` is `NOT NULL`, so the join drops no rows). Lets balance rollups filter with `WHERE li.is_live` without a hop back to Entry.

**Schema** — `Property(name="is_live", type="BOOLEAN")` added to `Event` (`schemas/base.py`) and `Transaction`/`Entry`/`LineItem` (`schemas/extensions/roboledger.py`) so the flag surfaces in `get-graph-schema`. The graph loader reconciles source→target columns **by name** and already supports a property added mid-schema, so the new column aligns safely on rebuild.

**Read guidance flipped to lead with `is_live`** (keeping the per-node `status` filters as the "why"):
- `LEDGER_STATUS_GUIDANCE` (`middleware/mcp/tools/constants.py`)
- The three ledger example queries (`middleware/mcp/tools/example_queries_tool.py`)
- The `CypherOperator` system prompt (`operations/operators/implementations/cypher.py`)

**Tests** (`tests/operations/extensions/test_materialize.py`) — two new cases asserting `is_live` on all four tables and the LineItem entry-join.

## Design notes

- **Why a flag on all four nodes and not just Transaction?** Liveness does **not** cascade top-down. A reversal (`journal_entries.py:519`) marks the original `Entry.status='reversed'` and posts a *new* `posted` reversing entry without touching the parent Transaction — so a live Transaction routinely holds a `reversed` + a `posted` entry at once. Filtering only at Transaction would double-count; you must filter at Entry/LineItem. Ad-hoc queries also anchor at different nodes (balance rollups start at `LineItem → Element` and never touch Transaction), so each node needs to be independently filterable.
- **`Transaction.is_live` is uniformity insurance, not load-bearing.** `Transaction.status ∈ (pending, posted, void)`, but nothing in the codebase currently sets `'void'` (voiding happens at the Event/Entry layers). It's kept so the "one rule, every node" contract holds — dropping it reintroduces the non-uniform footgun.

## Testing

- `just test` — **full unit suite passed**: 10,679 passed, 15 skipped, 101 deselected (~7.5 min).
- `just test-code` — **passed clean**: ruff, format, basedpyright, cf-lint (0 errors/warnings).
- Pre-commit hooks passed on commit.

## Follow-ups / deployment note

- `is_live` only appears in a graph **after re-materialization** (blue/green rebuild or forced `materialize`). Because the guidance flip ships in the same branch, there's a window where guidance advises `WHERE n.is_live` but a not-yet-rebuilt graph lacks the column (query returns empty / binder error). Low-risk pre-launch (demo graphs are freshly materialized; few/no prod tenant graphs), but the clean rollout is **deploy → rematerialize existing tenant graphs → guidance is correct**. The staleness sensor rebuilds them on the next OLTP write regardless.
- Deliberately **skipped** the optional Phase 4 query-validator nudge (warning-only checks on AI Cypher — low value, added surface).

@claude review

### DIFF
--- a/robosystems/middleware/mcp/tools/constants.py
+++ b/robosystems/middleware/mcp/tools/constants.py
@@ -28,21 +28,28 @@ Note: Element.period_type indicates the expected period type for that metric - d
 # tenant roboledger ledger spine (Event / Transaction / Entry / LineItem). The
 # graph mirrors the FULL ledger including cancelled/replaced rows; readers must
 # filter to live rows or voided/reversed amounts inflate counts and sums.
-LEDGER_STATUS_GUIDANCE = """**⚠️ LEDGER STATUS FILTERING (Event / Entry / Transaction):**
+LEDGER_STATUS_GUIDANCE = """**⚠️ LEDGER STATUS FILTERING (Event / Entry / Transaction / LineItem):**
 The graph is a faithful mirror of the ledger and KEEPS cancelled and replaced rows —
-voided and superseded entries are NOT removed (they are real audit history). When you
-COUNT or AGGREGATE ledger-spine data you MUST filter to live rows, or voided/reversed
-amounts will inflate the result:
-- `Entry.status` ∈ {draft, posted, reversed}. For balances and debit/credit sums,
-  match ONLY `e.status = 'posted'`. `draft` = unposted (includes entries belonging to
-  voided events); `reversed` = superseded by a reversing entry.
-- `Event.status` ∈ {captured, classified, committed, pending, fulfilled, voided,
-  superseded}. EXCLUDE `voided` and `superseded` from counts/sums. For open
-  obligations use the positive set the question implies (e.g. committed/fulfilled/pending).
-- `Transaction` exposes only a `pending` boolean in the graph (NOT the full status),
-  so a voided transaction is indistinguishable at the Transaction node. To measure
-  realized economic effect, aggregate through `Entry`/`LineItem` filtered to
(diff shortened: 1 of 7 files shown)


jfrench9:
@claude please review this PR

PR #825, opened 2026-07-05
fix(extensions): Entry.provenance — expose on graph + permit event_handler in CHECK

jfrench9:
## Summary

Follow-up to #824, covering two related `Entry.provenance` fixes found while wiring the graph side:

1. **Expose `Entry.provenance` on the graph** — the materializer has carried the column since `b282c036`, but the `Entry` node schema never declared a matching `Property`, so the by-name graph loader silently dropped it. It was never queryable via `get-graph-schema` or Cypher.
2. **Fix a latent production bug the first PR uncovered** — the event-block engine writes `provenance='event_handler'`, a value the `ck_entries_provenance` CHECK constraint did not permit, so the first real event-handler-driven GL insert would `CheckViolation`.

### Expose provenance on the graph
- **`schemas/extensions/roboledger.py`** — add `Property(name="provenance", type="STRING")` to the `Entry` node (placed after `reversal_of`, matching the staging SELECT order).
- **`tests/operations/extensions/test_materialize.py`** — regression test asserting `Entry.provenance` is both materialized and declared in the schema. The materializer-vs-schema validator only checks drift at the **table** level, not per-column, which is why the gap persisted.

### Permit `event_handler` in the provenance CHECK
- **`models/extensions/roboledger/entry.py`** — add `ENTRY_PROVENANCE_VALUES` as the single source of truth (the 5 original values + `event_handler`) and declare the `ck_entries_provenance` `CheckConstraint` **on the model** — it previously existed only in migration `0001`, so model and DB had drifted.
- **`migrations/extensions/versions/0019_widen_entry_provenance_check.py`** — widens the live constraint across the `public` schema and every existing tenant schema via the `TenantOps` fan-out (matches `0018`). Fresh tenants get it from the model at provision. *(Autogenerate produced only spurious schema-per-tenant FK/index drift and did not detect the CHECK change, so the migration is hand-written — the autogenerated file was discarded.)*
- **`tests/operations/event_block/test_engine.py`** — regression assertion that the value the engine writes (`event_handler`) is in `ENTRY_PROVENANCE_VALUES`. Would have failed pre-fix. (The engine test uses a mocked session that never flushes, which is why the constraint violation went unnoticed — no real-DB fixture exists for this area, so the invariant is asserted at the model level instead.)

## Testing

- `just test` (targeted): `test_engine.py`, `test_materialize.py`, `tests/migrations`, `tests/operations/event_block`, `tests/schemas`, `tests/operations/roboledger/commands` — **all passed** (452 + 42 across the runs).
- `just test-code` — **passed clean** (ruff, format, basedpyright, cf-lint; 0 errors/warnings).
- Migration applied locally (`0019` is head) and the live `ck_entries_provenance` on the `public` schema verified to now include `event_handler`.

## Notes

- `provenance` appears on a graph after its next re-materialization (same blue/green mechanism as #824's `is_live`).
- Migration `0019` `downgrade()` re-narrows the constraint and will fail if any `provenance='event_handler'` rows already exist — expected once the value is in use.
- Production rollout needs the extensions migration applied (`just migrate-up extensions` via the prod tunnel) before/with the deploy, since the code path writes `event_handler`.

### DIFF
--- a/migrations/extensions/versions/0019_widen_entry_provenance_check.py
+++ b/migrations/extensions/versions/0019_widen_entry_provenance_check.py
@@ -1,0 +1,64 @@
+"""widen entry provenance CHECK to include event_handler
+
+The ``ck_entries_provenance`` CHECK shipped in 0001 permitting
+{source_sync, ai_generated, manual_entry, schedule_derived, system_computed}.
+The event-block engine (``operations/event_block/engine.py``) creates entries
+with ``provenance='event_handler'`` — a value the constraint rejected, so the
+first real event-handler-driven GL insert would fail with a CheckViolation.
+Widen the constraint to permit ``event_handler`` (kept in lockstep with
+``ENTRY_PROVENANCE_VALUES`` on the model).
+
+Public schema AND every existing tenant schema get the widened constraint here
+via the TenantOps fan-out (matches 0018); fresh tenants get it from the model's
+CheckConstraint at provision (``metadata.create_all``).
+
+Note: the downgrade re-narrows the constraint and will fail if any
+``provenance='event_handler'`` rows already exist (they would violate the
+narrower CHECK) — expected once the value is in use.
+
+Revision ID: 0019
+Revises: 0018
+Create Date: 2026-07-04
+
+"""
+
+from __future__ import annotations
+
+from alembic import op
+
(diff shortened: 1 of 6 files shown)


jfrench9:
@claude please review this PR

PR #830, opened 2026-07-07
fix(rate-limits): correct graph endpoint categorization and close coverage gaps

jfrench9:
## Summary

Fixes a rate-limiting bug where read-only SEC (shared-repo) queries reported
`Rate limit exceeded for graph write operations.`, and cleans up the wider
graph rate-limiting surface it exposed. The categorizer matched several routes
against a stale `endpoint_type == "graph"` condition that never corresponds to a
real route (paths are `/v1/graphs/{id}/query`, not `/.../graph/query`), so
direct Cypher queries fell through to the generic write default. While in here,
the change corrects mis-bucketed routes, gives expensive lifecycle operations
dedicated limits, removes a redundant limiter layer, and closes endpoints that
had no rate limiting at all.

## Changes

**Endpoint categorization** (`config/rate_limits.py`)
- `/query` → `GRAPH_QUERY` (was `GRAPH_WRITE` via the POST catch-all — the reported bug)
- `/analytics` → `GRAPH_ANALYTICS`; `/schema*` → `GRAPH_READ` (validate was mis-bucketed as write)
- `connections`: reads → `GRAPH_READ`, writes → `GRAPH_WRITE`, and only the `/sync` trigger stays `GRAPH_SYNC` (previously *every* connections route got the tight 3/min sync bucket)
- graph lifecycle ops (`/operations/{op}`) get dedicated buckets: `create-`/`restore-backup` → `GRAPH_BACKUP`, `materialize` → `GRAPH_IMPORT`, and a new `GRAPH_MANAGEMENT` category for subgraph/tier/style ops
- removed the dead `get_rate_limit_with_multiplier` (never called in the live path — all authenticated users are classified `ladybug-standard`) and fixed the comments/README that claimed a tier multiplier scaled enforcement

**Shared-repo limiter** (`middleware/rate_limits/repository_rate_limits.py` + query/mcp/search callers)
- collapsed the `DualLayerRateLimiter` to a single subscription-plan *volume* layer; the per-request *burst* limit it re-derived duplicated the upstream subscription dependency (same category, tier, and 60/min) under a separate counter
- dropped `_check_burst_limit` / `_operation_to_category` / the `user_tier` plumbing and the now-dead `burst_limit` branches in the callers
- MCP path now gates on `is_shared_repository_or_subgraph` (was exact-parent `is_shared_repository`) and resolves subgraph → parent for the subscription lookup, so `sec_historical` gets the SEC per-plan limits like the query path (authorization already covered subgraphs via `validate_mcp_access`)
- fixed `get_usage_stats` hour bucket: the read key used `window[:3]` (`"hou"`) but writes use `"hour"`, so hour usage always read 0

**Coverage gaps closed** (search / documents / subgraphs routers)
- `/search` now applies the per-tier burst limiter (`GRAPH_SEARCH`) on all graphs — previously only shared repos were limited, so user-graph search was unbounded
- `documents/*` and `subgraphs/*` had no rate-limit dependency at all; attached the subscription limiter at the router level

## Testing

Run this session, all green:
- `just test` — full unit suite: **10,706 passed, 15 skipped, 0 failed**
- `just test-code` — ruff + format (1541 files) + basedpyright (0 errors) + cf-lint: **all passed**
- Verified the categorizer output for ~25 real routes and confirmed the router-level dependency attaches to every newly-wired search/documents/subgraphs route

## Notes / Follow-ups (not in this PR)

- `/v1/operations/{id}/status` and `DELETE` still use the generic limiter rather than a subscription bucket — intentional, since forcing a tight bucket would throttle status polling.
- Cosmetic dead code remains: `AllowedSharedEndpoints` / `BLOCKED_SHARED_ENDPOINTS` disagree with the SEC manifest, and `SharedRepositoryRateLimits.is_endpoint_allowed` ignores its `repository` arg (latent only — single shared repo today).
- `limits.py` still *displays* multiplier-based numbers that don't match the now-flat enforcement; left as a separate display-vs-reality question.

### DIFF
--- a/robosystems/config/README.md
+++ b/robosystems/config/README.md
@@ -209,9 +209,10 @@ class EndpointCategory(str, Enum):
     # Graph-scoped endpoints
     GRAPH_READ = "graph_read"              # GET operations
     GRAPH_WRITE = "graph_write"            # POST/PUT/DELETE
-    GRAPH_ANALYTICS = "graph_analytics"    # Heavy computations
-    GRAPH_BACKUP = "graph_backup"
-    GRAPH_SYNC = "graph_sync"
+    GRAPH_ANALYTICS = "graph_analytics"    # Usage analytics (aggregation reads)
+    GRAPH_BACKUP = "graph_backup"          # create-/restore-backup operations
+    GRAPH_MANAGEMENT = "graph_management"  # Lifecycle ops (subgraph/tier/style)
+    GRAPH_SYNC = "graph_sync"              # Connection /sync trigger only
     GRAPH_MCP = "graph_mcp"                # MCP operations
     GRAPH_OPERATOR = "graph_operator"      # AI Operator operations
     GRAPH_SEARCH = "graph_search"          # OpenSearch full-text search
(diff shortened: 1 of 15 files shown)


jfrench9:
@claude please review this PR

PR #839, opened 2026-07-10
refactor(content-ops): Phase A — move corpus + raw writes onto the operation envelope

jfrench9:
## Content-Ops Hard Cutover — Phase A (corpus + raw writes → the operation envelope)

First slice of the content-ops cutover (multimodal spec §4). Every **non-language write** leaves its bespoke REST verb for the operation envelope (`middleware/operations.py`), cloning the merged `remember`/`forget` template. Transport-independent command kernels so REST and (Phase C) MCP call the same functions.

**Hard cutover — breaking API changes, no backward-compat.** None of these surfaces are used in production. SDKs + frontends are regenerated after the cutover phases land. **One frozen exception:** `materialize source=extensions` (live RoboLedger OLTP→OLAP) is untouched and pinned by a regression test.

### New ops (`POST /v1/graphs/{graph_id}/operations/{op}` → `OperationEnvelope`, idempotency + audit)

| Op | Modality / flow | Sync/async |
|---|---|---|
| `index-document` / `delete-document` | corpus write | sync |
| `ingest-file` | raw → DuckDB staging (auto-chains materialize via `ingest_to_graph`) | async `wrap_pending`; small files stage inline → `completed` |
| `delete-file` | raw write (cascade DuckDB row removal + mark-stale) | sync |

`index-document` creates when `document_id` is absent, partial-updates when present.

### Removed REST write verbs (GETs all kept)
- `POST` / `PUT` / `DELETE /documents/{…}` → deleted
- `DELETE /files/{id}` → deleted; `PATCH /files/{id}` slimmed to disabled/archived status only
- `POST /files` (S3 presign) stays REST (control-plane, not a knowledge write) and now points callers at `POST /operations/ingest-file`

### Structure
- Staging + delete business logic extracted out of the routers into `operations/graph/commands/{ingest_file,delete_file}.py` (kernel-first; verbatim moves of the upload uploaded-branch and delete body).
- Generalized the write-access guard `_require_memory_write_access` → `_require_graph_write_access`, now shared by all content-op writes.

### Verification
- **Unit**: 8 content-op tests + materialize-extensions freeze test; broad regression **580 passed, 6 skipped**; `just test-code` clean.
- **Live e2e** (X-API-Key, real stack):
  - documents: index → list → search → delete round-trip ✓
  - files: presign → S3 PUT → `ingest-file` staged **3 rows to DuckDB** (fast path, `completed`) → `delete-file` cascade removed them from the `e2etest` table + marked graph stale ✓
[...]

PR #840, opened 2026-07-10
refactor(query): Phase B — StatementKernel + /query/cypher + /query/sql

jfrench9:
## Content-Ops Cutover — Phase B: StatementKernel + `/query/cypher` + `/query/sql`

Second phase of the content-ops cutover. The **query layer** becomes one surface with a language namespace, backed by a shared, transport-independent `StatementKernel`. Stacked on Phase A (#839, merged); diff here is Phase B only.

**Hard cutover — breaking API changes, no backward-compat.** SDKs + frontends regenerate after. The live `materialize source=extensions` path is untouched (not in this diff).

### B1 — extract the authorization gauntlet into `StatementKernel`
The Cypher `/query` handler's inline pre-execution gauntlet (write detection, three-tier write policy — main-graph reads-only / subgraph read+write+role / shared-repo reads-only — plus bulk/admin/DDL validation) moves into `middleware/graph/statement_kernel.py` so `/query/sql`, MCP, and Operators share **one** policy path instead of each re-implementing it and drifting.

- `StatementKernel.authorize(engine, graph_id, statement, user, session) -> StatementAuthorization`; raises the same 403/400s with identical detail strings.
- The handler delegates at the exact position the inline block held — inside the same `try:`, circuit-breaker (before) and dual-layer rate limit (after) left in place — so the streaming hot path's try/except is **byte-for-byte unchanged**. Behavior-preserving, zero API change.
- Scope is authorization only; the strategy/streaming/queue machinery stays in the handler ("wrap, don't relocate"). This is the rail Phase C (MCP) rides.

### B2 — namespace the query layer; promote SQL to first-class
- `POST /query` → **`POST /query/cypher`** (op `executeCypher`, LadybugDB)
- `POST /tables/query` → **`POST /query/sql`** (op `executeSql`, DuckDB) — handler moved into the query package, authorized through the kernel's SQL engine (read-only; shared repos blocked — they have no user columnar tables). Old `/tables/query` deleted; the **graph_api-internal** `tables/query` model pair + endpoint are untouched.
- Models renamed for full symmetry with `StatementKernel`: `CypherQueryRequest/Response → CypherStatementRequest/Response`, `TableQueryRequest/Response → SqlStatementRequest/Response`. Responses stay **independent siblings** — Cypher returns dict-rows (`data`), SQL returns positional rows (`rows`) — because graph and relational results are genuinely different shapes.
- SQL reframed off "staging": a relational lens on the same graph-centric data (relationship tables, no FKs), often *ahead* of the materialized graph (it's upstream in every path: uploads/OLTP → DuckDB → graph).
- Rate-limit categorizer: `/query/cypher` → `GRAPH_QUERY`; `/query/sql` keeps the DuckDB/columnar bucket (`TABLE_QUERY`).

### Verification
- **1201 tests** pass (query + MCP + models + rate-limits + tables/files); new kernel unit tests (each policy tier isolated) + `/query/sql` handler tests; `just test-code` clean.
- **Live e2e:** `/query/cypher` read → `200 {"data":[{"c":74}]}` (dict-rows); `/query/sql` `SELECT 1` → `200 {"columns":["x"],"rows":[[1]]}` (array-rows); old `/query` & `/tables/query` → **404**; SQL on shared repo `sec` → **403**; Cypher main-graph write → **403**.

### Deferred / follow-ups
- **B3:** SQL writes gated on a DuckDB write-connection sandbox variant (`get_writable_connection`) — seam noted in the kernel.
- **Phase C:** route MCP `read/write-graph-cypher` through `StatementKernel.authorize()`, retire inline-policy tool dispatch (`MCPRegistrar`).


jfrench9:

Focus areas:
- **Faithfulness of the B1 extraction** — `middleware/graph/statement_kernel.py` `_authorize_cypher` vs the inline gauntlet it replaced in `routers/graphs/query/execute.py`. Did any check, ordering, log message, exception status, or detail string drift? Critically: the extraction must **not** change the streaming hot path's try/except semantics — confirm the kernel call sits inside the same `try:`, with the circuit-breaker check (before) and dual-layer shared-repo rate limit (after) still in their original positions, so no exception that previously escaped/was-caught now behaves differently.
- **SQL engine parity** — the moved `/query/sql` handler (`routers/graphs/query/sql.py`, formerly `tables/query.py`) now routes shared-repo blocking through `StatementKernel._authorize_sql` instead of an inline check. Confirm SQL is still read-only, still blocked on shared repos (403), and that the circuit-breaker key (`table_query`) + graph_api `query_table` execution path are unchanged.
- **Rename completeness** — `CypherQueryRequest/Response → CypherStatementRequest/Response`, `TableQueryRequest/Response → SqlStatementRequest/Response`. Any dangling references, missed exports (`models/api/__init__.py`, `models/api/graphs/__init__.py`), or accidental edits to the **graph_api-internal** `TableQueryRequest/Response` pair (which must keep its names)?
- **Rate-limit categorizer** (`config/rate_limits.py`) — `/query/sql` must map to `TABLE_QUERY` and `/query/cypher` to `GRAPH_QUERY`; check the `path_parts[3] == 'sql'` branch can't misfire for other `/query/...` shapes.
- **Dead references** to the retired `POST /query` and `POST /tables/query` routes anywhere (tests, docs, scripts, OpenAPI-dependent code).

Context: hard cutover, no backward-compat expected; SDK/frontend regen after. Prefer correctness/regression findings over style.

PR #841, opened 2026-07-10
refactor: Phase C — MCP through the StatementKernel + content-ops surface cleanup

jfrench9:
## Content-Ops Cutover — Phase C (MCP through the kernel) + surface cleanup

Third phase of the content-ops cutover plus the incremental API-surface cleanup it surfaced. Seven commits, each self-contained.

**Hard cutover — breaking API changes, no backward-compat.** SDKs + frontends regenerate after. The MCP cypher path is live (SEC MCP server, Operators), so the reroute is behavior-preserving except where it deliberately converges MCP onto the REST policy.

### 1. `refactor(mcp)` — route cypher through the StatementKernel + shared analyzer
The MCP cypher surface re-implemented the statement policy in four inline copies that drifted from REST `/query/cypher`.
- **HTTP path** (`mcp/execute.py`): the inline read-cypher gauntlet → `statement_kernel.authorize(engine=CYPHER, …)` inside the short-lived session. MCP now applies the *identical* policy REST does — including the main-graph write block it lacked (a write via `read-graph-cypher` on a main graph is now a clean boundary **403**, not a tool-layer 500). Empty queries skip the kernel (validation, not policy).
- **Tool layer** (`cypher_tool`, `WriteCypherTool`): **stay** as defense-in-depth (the Operator path invokes tools directly, no FastAPI session — `tools/_gate.py`), but now use the same `cypher_analyzer` predicates the kernel composes instead of hand-rolled regexes.
- **Security fix**: `is_schema_ddl` missed bare `DROP TABLE` (LadybugDB/Kuzu syntax), `CREATE/DROP INDEX`, sequences — so a `DROP TABLE` on a subgraph was allowed through REST + MCP alike. Now caught everywhere the analyzer is used.

### 2–3. `refactor(api)` — Content Operations + Memory tags, memory reshape, content_ops split
- New **Content Operations** + **Memory** OpenAPI tags (Memory was used but never declared); the content-ops move off "Graph Operations" (now lifecycle-only).
- Memory mutations become content-ops: add **`update-memory`**; remove redundant `DELETE /memory/{id}` (forget is canonical) + `PUT /memory/{id}`. **recall** moves `POST /search/recall` → `POST /memory/recall`. `/memory` = recall + list + get.
- Memory prompted in MCP instructions (a MEMORY section) — it existed but was never surfaced.
- **Split** the 7 content-ops out of the 1519-line `operations.py` into `routers/graphs/content_ops.py` (lifecycle stays; shared `_ctx`/`_dispatch` imported one-way). Moved "Graph Operations" tag up next to "Graphs".

### 4–7. `refactor(files)` — a coherent, read-only-resource file surface
Every file **write** is now a content-op; `/files` is read-only.
- **Drop the speculative `disabled`/`archived` flags** (+ the `PATCH /files/{id}` whose only purpose was setting them, + `FileStatusUpdate`): added in a model-refactor, never wired to a real feature (no frontend, SDK stub only, `archived`'s hide-from-listings never implemented). A file is present or deleted.
- **Move `create-file-upload` to a content-op** (`POST /operations/create-file-upload`): creating an upload is a write (registers a GraphFile + presigns S3); no reason to be file-scoped while ingest/delete are content-ops. Logic extracted to `operations/graph/commands/create_file_upload.py`; `POST /files` deleted.
- **Order the file content-ops** create-file-upload → ingest-file → delete-file (lifecycle order).

Result: **`/files` = GET list + GET get**; every file write (create-upload, ingest, delete) is a content-op — matching documents and memory.

### Verification
- Full unit suite green (analyzer blast radius); MCP + security + kernel + REST query + router/file/command suites pass; `just test-code` clean throughout.
- Live e2e: MCP read 200 / main-graph write 403 / `DROP TABLE` 403 (both surfaces); `/operations/{create-file-upload,update-memory}` + `/memory/recall` work; retired routes (`POST /files`, `PATCH /files/{id}`, `PUT/DELETE /memory`, `/search/recall`, old `/query`, `/tables/query`) → 404/405; tags + op order correct.

### Deferred (later)
- Full `MCPRegistrar` (read-op specs → one-spec/three-surfaces); route the search/document MCP tools through a shared reader (Phase D).


jfrench9:

Focus areas:
- **MCP cypher reroute faithfulness** (`routers/graphs/mcp/execute.py`): the inline read-cypher gauntlet now calls `statement_kernel.authorize(engine=CYPHER, …)` inside the short-lived session. Confirm the two invocation paths stay correct: (a) the HTTP handler now gets the kernel's full policy including the main-graph write block it previously lacked, and (b) the **Operator path** (`operations/operators/*` → `GraphMCPTools.call_tool`, no FastAPI session) still enforces read-only/write policy via the tool-layer guards. Does the empty-query guard (`if cypher_query.strip()`) correctly let a missing query fall through to the tool's "query required" error instead of the kernel fail-safing it to a write?
- **Tool-layer predicate swap** (`cypher_tool._validate_read_only`, `memory_tools._validate_write_query`): confirm replacing the hand-rolled write-keyword / BLOCKED_WRITE_PATTERNS regexes with `is_write_operation`/`is_schema_ddl`/`is_bulk`/`is_admin` is behavior-equivalent — no dangerous statement that the old regex blocked now slips through, and no legitimate read/write now wrongly rejected.
- **The `is_schema_ddl` change** (`security/cypher_analyzer.py`): the new `\b(CREATE|DROP)\s+TABLE\b` / INDEX / SEQUENCE patterns must not false-positive on the existing `CREATE NODE/REL TABLE` cases (already handled above) or on property/label names, and must catch bare `DROP TABLE`. This predicate feeds the kernel + REST too — check the blast radius.
- **content_ops.py extraction**: pure move? Diff the 7 handlers + 5 guards against their prior form in operations.py — any drift in guards, ordering, or the shared `_ctx`/`_dispatch` import. Confirm no circular import (content_ops → operations one-way) and both routers mount at `/operations` without path collisions.
- **Memory reshape**: `update-memory` op parity with the removed `PUT /memory/{id}`; recall moved to `/memory/recall` with the shared-repo block added; no dangling refs to the removed `DELETE /memory/{id}` / `PUT /memory/{id}` / `/search/recall`.

Context: hard cutover, no backward-compat expected; SDK/frontend regen after. Prefer correctness/regression findings over style.

jfrench9:
@claude can you review the latest changes

jfrench9:
@claude the branch gained several commits since your last review (the file-surface reshaping). Please review the new work:

- **`create-file-upload` extraction** (`operations/graph/commands/create_file_upload.py` ← former `routers/graphs/files/upload.py`): is it a faithful move? Same validation order (entity-graph block, table_name required, shared-repo block, format/extension/name checks), same S3 presign + GraphFile.create(commit=True), same error handling. Confirm no behavior drift and that the content-op handler (`content_ops.py::create_file_upload_op`) correctly wraps it in the envelope (idempotency + result = {upload_url, file_id, s3_key}).
- **disabled/archived removal**: confirm dropping DISABLED/ARCHIVED from FileUploadStatus + deleting `PATCH /files/{id}` + `FileStatusUpdate` left no dangling references, and that nothing in the ingestion/staging path relied on those statuses beyond the implicit `upload_status == "uploaded"` filter (i.e., no regression for pending/failed files).
- **`/files` surface**: it should now be GET-only (list + get); `POST /files` and `PATCH /files/{id}` are gone. Any orphaned imports, router wiring, or SDK-facing operation_ids left behind?
- **Test migration**: four test files were repointed from the router handler to the command (`create_file_upload_cmd`) by re-aliasing the import. Confirm the patched attributes (GraphTable/GraphFile/S3Client/get_universal_repository/env) all still resolve against the command module, so the tests exercise the real code path rather than passing vacuously.
- **content-op ordering**: pure reorder (create → ingest → delete), no behavior change — sanity-check the block move didn't drop/duplicate a handler.

Context: hard cutover, no backward-compat; SDK/frontend regen after. Prefer correctness/regression findings.

PR #842, opened 2026-07-10
Move Reporting Style from graph to entity

jfrench9:
## Summary

Moves **Reporting Style** from a platform-DB `Graph` column to the extensions-DB `Entity`. It was a NOT-NULL column on `graphs` — a cross-DB pointer into the extensions `structures` / `reporting_style_networks` it references, with no possible FK. Relocating it onto the entity co-locates the pointer with the definitions it points at, and puts it at the grain where `entity_type` (which the equity-form default derives from) and per-entity facts already live.

This is the prerequisite for **heterogeneous per-entity styles in a multi-entity hierarchy** — each entity resolves to the same canonical calc-DAG subtotals, which are the consolidation interface. Consolidation itself (NCI concepts, intercompany eliminations, ownership-% treatment) is still future work; this just makes the grain correct.

One-shot cutover (no dual-write phase).

## Changes

- **Storage**: add `entities.reporting_style_id`; drop `graphs.reporting_style_id` (+ index, `Graph.create` param).
- **Resolution**: `network_picker` gains `load_entity_reporting_style(session, entity_id)` / `load_primary_reporting_style(session)`; the render path threads the already-resolved `entity_id` through. Reporting Style is no longer cached on `GraphExtensionContext` / the GraphQL context — resolvers resolve it from their own extensions session, **deleting the per-render platform round-trip**.
- **Write op** (dedicated, re-homed): `POST /extensions/roboledger/{g}/operations/change-reporting-style` (optional `entity_id`, defaults to primary) — a pure tenant-schema write. The old graph-level op, its request model, and command are removed.
- **Creation / reads**: `_provision_entity` stamps the derived style (from `entity_type`, explicit override wins) onto the entity at creation; `LedgerEntityResponse` exposes `reporting_style_id` (GraphQL auto-derives).

## Migrations (two DBs)

- **extensions `0020`** — add `entities.reporting_style_id`, backfill by re-deriving from `entity_type` (partnership→PART, llc→LLC, else Default), then drop the server default to match the model.
- **platform `ff27f43e5f7f`** — drop `graphs.reporting_style_id` + index (autogenerated).

## Deploy / rollout — reviewer attention

Because the old op/column already shipped to `main`, this needs expand/contract sequencing across the two DBs:

1. Run **extensions `0020`** (add + backfill) **before** the new code serves traffic — additive, safe while old code still runs.
2. Deploy **new code**.
3. Run the **platform drop last** — only safe once no live code reads the column. Old code reads `graphs.reporting_style_id`, so dropping it mid-rollout breaks old code, and **once dropped, rollback to old code is impossible**.

**Pre-drop data check**: the backfill re-derives from `entity_type`, so any prod graph whose style was manually changed (via the old op) to diverge from its legal form would be silently reset. Reconcile `graphs.reporting_style_id` vs `default_style_for(entity_type)` and re-stamp divergent entities **before** the irreversible platform drop. (Likely zero pre-launch.)

**SDK follow-up**: Python + TS clients currently expose the old graph-level `opChangeReportingStyle`; regenerate from the new spec after merge.

## Behavior change to note

Auth on the change-style op moves from **org-owner-only** (old graph-level op) to the **standard roboledger gate** (graph access + roboledger provisioned), matching `update-entity` and the rest of the entity-write surface.

## Not affected (verified)

- Frontends: no live code calls the op (only a roboledger-app test fixture).
- `cloudformation/api.yaml`: API-Gateway proxy with zero per-route paths — unaffected, under the CFN inline limit.

## Testing

- New `test_reporting_style.py` for the entity-level command (13 cases); rewrote the two tests coupled to the old graph-level command; fixed `GraphExtensionContext` constructions in 3 tests.
- Both migrations applied + verified locally.
- `just test-code` clean (ruff / format / basedpyright 0 errors / cf-lint); all affected suites green (models, graph, roboledger, graphs, graphql, serialization, taxonomy, mcp).

## Deferred

`reporting_style_id` not added to the `Entity` graph-node (`schemas/base.py`) — touches the shared SEC base schema + materialization mapping and isn't needed for the render path (reads OLTP). Easy follow-up if it should be queryable in Cypher.


jfrench9:
@claude can you review

PR #850, opened 2026-07-10
fix(security): viewer read-only enforcement + cypher write-bypass (F1, F2, F11)

jfrench9:
## Summary

Phase 1 of the application-security remediation (`specs/application-security-remediation.md`): the **two High findings**, both authorization/validation gaps that let an *authorized-but-under-privileged* principal do more than their grant allows. Launch-relevant — RoboLedger's multi-user graph sharing invites an outside accountant/auditor onto a graph at a role, and the read-only `viewer` promise was not kept. No cross-tenant leak (isolation holds); this closes the intra-graph privilege gaps.

### F1 (High) — read-only `viewer` could write the whole OLTP command surface
Write-intent defaulted to **read**, so a viewer reached `delete/update-journal-entry`, `close/reopen-period`, `execute-event-block` (posts GL + can trigger QuickBooks write-back), the block/mapping/agent writes, and the content-op document/memory writes — on **both** MCP and REST.

- **MCP** (`routers/graphs/mcp/execute.py`): only 3 tool names were flagged as writes; everything else (incl. all registrar-generated OLTP command ops) fell through to `access_type="read"`, so `validate_mcp_access` ran only the membership check a viewer passes. Inverted to a **fail-closed read-only allowlist** (`READ_ONLY_MCP_TOOLS`) — anything not a known read is a write and requires the member/admin role. New tools default to write.
- **REST registrar** (`middleware/extensions.py`): every command op mounted with `get_current_user_with_graph` (membership only). Enforce the write role in the generated handler — all extension OLTP ops are commands.
- **Content-ops** (`routers/graphs/content_ops.py`): `_require_graph_write_access` *claimed* "subscription + write role" but only ran the billing/lifecycle gate (which takes no user). Added the real role check, threaded the user, corrected the docstring.
- **Single source of truth**: new `require_graph_write_role(user_id, graph_id)` (`middleware/auth/dependencies.py`) wrapping `GraphUser.user_has_write_access`, called by both REST command surfaces; MCP enforces the same via `validate_mcp_access(..., "write")`.
- Analytical **view** reads (`build-fact-grid`) are hand-written, not registrar-mounted → viewers keep read access (no regression).

### F2 (High) — cypher write-detection bypass via in-string comment marker
`cypher_analyzer._clean_query` stripped comments **before** strings, so a `//` (or `/*`) *inside a string literal* made the comment stripper eat a real write keyword after the closed string — `MATCH (n) WHERE n.x = '//' CREATE (m) RETURN m` classified as READ, skipping the read-only rule + the viewer write check on `/query/cypher`, MCP `read-graph-cypher`, and the Operator path. Replaced the strip-comments-then-strings staging with a **single left-to-right scanner** that decides string/comment/identifier context at each position; fail-closed default unchanged. All four Appendix-B PoC payloads now classify WRITE.

### F11 (Low, bundled) — MCP `update-document` bypassed the 500k content cap
Enforced the cap in `DocumentService.update_document` (the kernel) so create/REST/MCP all get it.

## Testing

- `just test-code` green throughout (ruff/format/basedpyright 0/0/0).
- New coverage: `require_graph_write_role` (viewer 403 / member ok); fail-closed classifier (dangerous writes not on the read allowlist, core reads are, cypher-read excluded); registrar handler denies a viewer; F2 PoC payloads + read-side sanity; `update_document` oversized-content cap.
- Blast-radius suites pass: cypher-analyzer 59, extensions 69, graphs/mcp 97, extensions-middleware 58, mcp 609, auth-deps 58, doc-service 14, content-ops 12, roboledger/roboinvestor ops + views.
- Registrar wiring suites get an autouse no-op of the write-role gate (their mock users have no DB-backed `GraphUser` row); a dedicated viewer-denied test asserts the gate fires.

- **F1 residual**: hand-written graph-lifecycle ops in `operations.py` (`materialize` / `set-write-policy` / `create-subgraph`) are covered on MCP by the classifier but still lack a REST-direct write-role check (lower severity than the ledger mutation closed here).
- **Phase 2 (Mediums)**: F3 SSO-token-as-bearer, F4 graph_api engine blocklist, F5 streaming timeout/row-cap, F6 `--forwarded-allow-ips` — separate PR.

### DIFF
--- a/robosystems/middleware/auth/dependencies.py
+++ b/robosystems/middleware/auth/dependencies.py
@@ -459,6 +459,34 @@ async def get_current_user_with_graph(
   )
 
 
+def require_graph_write_role(user_id: str, graph_id: str) -> None:
+  """Assert the user holds a write role (member/admin) on the graph.
+
+  ``viewer`` is read-only; bare graph *membership* (which
+  ``get_current_user_with_graph`` proves) is not sufficient to mutate a graph.
+  This is the single write-role authorization check the command surfaces share
+  — REST command ops (the extensions registrar, content-ops) call it before
+  dispatching, and the MCP surface enforces the same via
+  ``validate_mcp_access(..., "write")``. Opens a short-lived platform session
+  (``GraphUser`` lives in the platform DB, not the per-graph OLTP DB).
+
+  Raises:
+      HTTPException: 403 if the user's role on the graph is read-only.
+  """
+  from robosystems.database import SessionFactory
+  from robosystems.models.core import GraphUser
+
+  session = SessionFactory()
+  try:
+    if not GraphUser.user_has_write_access(user_id, graph_id, session):
+      raise HTTPException(
+        status_code=status.HTTP_403_FORBIDDEN,
+        detail=f"Write access denied to graph {graph_id}; your role is read-only.",
+      )
+  finally:
+    session.close()
+
+
 async def get_current_user_with_repository_access(
   request: Request,
   repository_id: str,
--- a/robosystems/middleware/extensions.py
+++ b/robosystems/middleware/extensions.py
@@ -37,6 +37,7 @@ from sqlalchemy.exc import ProgrammingError
 from sqlalchemy.orm import Session
 
 from robosystems.database import get_db_session
+from robosystems.middleware.auth.dependencies import require_graph_write_role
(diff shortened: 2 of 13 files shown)


jfrench9:
@claude please review this PR

PR #851, opened 2026-07-10
fix(security): appsec phase 2 — mediums (F3–F6) + Phase-1 carry-overs

jfrench9:
## Summary

Phase 2 of the application-security remediation (`application-security-remediation.md`): closes the four **Medium** findings from the 2026-07-10 review, plus the carry-overs deferred from the Phase-1 (High) PR #850. No cross-tenant issues — these are authorization/validation/DoS-hardening gaps on the query and command surfaces.

### F3 — SSO handoff token accepted as a full bearer (`middleware/auth/jwt.py`)
- `verify_jwt_claims` now rejects any token with `sso` truthy or a non-`access` `type`; `create_jwt_token` stamps `type:"access"`.
- Legacy session tokens with no `type` are **grandfathered** (no logout storm on deploy).
- `/sso-exchange` decodes SSO tokens independently, so the SSO flow is unaffected — **verified no frontend impact** (all three apps hand the SSO token to `/sso-exchange` in the request body, never as a bearer).

### F4 — graph_api engine-adjacent blocklist (`graph_api/core/ladybug/service.py`)
- `validate_cypher_query` now enforces the shared `cypher_analyzer` predicates (`is_bulk_operation`/`is_admin_operation`/`is_schema_ddl`) to block `LOAD`/`COPY`/`ATTACH`/`INSTALL`/`CREATE NODE TABLE` on the ad-hoc query endpoint (mirrors the MCP write-query guard).
- Ordinary graph writes (`CREATE`/`SET`/`MERGE`/`DELETE`) still pass so writer instances work; legitimate DDL/bulk-load runs through the schema-install / table-materialize paths, which never reach this validator.

### F5 — streaming query DoS (`graph_api/core/ladybug/service.py`)
- `execute_query_streaming` gets a thread-based timeout on the initial execute (yields a `QueryTimeout` chunk), mirroring the non-streaming path.
- Adds a `MAX_STREAMING_ROWS` backstop that stops the stream and flags the final chunk `truncated` (so clients can detect and paginate).

### F6 — rate-limit IP collapse behind the ALB (`bin/entrypoint.sh`)
- uvicorn now passes `--forwarded-allow-ips "${FORWARDED_ALLOW_IPS:-127.0.0.1}"` (behavior-preserving default).
- `cloudformation/api.yaml` sets `FORWARDED_ALLOW_IPS` on the API container to `!Ref VpcCidr` (VPC CIDR, default `10.0.0.0/16`) — scoped to in-VPC callers (the ALB), never `*`. So on deploy the per-IP / brute-force limiters recover the real client IP.

### Phase-1 carry-overs
- **F1 residual:** `require_graph_write_role` gate added to the membership-only REST lifecycle ops — `materialize`, `create-subgraph`, `set-write-policy`. (`change-tier` is already org-owner-gated; `delete-subgraph`/`delete-graph`/`create-backup`/`restore-backup` are already admin-gated.)
- Registrar handler checks write-role **before** `pre_validate` (fail-fast authz).
- Enforcement-point comment on `_RegistrarMCPTool.execute` (MCP write-role is enforced upstream in `execute.py`'s `validate_mcp_access`).

### Audit logging (both new denial paths, for the SOC2 trail)
- `require_graph_write_role` emits `log_authorization_denied` (→ `AuthorizationDenied` detective-control metric) on every under-privileged write attempt across all surfaces.
- `verify_jwt_claims` emits a `SUSPICIOUS_ACTIVITY` (high) event when an `sso`/non-`access` token is presented as a bearer.

## Testing
- New/extended unit tests: F3 token-type (6), F4 engine verbs (12), F5 streaming timeout + row-cap truncation (3), lifecycle write-role wiring (3), audit-event assertions (2); `_bypass_write_role` fixture on the write-policy happy-path tests.
- `just test-code` clean (ruff + format + basedpyright + cf-lint).
- ~800 auth/security/graph_api/extensions-suite tests green locally. Full `just test` needs the local stack — not run in-session.

## Follow-ups
- **Post-deploy (F6):** verify the rate limiter sees the real client IP (not the ALB address) once this ships via the deploy workflow — this is a task-def env change, so CloudFormation auto-cycles the API tasks (no manual ASG refresh).
- Phase 3 (Low/hardening — F7–F10, F12) remains.

### DIFF
--- a/bin/entrypoint.sh
+++ b/bin/entrypoint.sh
@@ -180,13 +180,22 @@ configure_dagster() {
 case $DOCKER_PROFILE in
   "api")
     echo "Starting API service..."
+    # F6: --proxy-headers alone trusts X-Forwarded-For from ANY peer only when
+    # --forwarded-allow-ips permits it; uvicorn defaults that allow-list to
+    # 127.0.0.1, so behind the ALB every external client collapses to the ALB's
+    # IP and the per-IP / brute-force limiters degrade to one shared bucket.
+    # Set FORWARDED_ALLOW_IPS to the ALB/VPC CIDR (e.g. "10.0.0.0/16") so the
+    # real client IP is recovered. NEVER set it to "*" — that lets any caller
+    # spoof X-Forwarded-For and fully bypass brute-force protection. Default
+    # stays 127.0.0.1 (behavior-preserving) until the deployment sets the CIDR.
     exec uv run uvicorn main:app \
       --host 0.0.0.0 \
       --port 8000 \
       --loop uvloop \
       --http httptools \
       --access-log \
-      --proxy-headers
+      --proxy-headers \
+      --forwarded-allow-ips "${FORWARDED_ALLOW_IPS:-127.0.0.1}"
     ;;
   "dagster")
     echo "Starting Dagster webserver..."
--- a/cloudformation/api.yaml
+++ b/cloudformation/api.yaml
@@ -686,6 +686,12 @@ Resources:
             # AWS & Container Configuration
             - Name: AWS_REGION
               Value: !Ref AWS::Region
+            # Trust X-Forwarded-For only from inside the VPC (the ALB), so the
+            # per-IP / brute-force limiters see the real client IP instead of
+            # collapsing to the ALB's address (appsec F6). Scoped to the VPC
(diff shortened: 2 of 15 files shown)


jfrench9:
@claude please review this PR

PR #853, opened 2026-07-11
fix(security): reject subgraph IDs at the extensions GraphQL surface

jfrench9:
## Summary

Closes the last open item from the appsec sprint (multimodal §11 P2): the extensions GraphQL surface (`/extensions/{graph_id}/graphql`) admitted subgraph IDs and passed auth by resolving them to the parent graph, then dead-ended in extensions schema resolution (`_sanitize_schema` rejects `kg…_name`). No data leak today — but the asymmetry (auth resolves subgraph→parent, schema resolution doesn't) is latent risk: a future "helpful" subgraph→parent strip in the schema path would silently serve parent OLTP data at a subgraph address. A subgraph is a modality container (knowledge/scratch), not an extensions domain target, so this surface now rejects subgraph IDs outright.

## Changes

- **`robosystems/graphql/context.py`** — `get_context` rejects any subgraph ID with 403 regardless of authentication, placed after credential resolution (bad creds still get 401) and before `check_graph_access` (the parent-resolving step this closes off). The guard uses `is_subgraph()` — not a naive non-`kg` reject — so the `library` sentinel and shared repositories are unaffected, while shared-repo subgraphs (e.g. `sec_historical`) are also refused. Authenticated attempts emit `log_authorization_denied`, landing on the SOC2 audit trail per the appsec Phase-2 pattern.
- **`robosystems/middleware/mcp/tools/graphql_tool.py`** — MCP `query-graphql` mirrors the guard (returns an `invalid_target` error for subgraph-bound workspaces). It builds its own context and bypasses `get_context`, so the HTTP guard alone would not cover it.
- **`tests/graphql/extensions/test_context.py`** — new `TestSubgraphGuard` class: authenticated rejection fires before `check_graph_access` and is audited; anonymous rejection (introspection does not bypass) is not audited; shared-repo subgraph rejected; `library` sentinel and plain shared-repo IDs keep working.
- **`tests/middleware/mcp/tools/test_graphql_tool.py`** — subgraph-bound workspace is rejected before parse/execution.

## Testing

- `uv run pytest tests/graphql/ tests/middleware/mcp/` — 700 passed.
- `just test-code` (ruff + format + basedpyright + cf-lint) — all checks passed.

## Notes

- Merging this closes the final P2 from the 2026-07 application-security review — CodeQL HIGHs go to 0 in code once merged.
- The `query-interface.md` spec (design vault) was refreshed alongside this change; its §5 "tighten GraphQL" item is what this PR implements.

### DIFF
--- a/robosystems/graphql/context.py
+++ b/robosystems/graphql/context.py
@@ -78,6 +78,7 @@ from robosystems.middleware.auth.dependencies import (
 )
 from robosystems.middleware.extensions import load_graph_metadata
 from robosystems.middleware.graph.types import GRAPH_OR_SUBGRAPH_ID_PATTERN
+from robosystems.middleware.graph.utils.subgraph import is_subgraph
 from robosystems.models.core import User
 
 
@@ -133,6 +134,10 @@ async def get_context(
        then loaded once and its `schema_extensions` + `graph_type` are
        placed on the context so resolvers can enforce the per-domain
        feature gate via `require_extension` without a second lookup.
+
+  - **Subgraph IDs are rejected (403) regardless of auth** — this is
+    the extensions domain-app surface, not a subgraph modality. See
+    the inline guard below.
   """
   has_credentials = bool(api_key) or bool(request.headers.get("Authorization"))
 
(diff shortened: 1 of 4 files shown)


jfrench9:
@claude please review this PR

PR #858, opened 2026-07-11
feat(graph): LadybugDB 0.18.1 + Arrow-streaming materialization

jfrench9:
## Summary

Upgrades the graph engine from `real-ladybug` 0.13.1 to `ladybug` 0.18.1 (the package was renamed on PyPI), and replaces the DuckDB **ATTACH** materialization path with an **in-memory Arrow streaming** handoff — the change that makes the upgrade possible. LadybugDB 0.14+ creates persistent shadow catalog entries when ATTACHing a DuckDB database, which collide with the installed graph schema and break every graph build; probing on 0.18.1 confirmed the behavior persists with no workaround (ATTACH options, staging-name prefixes, post-DETACH DROP all fail).

The new path: DuckDB emits Arrow record batches straight from its result vectors (`fetch_record_batch`), and LadybugDB scans each batch's buffers via `COPY <Table> FROM <batch>` — no intermediate file. A first cut used a parquet file as the intermediary; that was superseded because the disk round-trip is a regression at SEC scale (millions of nodes shouldn't serialize to disk and back).

**On "zero-copy" (to be precise):** end-to-end is *not* zero-copy. A persistent, traversable graph requires LadybugDB to build its own CSR adjacency storage, so the data is always materialized into `.lbug` — true of ATTACH, parquet, and Arrow alike. Zero-copy applies only to the DuckDB→Arrow step (result vectors → Arrow buffers, no serialization). What this PR eliminates vs. parquet is the *intermediate* serialization + disk round-trip; vs. ATTACH it swaps DuckDB-storage-format deserialization for an in-memory Arrow scan and dodges the shadow-catalog bug.

## Changes

**Dependency + imports**
- `pyproject.toml` / `uv.lock`: `real-ladybug==0.13.1` → `ladybug==0.18.1`
- `import real_ladybug as lbug` → `import ladybug as lbug` across 8 modules + 4 test files

**Materialization (`graph_api/routers/databases/tables/materialize.py`)**
- `_materialize_table_impl` and `_fork_from_parent_duckdb_impl` stream DuckDB→Arrow→LadybugDB. Temp tables, `INSTALL/LOAD duckdb`, ATTACH/DETACH, and (in the interim) the parquet file are all gone. Both connections stay open and the copy interleaves per batch, bounding peak memory.
- The DuckDB batch is scanned by LadybugDB via a Python replacement scan — the local variable name must match the identifier in the COPY string (documented API pattern; `# noqa: F841` because the linter can't see the engine's frame read).
- The export SELECT is reconciled to the target schema when known (column order, NULL fill, target-type casts); otherwise DECIMAL→DOUBLE. LadybugDB has no DECIMAL type and `postgres_scan` stages Postgres NUMERIC (ledger amounts) as DuckDB DECIMAL, so it's cast at the source. `_needs_reconciliation` removed — reconciliation is unconditional now.

**Image (`Dockerfile`)**
- Drops the duckdb lbug extension (3 files) and `libduckdb.so` from both stages — unused after the refactor; httpfs + vector remain.
- Extension install path moved `~/.lbug/extension/` → `~/.lbdb/extension/` (the 0.15+ location); `graph_api/app.py` stale-cache clearing covers both generations.

**Docs/comments** — corrected stale "ATTACH pipeline" references in `operations/extensions/materialize.py` + `dagster/jobs/extensions.py`, and scoped the zero-copy wording to the DuckDB→Arrow step.

**CI** — restored `id-token: write` on the Claude review workflow (a prior hardening pass dropped it as "unused", but `claude-code-action` needs it to mint a GitHub OIDC token).

## Storage/migration semantics (verified, drives the rollout)

- A **checkpointed** v40 (0.13.1) database opens on 0.18.1 and auto-upgrades in place to v42; an un-checkpointed WAL fails replay — checkpoint on the old version first.
- The upgrade is **one-way per file** (0.13.1 refuses v42). Rollback is restore-from-backup, never re-open.
- **Read-only** open of a v40 file on 0.18.1 works and leaves it at v40 — shared replicas can run the new engine against existing S3 files.
- `EXPORT DATABASE` (0.13.1) → `IMPORT DATABASE` (0.18.1) round-trips cleanly, so the existing `ladybug_migration_export_job` / `import_job` pair is the prod migration vehicle.

## Coverage

Every DuckDB→LadybugDB materialization routes through `client.materialize_table()` / `client.fork_from_parent()` (SEC, extensions/tenant, generic-graph, chunked, subgraph fork) — the two functions this PR converts. `grep "DBTYPE duckdb"` across non-test code is empty: no ATTACH-based copy remains. The file-based `COPY FROM <file>` paths (generic-graph ingest, backup restore, SEC classify) are a separate mechanism that never used ATTACH and are unaffected.

## Testing

- `just test-all` green (unit + dbt + ruff + format + basedpyright + cf-lint) prior to the last doc-only commits.
- End-to-end on a clean-wipe local stack on 0.18.1:
  - SEC load from scratch (download → process → stage → materialize): 36/36 tables, 19,151 nodes, no frame-scan failures.
  - RoboLedger demo (tenant `postgres_scan` NUMERIC path): filed FY 2025 report, SHACL conforms, Arelle-valid XBRL, fact amounts precise to the cent (133.33 / 266.66 / 399.99 / 533.32) — confirming NUMERIC→DECIMAL→Arrow→DOUBLE preserves value.
- Isolated 0.18.1 probes: constructor kwargs, DDL, node/rel COPY FROM Arrow (incl. `ignore_errors`), streaming record batches, the storage-version matrix, and the replacement-scan resolving through the connection-pool context manager.

## Notes

- Deploying requires the graph-engine migration for existing v40 databases (export → deploy → import); the night-window sequence, verification, and rollback are documented internally.
- Arrow-vs-parquet throughput at SEC scale is now moot (no disk), but the next SEC materialization pass will give real numbers.
- `create_arrow_table` was investigated as a way to avoid the replacement-scan variable-name coupling; it creates a separate node table rather than a COPY source, so the replacement scan stays.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -65,10 +65,10 @@ jobs:
       pull-requests: read
       issues: read
       actions: read # Required for Claude to read CI results on PRs
-      # id-token: write intentionally omitted — Claude authenticates with
-      # CLAUDE_CODE_OAUTH_TOKEN and assumes no cloud (OIDC) role in this workflow,
-      # so GitHub OIDC token-minting is not needed. Restore it only if a step
-      # here ever assumes an AWS role via configure-aws-credentials.
+      id-token: write # Required: claude-code-action mints a GitHub OIDC token to
+        # authenticate to GitHub (setupGitHubToken → getOidcToken), independent of
+        # any cloud role. Without it the action fails: "Could not fetch an OIDC
+        # token." (This is GitHub auth, not AWS — do not drop it as "unused".)
     steps:
       - name: Checkout repository
         uses: actions/checkout@v7
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,8 +31,10 @@ ARG TARGETARCH=arm64
 ARG LADYBUG_EXT_VERSION=0.18.1
 
 # Create extension directories using internal version (where LadybugDB looks)
+# The duckdb extension is deliberately absent: materialization moved from
+# DuckDB ATTACH to a parquet handoff (LadybugDB 0.14+ creates persistent
+# shadow catalog entries on ATTACH that collide with installed schema).
 RUN mkdir -p /ladybug-extension/${LADYBUG_EXT_VERSION}/linux_${TARGETARCH}/httpfs \
-             /ladybug-extension/${LADYBUG_EXT_VERSION}/linux_${TARGETARCH}/duckdb \
              /ladybug-extension/${LADYBUG_EXT_VERSION}/linux_${TARGETARCH}/vector
 
 # Copy httpfs extension from extension repository (source: repo version, dest: internal version)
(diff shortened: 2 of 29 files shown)


jfrench9:
@claude can you review this

jfrench9:
@claude can you review this

PR #866, opened 2026-07-13
fix(materialize): restore complete relationship ingestion on ladybug 0.18

jfrench9:
## Summary

Two materialize-path fixes required to rebuild the SEC graph on LadybugDB 0.18. Together they restore complete, correct materialization at scale. The primary fix (ignore_errors) is new; the Arrow-buffer fix already shipped as the 1.6.1 hotfix and is included here for the merge to `main`.

### 1. `ignore_errors=true` silently drops valid rows (primary)

LadybugDB 0.18's `COPY ... (ignore_errors=true)` discards **valid** rows in proportion to the Arrow batch size:

| batch size | rows dropped |
| ---------- | ------------ |
| 5M | 2.8% |
| 10M | 17% |
| 20M | 37% |

On the full SEC rebuild (~20M-row, wide-string hash-batches) this collapsed **every relationship table to ~11.5% of its rows** — nodes loaded 100%, edges lost ~88% (e.g. `FACT_HAS_ELEMENT` 8.8M of 76.4M; `STRUCTURE_HAS_FACT_SET` 529K of 4.47M). The hash-batch `WHERE` was correct (100% coverage, verified) and all endpoint nodes were present, so the loss was entirely inside the COPY.

`ignore_errors` is also unnecessary: DuckDB staging already dedupes nodes (on `identifier`) and relationships (on `from`/`to`), and the schema materializes **all nodes before any relationship**, so every FK resolves. Removing it restores 100% (verified locally at 20M- and 40M-row batches).

### 2. Arrow 2 GiB string-buffer cap (already shipped as 1.6.1 hotfix)

The `Fact` node's `uri` column overflows Arrow's 2 GiB per-column regular-string buffer on ~19M-row batches. `SET arrow_large_buffer_size=true` makes DuckDB emit 64-bit LargeString offsets; ladybug's `COPY FROM <arrow>` reads LargeString (verified against a real Fact part-file). Batch stays 25M (one COPY per 20M hash-batch).

## Changes

- **`graph_api/.../materialize.py`** — plain `COPY` at both sites (no `ignore_errors`); `arrow_large_buffer_size=true` on both DuckDB connections; fork path **fails fast** on a copy error instead of swallowing it.
- **Full `ignore_errors` removal** across the materialize chain: the two graph-api request models, three client methods, the chunked/direct/Dagster/extensions/entity_sync/subgraph/SEC callers, the materialize command, the `MaterializeOp` API model (regenerates the SDKs), and the MCP `MaterializeTool` schema.
- **All materialize error handlers now fail fast** — a failed table/batch means missing data and is surfaced loudly (the tenant extensions loop keeps its deliberate per-table continue).
- Updated the staging-dedup rationale comment (`duckdb/manager.py`) + README; updated affected tests (dropped the obsolete "continues on ignore_errors" cases).

**Out of scope:** the separate file-based ingest path (`operations/graph/engine/ingest.py`, `COPY FROM 'file.parquet'`) and the unused `ingest` client/models — a different mechanism, left untouched.

## Testing

- Root cause and fix proven locally against `ladybug==0.18.1`: `ignore_errors=true` drops 37% at 20M rows; a plain COPY loads **100% at both 20M and 40M** rows.
- `basedpyright` clean across the repo (pre-commit hook); `ruff check`/`format` clean.
- Updated 5 test files. The unit suite runs in CI — the local Docker stack was down this session, so `just test` was not run locally.

## Rollout

Requires the new image on the shared master, then a **re-materialize only** (the staged `sec.duckdb` persists — no reprocess/restage). Verify edges ≈ nodes (`FACT_HAS_ELEMENT` ≈ 76.4M, `STRUCTURE_HAS_FACT_SET` ≈ 4.47M, 1:1 with nodes), then rebuild `sec_historical` → cutover/publish. The current prod `sec` graph is broken (~11.5% edges) and **unpublished**; the old graph still serves on the replicas.

### DIFF
--- a/robosystems/adapters/sec/pipeline/entity_sync/load.py
+++ b/robosystems/adapters/sec/pipeline/entity_sync/load.py
@@ -200,7 +200,6 @@ async def _load_tables(
         mat_result = await client.materialize_table(
           graph_id=config.graph_id,
           table_name=table_name,
-          ignore_errors=True,
         )
         rows = mat_result.get("rows_ingested", 0) if mat_result else 0
         total_rows += rows
--- a/robosystems/adapters/sec/processors/ingestion/materialization.py
+++ b/robosystems/adapters/sec/processors/ingestion/materialization.py
@@ -358,7 +358,6 @@ class LadybugMaterializer:
             response = await graph_client.materialize_table(
               graph_id=self.graph_id,
               table_name=table_name,
-              ignore_errors=True,
               batch_num=batch_num,
               num_batches=num_batches,
               timeout=CHUNKED_MATERIALIZATION_TIMEOUT,
@@ -399,7 +398,6 @@ class LadybugMaterializer:
           response = await graph_client.materialize_table(
             graph_id=self.graph_id,
             table_name=table_name,
-            ignore_errors=True,
             timeout=timeout,
           )
 
--- a/robosystems/dagster/jobs/graph.py
+++ b/robosystems/dagster/jobs/graph.py
@@ -163,7 +163,6 @@ class MaterializeGraphConfig(Config):
   user_id: str
   force: bool = False
   rebuild: bool = False
-  ignore_errors: bool = True
   materialize_embeddings: bool = False
   operation_id: str | None = None  # For SSE result updates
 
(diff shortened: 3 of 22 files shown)


jfrench9:
@claude please review this PR

PR #870, opened 2026-07-13
feat(sec): incremental LadybugDB materialization mode

jfrench9:
## Summary

Adds an **incremental materialization mode** for the SEC LadybugDB pipeline. Today every run does a full LadybugDB rebuild (delete the graph, recreate, COPY every table). This adds a mode that COPYs only rows not already in the populated graph — via a per-table keyset anti-join — keeping the full rebuild as a periodic reconciliation path. It is entirely **default-off**: the new server flag defaults to `false`, and both the sensor branch and the reconciliation schedule ship `default_status=STOPPED`, so merging has zero effect on prod until deliberately enabled in the Dagster UI.

The SEC graph is additive-only except `Entity`, and every node id is a deterministic UUID5 of its source URI, so "already loaded" is decidable by a key match — which is what makes the anti-join sound.

## Why the anti-join is mandatory (verified empirically on ladybug 0.18.1)

- **Node**: COPYing a row whose PRIMARY KEY already exists **hard-fails** (raises, aborts the batch). `ignore_errors=true` is not an option — it silently drops ~37% of valid rows on 0.18 (per the standing comment in `materialize.py`).
- **Relationship**: all SEC rel tables are `MANY_MANY` with no PK, so re-COPYing an existing `(src,dst)` edge **silently duplicates** it.

So COPYing the delta requires filtering out both existing node PKs and existing edges — a correctness requirement, not an optimization. The existing COPY handler already appends into a populated table; it only needs to be fed rows that are genuinely new.

## How it works

The diff runs **server-side** in the Graph API (`_materialize_table_impl`), where DuckDB staging and the `.lbug` file are colocated:

1. Export the target graph's existing keys to a transient parquet snapshot via ladybug's server-side `COPY (MATCH … RETURN …) TO 'file.parquet'` — node keyset = `identifier`, rel keyset = `(src,dst)`. Streaming server-side, so a 200M-edge keyset never materializes in Python memory.
2. Anti-join the DuckDB export SELECT against that snapshot (`… FROM {table} AS t WHERE NOT EXISTS (SELECT 1 FROM read_parquet(snapshot) k WHERE k.identifier = t.identifier)`), reusing the existing DuckDB→Arrow→COPY path unchanged. Only the surviving delta rows are ingested.

The snapshot lives under `DUCKDB_STAGING_PATH/incr_keys/` (instance-local, never published to S3), is exported once per table and reused across hash batches (a DuckDB temp table can't survive the per-call connection-pool teardown), and is cleaned up after the last batch (with batch-0 self-heal on a mid-run crash). An empty graph yields an empty snapshot, so the anti-join passes everything and the first run degrades cleanly to a full load.

**Entity** (the only mutable SEC node table) rides the same anti-join — new entities are added; a *changed attribute* of an already-materialized entity is refreshed by the periodic full reconciliation, not sub-reconciliation. A node-level DELETE+re-COPY can't be used: `DETACH DELETE` drops the node's edges (verified), which would orphan `FACT_HAS_ENTITY`.

## Changes

- **`graph_api/models/tables.py`** — `incremental: bool = False` on `TableMaterializationRequest`.
- **`graph_api/routers/databases/tables/materialize.py`** — `_export_incremental_keyset` + `_incr_keyset_path` helpers; node/rel detection; keyset snapshot lifecycle; the WHERE builder refactored into a fragment list that composes the file-filter, hash-batch, and incremental anti-join predicates; the export SELECT aliased `FROM {table} AS t` so the subquery can qualify outer columns.
- **`graph_api/client/client.py`** — `incremental` param on the `materialize_table` shim (forwarded only when true).
- **`adapters/sec/processors/ingestion/materialization.py`** — `materialize_mode` threaded through `materialize_from_duckdb` → `_trigger_ingestion` → both `materialize_table` calls; incremental skips the rebuild branch.
- **`adapters/sec/pipeline/configs.py`** — `materialize_mode: Literal["full","incremental"]` on `SECMaterializeConfig` (takes precedence over `rebuild_graph`).
- **`adapters/sec/pipeline/materialize.py`** — assets pass `materialize_mode` through and report it in run metadata.
- **`adapters/sec/pipeline/sensors.py`** — `sec_stage_to_materialize_sensor` now emits `materialize_mode: "incremental"`; new `sec_materialize_reconcile_schedule` (weekly full rebuild, STOPPED). Registered in `__init__.py`.
- **`tests/…/test_incremental_materialize.py`** — real-engine (ladybug + duckdb) tests exercising the actual helpers: incremental == full-load parity, no duplicate edges, idempotency (a re-run with no new staging ingests 0 rows), empty-graph-degrades-to-full, and the snapshot-path contract.

## Testing

- `just test-code` (ruff + format + basedpyright + cf-lint): **passed** (0 errors).
- New `test_incremental_materialize.py` (3 tests, real engines): **passed**.
- Existing `test_tables_materialize.py` + `tests/adapters/sec/pipeline/test_materialize.py` (19 tests): **passed** — no regression from the WHERE refactor / `AS t` alias.
- Design de-risked with standalone scripts against ladybug 0.18.1 (node dup-PK hard-fail, rel silent-dup, `COPY (query) TO parquet` support incl. empty/overwrite, `DETACH DELETE` edge cascade).
- Full `just test-all` unit suite: **not run** in this session (needs local env).

## Notes / Follow-ups

- **Enable-gate, not merge-gate.** Before enabling in prod, run one incremental materialize in staging and time the keyset export for the largest rel tables (`FACT_HAS_ELEMENT`, `ASSOCIATION_HAS_*`) against the ~40-min HTTP budget. That measurement decides whether to enable as-is or build the optimization below first.
- **Cost profile.** This is O(N)-read / O(Δ)-write: the CSR-write of every row (the expensive part of a full rebuild) is avoided, but the rel keyset export is a full traversal and the DuckDB anti-join scans the full staging table. Cost still scales with graph size, not changeset size — it reduces full-rebuild *frequency* (nightly → weekly reconciliation), it doesn't eliminate it.
- **Future O(Δ) path (no rewrite).** Capture the staging-insert delta (a `run_id`/`ingested_at` tag or `INSERT … RETURNING`) and use it as a candidate set, pushing only candidate keys through the LadybugDB anti-join — makes the export O(Δ) too. The current anti-join stays the correctness backbone.
- **Reconciliation cadence** defaulted to weekly (Sun 06:00 ET); confirm before enabling.
- `sec_historical` (frozen corpus) stays full-only.

### DIFF
--- a/robosystems/adapters/sec/pipeline/__init__.py
+++ b/robosystems/adapters/sec/pipeline/__init__.py
@@ -114,6 +114,7 @@ from robosystems.adapters.sec.pipeline.sensors import (
   sec_incremental_pipeline_sensor,
   sec_index_retry_sensor,
   sec_master_sleep_on_failure_sensor,
+  sec_materialize_reconcile_schedule,
   sec_post_materialize_publish_sensor,
   sec_post_stage_index_sensor,
   sec_processing_sensor,
@@ -194,6 +195,7 @@ def get_dagster_components():
     ],
     "schedules": [
       sec_incremental_download_schedule,
+      sec_materialize_reconcile_schedule,
     ],
   }
 
@@ -246,6 +248,7 @@ __all__ = [
   "sec_master_sleep_on_failure_sensor",
   "sec_master_wake_job",
   "sec_materialize_job",
+  "sec_materialize_reconcile_schedule",
   "sec_narratives_index_job",
   "sec_narratives_indexed",
   "sec_post_materialize_publish_sensor",
--- a/robosystems/adapters/sec/pipeline/__init__.py
+++ b/robosystems/adapters/sec/pipeline/__init__.py
@@ -114,7 +114,6 @@ from robosystems.adapters.sec.pipeline.sensors import (
   sec_incremental_pipeline_sensor,
   sec_index_retry_sensor,
   sec_master_sleep_on_failure_sensor,
-  sec_materialize_reconcile_schedule,
   sec_post_materialize_publish_sensor,
   sec_post_stage_index_sensor,
   sec_processing_sensor,
@@ -195,7 +194,6 @@ def get_dagster_components():
     ],
     "schedules": [
       sec_incremental_download_schedule,
-      sec_materialize_reconcile_schedule,
     ],
   }
 
@@ -248,7 +246,6 @@ __all__ = [
   "sec_master_sleep_on_failure_sensor",
   "sec_master_wake_job",
   "sec_materialize_job",
-  "sec_materialize_reconcile_schedule",
   "sec_narratives_index_job",
   "sec_narratives_indexed",
   "sec_post_materialize_publish_sensor",
--- a/robosystems/adapters/sec/pipeline/configs.py
+++ b/robosystems/adapters/sec/pipeline/configs.py
@@ -5,6 +5,7 @@ for maintainability and reuse.
 """
 
 from datetime import UTC, datetime
+from typing import Literal
 
 from dagster import Config, StaticPartitionsDefinition
 from pydantic import Field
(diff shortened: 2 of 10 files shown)


jfrench9:
@claude can you review?

PR #923, opened 2026-07-23
feat(examples): provision showcase history through real monthly closes

jfrench9:
## Summary

Follow-up to #922 (stacked on its branch — retargets to `main` automatically when #922 merges). The showcase runner stops minting ~14 monthly draft Reports as series scaffolding and instead **closes each historical month for real** — every close stamps that month's canonical statement FactSets, so the monthly series the metrics, forecast, and Plan grid bind is produced by genuine closes.

- `initialize_fiscal_calendar`: `closed_through=None` (never-closed calendar, all periods open) + explicit `set_close_target(last_completed)`.
- `generate_monthly_reports` → `close_monthly_history`: sequential `close_period` over every window except the last — **the close-target month stays open** (the AI-close demo beat); fail-loud on any close error; runtime assertion that the window derivation agrees with the seeded earliest period (the sequence gate would otherwise block every close). Returns the same window list as before (downstream contract preserved).
- `run_demo` reorder: the close loop runs **before** `create_schedules` (the closeable gate counts pending schedule obligations globally; post-loop schedules scope to the advanced `closed_through` and queue only the close-target month's obligation).
- Metric computes take `period_windows[:-1]` — the open month has no stamped facts until the AI close lands; its column arrives then.
- The annual report is unchanged: the deliberate publication act.

`examples/roboledger_demo/main.py` (`just demo-roboledger`) mints no monthly reports and is untouched. This is also the shape of the Harbinger 7-year history run: init → set target → sequential closes, each stamping its month.

## Testing

- `examples/` only — repo gates green (ruff + format + typecheck via pre-commit).
- Live verification: both showcase episodes re-provisioned end-to-end against the #922 backend (results in the PR thread / #922 verification).

### DIFF
--- a/examples/_scenario/runner.py
+++ b/examples/_scenario/runner.py
@@ -1080,10 +1080,12 @@ def verify_metrics(
 ) -> None:
   """Compute + render the Key Financial Metrics standing time series.
 
-  With ``period_windows`` (from :func:`generate_monthly_reports`) the
-  library-seeded rs-metric catalog computes at every month-end — the
-  monthly series the FP&A cadence doctrine calls for. Detailed values
-  print for the final window only; every window prints its counts.
+  With ``period_windows`` (the CLOSED windows from
+  :func:`close_monthly_history` — the caller slices off the open
+  close-target month) the library-seeded rs-metric catalog computes at
+  every month-end — the monthly series the FP&A cadence doctrine calls
+  for, bound to the close-stamped canonical statement sets. Detailed
+  values print for the final window only; every window prints its counts.
   InterestCoverage skips with a reason for a debt-free scenario,
   exercising the soft-fail path on purpose.
   """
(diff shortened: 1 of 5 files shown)


jfrench9:
Two live-found fixes landed on this branch (both surfaced by the first real never-closed provisioning runs — core code, riding this stacked PR so the stack stays linear; they merge to main with it):

- **`fix(close)`: advance the calendar before flipping the FiscalPeriod.** Latent ordering bug on `main`: with `closed_through IS NULL`, the advance's sequence check reads the earliest NON-closed period — which the just-flipped status had already moved to the following month, rejecting every first close on a fresh calendar. Never fired before because every calendar initialized with a baseline.
- **`fix(envelope)`: drop covering windows from the statement series.** With the FY-end month open, the annual report's publication set became that period_end's only set — FY totals rendered as a giant June column, and in scenario mode it shadowed the June forecast (actual-beats-forecast at equal period_end). The series now drops any set whose window strictly contains another loaded set's window; annual-only tenants keep their coarse series.

Live verification on both showcase episodes: 14/14 historical closes, 14/14 months stamped (4 canonical sets each, all with VerificationResults), `/reports` = 1 filed annual only, actuals series 14 columns, scenario series 26 columns / 12 forecast, forecast articulation identities exact. Reopen→reclose cycle on 2026-05: reopen retracted all 4 sets (results swept), re-close restamped fresh with `statements_stamped: true` + `statement_rule_summary` on the REST response.

PR #945, opened 2026-07-28
Resize graph tier instance ladder, reprice, and fix offerings metadata

jfrench9:
Downsizes the dedicated graph tiers to match observed load, reprices to follow the new cost base, and fixes three bugs in the public offerings catalog found along the way.

## Why

A month of prod telemetry across the three `ladybug-standard` m7g.large instances:

| Metric | Config allows | Actual |
|---|---|---|
| Memory | 6,144 MB cap, 2,048 MB/db | **~10% of 8 GiB (~800 MB)**, flat for a month |
| CPU (2 vCPU) | — | avg 0.6–1.2%, peak 27–42% |
| Disk IO | 630 Mbps EBS baseline | peaks 15–35 MB/interval (~1–2%) |
| Disk used | 50 GB volume | ~0% |

The 07/25 CPU and disk spikes correspond to the FP&A restamp — the heaviest recent materialization run — and it did not move memory off 10%.

## Instance ladder

| Tier | Was | Now |
|---|---|---|
| Standard | m7g.large (2 vCPU / 8 GiB) | **m7g.medium (1 vCPU / 4 GiB)** |
| Large | r7g.large (2 vCPU / 16 GiB) | **m7g.large (2 vCPU / 8 GiB)** |
| XLarge | r7g.xlarge (4 vCPU / 32 GiB) | unchanged |

## Pricing

$149 / $299 / $699 → **$99 / $249 / $599**.

At the effective SP/RI rates the fleet actually pays (~50% off on-demand, confirmed against June amortized Cost Explorer data — m7g.large ran $0.0384/hr vs $0.0816 list), $99 on m7g.medium earns the same margin that $149 on m7g.large earns today. The entry cut is margin-neutral. Existing m7g convertible RIs (8 normalization units) now cover 4 Standard graphs instead of 2.

## Memory model

Memory is re-expressed as parent + subgraph rather than one flat average divided across every database. XLarge's parent was previously capped at 1,180 MB — below Standard's 2,048 MB — because 30 GB was divided across all 26 databases. It now gets a full 8 GB pool with subgraphs at 1 GB.

Buffer pools are lazily-filled caps, not reservations, so nominal totals may exceed `max_memory_mb`; the hard ceiling is the admission-control guard at 85% of instance RAM. Sizing keeps parent + DuckDB under that even when both are hot.

## Dead config removed

The `memory: {baseline_mb, max_boost_mb, auto_reset}` blocks were read only by `GraphTierConfig.get_memory_config()`, which had no production caller. The real boost path reads `instance.ladybug_memory_boost_mb`, set only on the shared tier — so customer tiers never boosted at all. Also drops the unused `max_nodes` / `max_relationships` keys that staging and development carried.

Staging and development Standard are now value-identical to production.

## Offerings fixes

The catalog listed Large and XLarge with `max_subgraphs: 0` and `api_rate_multiplier: 1.0`, and omitted their storage line.

- `get_available_tiers(include_disabled=False)` gates on `deployment.always_enabled` / `enabled_default`. Large and XLarge set `enabled_default: false` and rely on `enable_var` (`LBUG_LARGE_ENABLED_PROD`), which only the deploy workflow reads — it appears nowhere in application or CloudFormation runtime config. The tiers were filtered out, `tier_config` resolved to `None`, and specs silently fell back to zeros while the tier was still listed and priced. Now passes `include_disabled=True`; the `customer_tiers` list and billing plans are the real gate.
- `GraphSubscriptionTier` now declares `instance_storage_limit_gb`. The router had been computing it, but the field was absent from the response model, so Pydantic dropped it for every tier including Standard.
- The infrastructure string reads new `vcpus` / `instance_ram_gb` keys instead of deriving from `duckdb_max_threads` and `max_memory_mb`. Those are tuning knobs — threads intentionally oversubscribe the CPU, and `max_memory_mb` is the LadybugDB budget after OS overhead. The old derivation already misreported Standard as 6 GB on an 8 GB box, and would have printed "2 vCPU, 3 GB RAM" for a 1 vCPU / 4 GB m7g.medium.

## Testing

`just test-all`: 2,783 passed. One pre-existing ordering-dependent failure (`test_incremental_equals_full_load`) reproduces identically on `main` and passes both in isolation and in a full `tests/graph_api` run — unrelated to this branch.

Verified against production config:

```
ladybug-standard   subgraphs=3    rate=1.0   m7g.medium (1 vCPU, 4 GB RAM)   storage=20.0
ladybug-large      subgraphs=10   rate=1.5   m7g.large  (2 vCPU, 8 GB RAM)   storage=50.0
ladybug-xlarge     subgraphs=25   rate=2.5   r7g.xlarge (4 vCPU, 32 GB RAM)  storage=100.0
```

No downstream changes needed — the frontends read prices from `/v1/offering` at runtime.

## Deployment notes

- The instance type is a CloudFormation parameter, so this **auto-cycles the three live prod Standard instances**. No ASG refresh needed, but it is a fleet replacement on customer graphs.
- Recommend deploying staging first and pushing a real book through it. The 4 GiB sizing is backed by telemetry from a near-idle three-instance fleet; it survived the FP&A restamp without denting memory, but has not been exercised by a busy multi-graph fleet.
- `LBUG_STANDARD_MAX_INSTANCES_PROD` was raised 5 → 20 separately (GitHub variable, takes effect on next deploy). RI coverage ends at 4 graphs; beyond that, graphs land on the Compute SP or on-demand.

### DIFF
--- a/.github/configs/graph.yml
+++ b/.github/configs/graph.yml
@@ -8,10 +8,16 @@
 # - Supports "MotherDuck for Graphs" pattern: DuckDB staging → LadybugDB materialization
 #
 # TIER PHILOSOPHY:
-# - Standard: Dedicated m7g.large, single database + 3 subgraphs, cost-efficient entry
-# - Large: Dedicated r7g.large, subgraphs for organization, team collaboration
+# - Standard: Dedicated m7g.medium, single database + 3 subgraphs, cost-efficient entry
+# - Large: Dedicated m7g.large, subgraphs for organization, team collaboration
 # - XLarge: Dedicated r7g.xlarge, maximum subgraphs, advanced scale
 # - Shared: Public data repositories (SEC) - separate infrastructure
+#
+# MEMORY SIZING: LadybugDB buffer pools are lazily filled caps, not reservations,
+# so nominal (parent + subgraphs) may exceed max_memory_mb — subgraphs are rarely
+# all active at once. The hard ceiling is the admission-control guard, which
+# trips at 85% of *instance* RAM (graph_api/core/admission_control.py), so size
+# parent + duckdb_memory_limit to stay under that even when both are hot.
 
 production:
   writers:
@@ -19,7 +25,7 @@ production:
     # LADYBUGDB STANDARD TIER - Dedicated m7g.large
     # =========================================================================
     - name: ladybug-standard
-      description: Dedicated m7g.large LadybugDB tier for cost-efficient entry with subgraph support
+      description: Dedicated m7g.medium LadybugDB tier for cost-efficient entry with subgraph support
       backend: ladybug
       stack_suffix: LadybugStandard
       tier: ladybug-standard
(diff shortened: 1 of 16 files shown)


jfrench9:
@claude can you review the latest changes?

PR #948, opened 2026-07-28
Differentiate rate limits by tier and bucket per graph

jfrench9:
Rate limiting worked as burst protection but was **undifferentiated by tier** — in four independent ways, each of which hid the others. A follow-up commit then makes every surface that *describes* the limits agree with what the first commit enforces.

## What was broken

| | |
|---|---|
| `rate_limiting.py:496` | pinned every authenticated user to `"ladybug-standard"`, making the `ladybug-large` and `ladybug-xlarge` tables **unreachable code** |
| the tables themselves | held **identical values**, so fixing the lookup alone would have changed nothing |
| `api_rate_multiplier` | read in four places, applied in **none** — `/v1/graphs/{id}/limits` reported 90/min to Large and 150/min to XLarge, who were served 60 |
| bucket key | keyed by **user**, so a customer with ten graphs shared one budget — per-graph pricing delivering no per-graph throughput |

The comment justifying the hardcode said *"graph-specific subscriptions are handled at the graph level."* The only graph-tier-aware limiter is `download_limits.py`, which governs monthly backup downloads. `graph_scoped_rate_limit_dependency` turns out to be a pure alias for the same function.

## What unlocked it

**LadybugDB writers are not shared.** Every customer tier is `databases_per_instance: 1`, so query load lands on the tenant's own box. The config comment asserted the opposite — *"protect shared resources (OpenSearch t3.medium, LadybugDB on m7g/r7g)"* — and that false premise is what justified flat limits.

Categories now split by what they actually touch:

| Resource | Categories | Treatment |
|---|---|---|
| Tenant's **own** instance | `graph_query`, `graph_read`, `graph_write`, `graph_mcp`, `graph_operator`, `graph_analytics` | scale **1×/2×/4×** by vCPU, bucketed **per graph** |
| **Shared** OpenSearch / RDS / API | `graph_search`, `graph_import`, `extensions_*`, auth, billing | **flat**, bucketed **per user** |

`graph_import` stays flat deliberately: LadybugDB ingests sequentially regardless of tier, so extra cores buy nothing.

## Shared repositories are excluded from per-graph bucketing

`sec` is **one graph every tenant queries**. Keying its bucket by graph id would put all of them in a single budget and let one heavy user exhaust it for everyone — worse than the bug being fixed. Those stay user-keyed, which is also what isolates tenants from each other. Subgraphs count too: `sec_historical` is as shared as `sec`, so the check uses `is_shared_repository_or_subgraph`.

## Subgraphs draw from the parent's budget

A subgraph of a *user* graph is the mirror-image case: `kg…_dev` lives on `kg…`'s instance and is not separately priced, so it buckets as `kg…`. Without this, an XLarge tenant with 25 subgraphs would hold 26 independent budgets against one machine.

## Failing closed

Tier resolution is Valkey-cached (5 min TTL, one lazily-created client per process) to preserve the limiter's zero-database-access property on the hot path. Cache miss, unknown graph, or lookup error all fall back to the **tightest** tier. A resolved tier string with no `SUBSCRIPTION_RATE_LIMITS` entry — `ladybug-shared`, or a value predating a tier rename — also floors at the fallback tier rather than falling through to the anonymous `base` table: degradation lands a paying customer on the cheapest paid tier, never below it. Failing open would hand out throughput nobody bought.

The static segments under `/v1/graphs` (`tiers`, `capacity`, `extensions`) are excluded from graph-id extraction. Latent — those routes use the general limiter — but defense in depth against one of them adopting the subscription-aware dependency and collapsing every caller into a single shared bucket.

## `api_rate_multiplier`

Removed from `graph.yml`. The API field remains — it's required on two public response models — but is now **derived** from the limits actually enforced. The follow-up commit finishes the job the first started: `/v1/offering` and `/v1/graphs/tiers` were still reading the deleted YAML key and advertising **1.0 for every tier** — the inverse of the original drift. Both now call the derived function, and tests pin the derived value at both the catalog level (`get_available_tiers` for production and staging) and the endpoint level, so neither surface can flatten again. **No SDK regen needed.**

## Reporting surfaces now match enforcement

- **`/v1/graphs/{id}/limits`** reports the enforced tier for shared repositories — it was reporting the `base` table (20/min) while enforcement grants the user-keyed standard tier (60/min), the exact "reporting a limit we do not honour" this PR exists to end. The same path covers legacy tier strings. Also in that response: `credits` was permanently `null` (the block read `monthly_credit_limit`, a column that does not exist, behind a bare `except: pass`), and `subscription_tier` came from a `User` attribute that also does not exist, reporting `ladybug-standard` to everyone. It now reports the enforced tier.
- **`/v1/graphs/tiers`** hid Large and XLarge in production by honouring `deployment.enabled_default`, which answers "is the CloudFormation stack deployed by default" rather than "can a customer buy this" — the same bug #946's sibling fix (1c25fbba) closed in `/v1/offering` — and fabricated $100/$300/$700 fallback prices matching neither billing nor Stripe. It now mirrors `/offering`: the billing catalog decides what is listed, and a tier it cannot price is omitted rather than invented.

## Deliberately not changed

Standard's numbers are **not** halved after its `m7g.large → m7g.medium` resize. Observed production load is ~1% CPU, and cutting existing customers' limits to match a smaller box trades real service for protection admission control (85% of instance CPU/memory) already provides.

## Caveat worth stating plainly

The 1×/2×/4× ratios are anchored to vCPU, **not measured**. The only load data is 27–42% peak CPU on a near-idle three-graph fleet. Admission control remains the real overload backstop; these are a product lever.

## Testing

Full suite: **11,413 passed**, `just test-code` clean. The only failures are the two pre-existing `test_incremental_materialize` cases that reproduce on `main`.

Tests are weighted toward the failure modes: fail-closed resolution, that the fallback really is the tightest tier — including for tier strings with no limits table, on both the cache and database paths — the 1×/2×/4× and flat-shared invariants, static-segment extraction, and bucketing pinned directly: two users on `sec` get **distinct** buckets, one user on two owned graphs gets distinct buckets, and a subgraph shares its parent's bucket. The derived multiplier is pinned at catalog and endpoint level.

Two existing tests were updated rather than fixed: one asserted `limit == 60` for Large with the note *"1.5x multiplier applied separately"*, and one pinned the user-keyed bucket. Both encoded the behaviour being corrected.

### DIFF
--- a/.github/configs/graph.yml
+++ b/.github/configs/graph.yml
@@ -22,7 +22,7 @@
 production:
   writers:
     # =========================================================================
-    # LADYBUGDB STANDARD TIER - Dedicated m7g.large
+    # LADYBUGDB STANDARD TIER - Dedicated m7g.medium
     # =========================================================================
     - name: ladybug-standard
       description: Dedicated m7g.medium LadybugDB tier for cost-efficient entry with subgraph support
@@ -32,9 +32,6 @@ production:
 
       # Subgraph configuration
       max_subgraphs: 3 # Up to 3 subgraphs (1 parent + 3 children = 4 databases total)
-
-      # Resource limits
-      api_rate_multiplier: 1.0 # Base rate limits (1x)
 
       # Copy operation limits
       copy_operations:
@@ -90,7 +87,7 @@ production:
         always_enabled: true # LadybugDB standard tier is the foundation - always available
 
     # =========================================================================
-    # LADYBUGDB LARGE TIER - Dedicated r7g.large
+    # LADYBUGDB LARGE TIER - Dedicated m7g.large
     # =========================================================================
     - name: ladybug-large
       description: Dedicated m7g.large with subgraph support (10 subgraphs)
@@ -100,9 +97,6 @@ production:
 
       # Subgraph configuration
       max_subgraphs: 10 # Support up to 10 subgraphs (1 parent + 10 children = 11 databases total)
-
-      # Resource limits
-      api_rate_multiplier: 1.5 # 1.5x rate limits (2x RAM, 2x CPU vs standard)
 
       # Copy operation limits
       copy_operations:
@@ -169,9 +163,6 @@ production:
       max_subgraphs: 25 # Practical limit of 25 subgraphs (marketed as "unlimited")
       # Note: 1 parent + 25 subgraphs = 26 databases total on xlarge instance
 
-      # Resource limits
-      api_rate_multiplier: 2.5 # 2.5x rate limits (4x RAM, 2x CPU vs standard)
-
       # Copy operation limits
       copy_operations:
         max_file_size_gb: 10.0 # Maximum file size for enterprise workloads
@@ -237,9 +228,6 @@ production:
 
       # Repository configuration
       max_subgraphs: 10 # Platform-managed subgraphs only (e.g., sec_historical)
-
-      # Resource limits
-      api_rate_multiplier: null # Separate rate limiting for public access
 
       # Copy operation limits (for platform data loading)
       copy_operations:
@@ -335,7 +323,6 @@ staging:
       tier: ladybug-standard
 
       max_subgraphs: 3
-      api_rate_multiplier: 1.0
 
       # Copy operation limits
       copy_operations:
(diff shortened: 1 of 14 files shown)


jfrench9:
@claude can you review these changes

PR #1042, opened 2026-08-04
fix(billing): honour webhook retry semantics and the monthly refill calendar

jfrench9:
## Summary

Three billing correctness fixes from the second-review pass over the v1.6.27–v1.6.29 wave (detail: `local/RoboSystems/specs/_review-raw/second-review-raw.md`, findings R4-1, R4-5, R4-6):

- **Webhook processing now completes before the response is sent.** The handler previously acknowledged events with a 200 and processed them as background tasks, so an event referencing a subscription we could not resolve was acknowledged and never redelivered. Processing now runs inline: an unresolvable subscription returns 409, a handler failure returns 500, and Stripe's redelivery becomes the real retry mechanism the code always assumed it had.
- **Monthly credit allocation gates on calendar month, not a 30-day count.** February is 28 days, so a day-count gate refuses the 1-Mar run for anything allocated on 1-Feb and silently skips March fleet-wide. The gate now matches the allocation cron and the ledger idempotency key (one allocation per calendar month), which also closes the same-month re-run window.
- **Checkout metadata is stamped onto the created subscription** via `subscription_data` — Stripe does not copy session metadata to the subscription, so the exact-resolution branch added in #1037 never saw it on the checkout path.

## Testing

- `tests/routers/admin/test_webhooks.py` rewritten for the synchronous contract: dispatch per event type, 409/500 paths leave the event unmarked, idempotency short-circuit unchanged.
- New allocation tests pin the Feb→Mar boundary and the same-month refusal.
- New checkout test pins `subscription_data.metadata`.
- `just test-code` green; targeted suites 176 passed.

### DIFF
--- a/robosystems/admin/commands/billing.py
+++ b/robosystems/admin/commands/billing.py
@@ -463,6 +463,33 @@ def add_bonus_credits(client, graph_id, amount, description):
   click.echo(f"   Description: {description}")
 
 
+@credits.command("reset")
+@click.argument("graph_id")
+@click.option("--reason", default=None, help="Reason recorded on the ledger rows")
+@click.confirmation_option(
+  prompt="Forfeit the remaining balance and refill to the monthly allocation?"
+)
+@click.pass_obj
+def reset_credit_pool(client, graph_id, reason):
+  """Reset a graph's credit pool to its monthly allocation.
+
+  The remaining balance is forfeited (recorded as an EXPIRATION ledger
+  row) and the pool refills to the monthly allocation. The scheduled
+  monthly reset still runs normally when the month turns.
+  """
+  data = {"reason": reason} if reason else {}
+
+  pool = client._make_request(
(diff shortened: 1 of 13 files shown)


jfrench9:
Second commit `55b15b20` adds the resolved product decision on forfeitures: the monthly reset now writes an EXPIRATION transaction for the discarded remainder (unspent allocation and bonus credits alike), so `SUM(transactions) == current_balance` holds through every reset. Also adds an admin mid-month pool reset — `POST /admin/v1/credits/graphs/{graph_id}/reset` and `just admin <env> credits reset GRAPH_ID [--reason]` — which forfeits and refills with both movements in the ledger, and leaves `last_allocation_date` alone so the scheduled monthly reset still fires when the month turns. Model tests run the full ledger-footing invariant (create → consume → reset) against a real database.

PR #1044, opened 2026-08-04
fix(orgs,views): real credit fields in org graph list, overlap-safe summary

jfrench9:
## Summary

Correctness fixes from the second-review pass (detail: `local/RoboSystems/specs/_review-raw/second-review-raw.md`, findings R5-1, R1-1039-1, R5-9):

- **`GET /v1/orgs/{org_id}/graphs` could not survive contact with a real credits row.** It read `credits.available_credits` / `credits.total_consumed` — attributes `GraphCredits` has never had — so any caller whose visible set contained a graph with credits got an AttributeError swallowed into a 500. It now reads `get_usage_summary` (balance = the column the consume path gates on; used = this month's consumption). The test that kept this green by stubbing the invented attribute names now drives a real credits row through a real consume, end to end.
- **`summarize_by_element` no longer double-counts overlapping duration windows.** After #1039 both the quarter and the YTD window ending the same day survive into `facts` (correct), which made `summary.total` sum them (wrong — NVDA Q3 FY2025's 35,082M was counted inside its own 91,166M nine-month window). Only the narrowest window per (entity, period_end) enters the aggregates now; the `facts` list is unchanged.
- **`cryptography` declared as a direct dependency** (`>=50,<51`). Auth cache, credential encryption, and connection credentials import it at module level, but it reached the prod closure only via `intuit-oauth → pyjwt[crypto]` — dropping that adapter would have broken the auth path at import time, and no version bound existed for the majors-individually rule to bind against.

## Testing

- Rewritten org-graphs test creates a real `GraphCredits` row and consumes against it — it 500s if the field names drift again.
- New summary tests pin the narrowest-window rule (NVDA's real figures), per-entity separation, and instant passthrough.
- `uv.lock` diff is 2 lines (the direct-dependency record); `just test-code` clean, targeted suites 51 passed.


jfrench9:
Second commit `a13f506b` also fixes the `daily_trend` aggregation in the org usage endpoint: it counted every `graph_usage` row (storage snapshots + credit events included) as an API call while `summary` filtered to API event types, so one payload carried two different figures. Both aggregations now share the `_API_EVENT_TYPES` definition; the aggregate test pins per-day trend contents against the same seeded rows the summary assertions use.

PR #1054, opened 2026-08-05
feat(mcp): remote MCP transport — Streamable HTTP at /v1/graphs/{graph_id}/mcp

jfrench9:
## Summary

Adds a remote MCP transport: `POST /v1/graphs/{graph_id}/mcp` speaks Streamable HTTP (JSON-RPC 2.0) directly over the existing per-graph tool layer, so any MCP client that supports remote servers (Claude custom connectors, Cursor, `mcp-remote`) can connect to a graph by pasting its URL — no npx bridge process, no client-side config beyond the API key header. The dispatch is hand-rolled rather than mounted from the MCP SDK server because the tool surface is dynamic per graph and every call must run behind the existing FastAPI dependency chain (auth, per-graph access, rate limits, write classification, circuit breaker).

## Changes

**`robosystems/routers/graphs/mcp/remote.py` (new)**
- JSON-RPC 2.0 dispatch for `initialize` / `notifications/initialized` / `ping` / `tools/list` / `tools/call`; protocol-version negotiation across the 2024-11-05 / 2025-03-26 / 2025-06-18 revisions; batching rejected; notifications answered `202` per Streamable HTTP.
- `graph_id` stays a URL path segment and never becomes a tool argument. `initialize` returns per-graph `serverInfo` (`robosystems-{graph_id}`) and instructions rebuilt per session from the live tool surface (the npx bridge freezes them at process start).
- Stateless by design: no `Mcp-Session-Id`, no GET-side SSE channel. Excluded from the OpenAPI schema (`include_in_schema=False`) so the JSON-RPC envelope never reaches the generated SDK clients.
- **SSE-on-POST streaming**: the strategy selector picks the response shape per call — fast tools return `application/json`; long/streaming strategies return `text/event-stream` carrying `notifications/progress` (emitted only when the request carries `_meta.progressToken`, with monotonically increasing values) followed by the final JSON-RPC response. Comment pings every 15s keep the stream alive across LB idle timeouts; a client disconnect cancels the in-flight tool work via generator teardown.
- **Queue bridge**: under load, cypher reads route through the shared query queue. The REST surface answers those with `202` + a polling URL, which has no MCP equivalent — the transport instead holds the stream, relays queue state as progress, and resolves with the result under an explicit 300s end-to-end ceiling. Disconnect or timeout cancels the queued query.
- **Workspace semantics**: a remote connector is URL-anchored, so `switch-workspace` no longer pretends to switch — after the normal access gauntlet it returns the target workspace's connector URL (subgraph-as-URL), and `tools/list` carries a remote-appropriate description for it. Read-only tools are annotated `readOnlyHint` per the MCP spec.
- Tool-execution failures (subscription, rate limit, timeout) return as `result.isError` content so the model can read and relay them; protocol failures return JSON-RPC errors. Client capability profile is asserted internally instead of sniffing User-Agent/`X-MCP-Client` headers, which remote clients don't send.

**`robosystems/routers/graphs/mcp/execute.py`**
- The tool-call authorization gauntlet (fail-closed write classification, StatementKernel statement policy for cypher tools, per-graph role validation, shared-repo subscription lookup, dual-layer volume rate limiting) is extracted verbatim from `call_mcp_tool` into `authorize_mcp_tool_call()`, now shared by the REST endpoint and the remote transport so the two surfaces cannot drift. No behavior change on the REST path.
- New `execute_tool_to_json()` executes a call to a single JSON result for a given strategy: cacheable strategies go through the existing Valkey MCP cache, streaming strategies are aggregated server-side, the rest execute directly.

**`robosystems/routers/graphs/mcp/__init__.py`, `robosystems/routers/__init__.py`**
- The remote router is exported separately and mounted with the `/mcp` prefix at the top level (FastAPI rejects an empty path on a prefix-less include).

**`tests/routers/graphs/mcp/test_mcp_remote.py` (new, 34 tests)**
- Envelope/error-code contract, protocol-version negotiation, per-graph `serverInfo`/instructions, `tools/list` annotations + description rewrite, gauntlet denials surfacing as `isError` content, token-gated monotonic progress mapping, and the queue bridge (progress relay, failure surface, cancel-on-disconnect).

## Testing

- `just test` — full unit suite: **12,084 passed, 24 skipped** (~4m41s).
- `just test-code` — ruff check, format, basedpyright, cf-lint: all clean.
- Live-verified against the local Docker stack on a real tenant graph: full lifecycle (`initialize` → `initialized` → `ping` → `tools/list` → `tools/call`), SSE stream with progress notifications and final response, token-gating (no token → no notifications), JSON fallback when the client doesn't accept SSE, StatementKernel write barrier surfacing as `isError`, `switch-workspace` URL resolution, `/openapi.json` confirmed free of the new route, bad key → 403.

## Notes / Follow-ups

- The app-side "Connect" page (copy-paste connector config per graph off `GET /v1/graphs`) is frontend work in the app repo, not this PR.
- Auth-failure shape deliberately keeps the existing `403` (no `WWW-Authenticate`/PRM) for the header-auth increments; revisited when OAuth lands.
- At deploy time, confirm the ALB idle timeout against the 15s SSE keepalive interval (fine against the 60s default).
- The queue bridge only engages under real load (queue depth/running thresholds); it is unit-tested with mocks and not exercisable on an idle local stack.

### DIFF
--- a/robosystems/routers/__init__.py
+++ b/robosystems/routers/__init__.py
@@ -60,6 +60,7 @@ from .graphs import (
   subscriptions_router as graph_subscriptions_router,
 )
 from .graphs.content_ops import router as graph_content_ops_router
+from .graphs.mcp import remote_router as mcp_remote_router
 from .graphs.mcp import router as mcp_router
 from .graphs.operations import router as graph_operations_router
 from .graphs.operator import (
@@ -87,6 +88,9 @@ router.include_router(
   operator_router
 )  # No prefix - handled in the operator module itself
 router.include_router(mcp_router, prefix="/mcp")
+# Streamable-HTTP MCP transport at the bare /mcp path (POST, JSON-RPC 2.0);
+# schema-excluded, so it never appears in the generated SDK clients.
+router.include_router(mcp_remote_router, prefix="/mcp")
 router.include_router(backups_router, prefix="/backups")
 router.include_router(
   usage_router
--- a/robosystems/routers/graphs/mcp/__init__.py
+++ b/robosystems/routers/graphs/mcp/__init__.py
@@ -12,6 +12,7 @@ This module provides a comprehensive MCP tool execution system optimized for AI 
 from fastapi import APIRouter
 
 from .execute import router as execute_router
+from .remote import router as remote_router
 from .tools import router as tools_router
 
 # Create main MCP router
@@ -23,5 +24,9 @@ router = APIRouter(
 router.include_router(tools_router)
 router.include_router(execute_router)
 
-# Export main router
-__all__ = ["router"]
+# The Streamable-HTTP JSON-RPC transport lives at POST /v1/graphs/{graph_id}/mcp
+# (the bare /mcp path). FastAPI rejects an empty path on a prefix-less include,
+# so it is exported separately and mounted with the "/mcp" prefix alongside this
+# router in robosystems/routers/__init__.py. Schema-excluded so the JSON-RPC
+# envelope never reaches the generated SDKs.
+__all__ = ["remote_router", "router"]
--- a/robosystems/routers/graphs/mcp/execute.py
+++ b/robosystems/routers/graphs/mcp/execute.py
@@ -222,6 +222,207 @@ async def execute_tool_directly(
     )
 
 
+async def authorize_mcp_tool_call(
+  graph_id: str,
+  tool_call: MCPToolCall,
+  current_user: User,
+) -> str:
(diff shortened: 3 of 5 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Applied the review follow-ups: notification-before-params-validation ordering (with regression test), deduped `_get_mcp_operation_type`, and documented the non-SSE queue fallthrough as deliberate. Per-method metrics breakdown stays a post-deploy follow-up.

PR #1055, opened 2026-08-05
feat(telemetry): per-user query telemetry and alarms for shared repositories

jfrench9:
## Summary

Adds a detection layer for the shared-repository query surface: classified query and tool outcomes are recorded through the security audit log with per-user and per-API-key attribution, every shared-repo query emits structured start/end cost lines, and the shared-replica stack gains CloudWatch alarms (including a container-start metric filter). Scope is deliberately shared repositories only — user graphs run on dedicated instances and are excluded by design at every entry point.

## Changes

**Telemetry kernel**
- `middleware/graph/query_telemetry.py` (new): outcome classification (status-code map, engine-connection-loss shapes, stream-aggregation fallback envelope), `SHARED_QUERY_COST` start/end line pairs joined by an execution id, and a shared-repository gate on every function. All best-effort — telemetry never raises into the request path.

**Attribution**
- `middleware/auth/dependencies.py`: API-key auth success now stashes the key's stored identification prefix on `request.state.api_key_prefix` (mirrors the existing admin-surface pattern), so telemetry carries both `user_id` and key identity. JWT requests leave it unset.

**Event vocabulary + metrics**
- `security/audit_logger.py`: two new `SecurityEventType`s (`QUERY_ABUSE_SIGNAL`, `QUERY_ENGINE_DISRUPTION`) mapped to CloudWatch metrics in the existing `RoboSystems/Security/{env}` namespace, plus a `log_query_abuse_signal` helper. Disruption outcomes publish their own metric so they can alarm at a lower threshold.

**Query router**
- `routers/graphs/query/execute.py`: outcome recording at the `HTTPException` chokepoint and the directly-returned timeout response, engine-disruption detection in the generic error handler, and cost lines on all terminal paths (streaming dispatch and queue handoff are labeled as such).
- `routers/graphs/query/sql.py`: records statement-kernel rejections; handler gains a `Request` parameter for key attribution.

**MCP surfaces** (both route around the query router, so they get their own hooks — recorded at outcome classification, since the remote transport wraps tool failures as HTTP 200 + `isError`)
- `routers/graphs/mcp/execute.py`: recording at the REST endpoint's exception chokepoints (covers the shared `authorize_mcp_tool_call` gauntlet and tool timeouts), aggregation-envelope check on the stream-aggregated path, cost lines with a `dispatched_stream`/`completed`/`error` outcome from the execution `finally`.
- `routers/graphs/mcp/remote.py`: the SSE tool-call and queue-bridge generators record timeouts, failures, and disrupted aggregations with full duration coverage; generator closure from a client disconnect is labeled `client_disconnected`. The direct JSON path records at its exception handlers and checks the aggregated result.

**Infrastructure**
- `cloudformation/graph-ladybug-replicas.yaml`: metric filter on the replica container-start log marker plus three alarms (`container-start`, `query-abuse-signal`, `query-engine-disruption`) wired to the existing replicas SNS topic. Homed here rather than `api.yaml` (which sits at the CloudFormation inline-template size ceiling). Thresholds are initial values, expected to be tuned once real traffic data exists.

**Docs**
- `static/description.md`, `static/graph-api-description.md`: regenerated API surface descriptions (drift from the recently merged MCP transport work).

## Testing

- Full gate `just test-all` run this session: **12,111 passed, 24 skipped**, ruff + format + basedpyright clean, cf-lint 71/71 PASS.
- New coverage: `tests/middleware/graph/test_query_telemetry.py` (classification, scope gate, cost-line pairing), `tests/security/test_audit_metrics.py` (new event→metric mappings), `tests/middleware/auth/test_dependencies.py` (key-prefix stash on success, nothing on failure).
- Existing query/MCP router suites (150 tests) pass unchanged apart from `test_sql.py` gaining the new handler parameter.

## Notes

- Detection only — no new enforcement or limiting anywhere in this PR.
- Alarms take effect on the next shared-replicas stack deploy; the emit side rides the API deploy. Both should land in the same release as the remote MCP transport.
- Known Phase-0 residual: query-router SSE/NDJSON streaming cost lines stop at dispatch; the MCP aggregation paths have full end-line coverage.

### DIFF
--- a/cloudformation/graph-ladybug-replicas.yaml
+++ b/cloudformation/graph-ladybug-replicas.yaml
@@ -799,6 +799,73 @@ Resources:
         - !Ref ReplicaAlertTopic
       TreatMissingData: notBreaching
 
+  # Container-start marker from the replica entrypoint. Fires on every
+  # container start on an existing instance (unexpected — the serving process
+  # should start once per instance) and on fresh instance launches (expected
+  # during deploys and scale-out), so pages during an instance refresh are
+  # normal. The alarm exists for the unexpected case.
+  ReplicaContainerStartMetricFilter:
+    Type: AWS::Logs::MetricFilter
+    Properties:
+      LogGroupName: !Sub /robosystems/${Environment}/graph-api
+      FilterPattern: '"Starting LadybugDB Shared Replica API"'
+      MetricTransformations:
+        - MetricName: ReplicaContainerStart
+          MetricNamespace: !Sub RoboSystems/SharedReplicas/${Environment}
+          MetricValue: "1"
+          Unit: Count
+
+  ReplicaContainerStartAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Properties:
+      AlarmName: !Sub robosystems-shared-replicas-${Environment}-container-start
+      AlarmDescription: Replica serving container started - expected only during deploys and scale-out
+      MetricName: ReplicaContainerStart
+      Namespace: !Sub RoboSystems/SharedReplicas/${Environment}
+      Statistic: Sum
+      Period: 300
+      EvaluationPeriods: 1
+      Threshold: 0
+      ComparisonOperator: GreaterThanThreshold
+      AlarmActions:
+        - !Ref ReplicaAlertTopic
+      TreatMissingData: notBreaching
+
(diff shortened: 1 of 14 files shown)


jfrench9:
@claude please review this PR

PR #1064, opened 2026-08-05
feat(auth): graph-scoped API keys with URL carriage for MCP connectors

jfrench9:
## Summary

Adds an optional **graph scope** to API keys, and lets the MCP endpoint accept a scoped key via a `token` query parameter — the connection path for MCP connector clients (claude.ai / Claude Desktop custom connectors) that cannot send custom headers.

- **Model/migration**: nullable `graph_id` on `user_api_keys` (+ index). NULL = account-wide key, behavior unchanged. Scoped keys mint with an `rfsc` prefix for legibility; the authoritative check is always the row's scope.
- **Scope enforcement** in the shared validation layer, on every carriage path: a scoped key is honored only for its graph and that graph's subgraphs, and is rejected on endpoints with no graph context. Enforced on both the cache-hit and DB paths.
- **URL carriage** (`?token=`) is accepted only on the MCP transport route, and only for graph-scoped keys — account-wide keys are rejected there. Header and JWT auth take precedence and are unchanged. `token` is already in the request-log redaction list.
- **Creation endpoint**: `createUserAPIKey` gains an optional `graph_id` (validated against the caller's graph access at mint time); `APIKeyInfo` responses now carry `graph_id`. Additive OpenAPI change — rides a client minor, no breaking SDK impact.
- **Telemetry hygiene**: new span processor redacts sensitive query parameters from URL-bearing trace attributes (OTel's built-in redaction covers only cloud-signature params), matching the request-log redaction list.

## Testing

- 20+ new unit tests: scope matching (incl. subgraph and prefix-collision cases), cache-hit and DB-path enforcement, the no-cache-poisoning rule for the URL-carriage rejection, dependency precedence (header > URL token), mint-time access validation, `rfsc` minting, and span redaction with an in-memory exporter
- Full gate green: 12,149 passed + dbt + ruff + format + basedpyright + cf-lint
- Live end-to-end matrix on the local stack (8 cases): scoped key via URL on its graph 200; account key via URL 403; scoped key on another graph 403; scoped key via header 200; account key via header 200; no credentials 401; scoped key on an account-level endpoint 401; malformed token 403. Verified the token appears in no server log line.

## Migration

One platform migration (`506bb56c5347`): add nullable column + index. Applied and verified locally; reversible.


jfrench9:
@claude  can you review

PR #1068, opened 2026-08-06
fix(mcp): harden the remote transport

jfrench9:
## Summary

Hardening pass on the remote MCP transport and its supporting middleware, from an internal review of the connector work:

- **Rate limiting now recognizes the authenticated principal.** The auth dependencies publish a sanitized identity on `request.state`, and the rate limiter consumes it for credentials it cannot reparse from headers — connector clients now get their graph's proper tier bucketing instead of falling to the anonymous tier.
- **Tool execution failures propagate as MCP `isError` results** and count as circuit-breaker failures (timeouts and backend errors; caller-side query-constraint errors don't trip the breaker). Failure results are never cached.
- **Transport conformance:** Origin validation and `application/json` enforcement per MCP Streamable HTTP; the pre-Streamable `2024-11-05` revision removed from the negotiable set; post-initialize `MCP-Protocol-Version` headers validated (400 on unsupported values).
- **`switch-workspace`** targets are constrained to the connector's graph family, with clearer guidance on setting up a connector for the target workspace.

Companion change: robosystems-mcp-client#64 (bridge log hygiene).

## Testing

- Full suite: 12,181 passed, 24 skipped; ruff + format + basedpyright + cf-lint clean
- New regression coverage: rate-limit identity resolution, `isError`/breaker outcomes across direct, streamed, and queued paths, transport gate, protocol-version validation, switch-workspace family constraint

### DIFF
--- a/robosystems/middleware/auth/dependencies.py
+++ b/robosystems/middleware/auth/dependencies.py
@@ -168,15 +168,29 @@ def _get_user_for_verified_jwt(user_id: str, token_session_version: int) -> User
 API_KEY_HEADER = APIKeyHeader(name="X-API-Key", auto_error=False)
 
 
-def _stash_api_key_identity(request: Request, api_key: str) -> None:
+def _stash_api_key_identity(
+  request: Request,
+  api_key: str,
+  user: "User | None" = None,
+  auth_method: str = "api_key",
+) -> None:
   """Record which API key authenticated this request on request.state.
 
   The first 8 characters are the key's stored identification prefix
   (``UserAPIKey.prefix``), so downstream telemetry can attribute activity
   to a specific key without access to the raw header. JWT-authenticated
   requests leave the attribute unset.
+
+  When the resolved ``user`` is passed, the sanitized principal is also
+  published as ``request.state.auth_user_id`` / ``auth_method``, so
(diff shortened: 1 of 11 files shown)


jfrench9:
Follow-up commit 541a1c20 from the second review pass:

- Streaming-wrapper error chunks now surface as error events instead of aggregating into a successful empty result (regression-tested through the real `execute_query_streaming` path, including rows-then-error mid-stream).
- The hard SSE-timeout and queue-bridge-timeout branches now record circuit-breaker failures; the bridge-timeout accounting is a deliberate call, documented inline.
- Content-Type validation is an exact media-type essence match (case-insensitive), not a substring check.
- The negotiable protocol set is exactly `2025-06-18` — earlier revisions promised batching semantics this transport rejects. Absent version headers are still served.

Full suite: 12,191 passed + quality gate clean.

Note for reviewers running the suite locally twice within 30 minutes: `test_sync_connection_disabled_provider` can fail with a 409 from a leaked `qb_sync` distributed lock in local Valkey (pre-existing, unrelated to this PR — dispatch failures after lock acquisition don't release the lock). `just admin dev cache flush locks` clears it; CI is unaffected. Worth a separate small fix.

PR #1080, opened 2026-08-07
feat(storage): daily reclaim of stranded build artifacts and orphan estates

jfrench9:
## Summary

Adds the collection mechanism behind the storage labels #1079 introduced: a daily platform janitor that deletes `transient` blue-green build artifacts stranded by crashed materializations and `orphan` estates left behind by deleted subgraphs. The itemization made these visible on `/usage`; nothing until now could actually reclaim them (a stranded `-wip` waited for the next materialize of the same database, and an orphan is unreachable through `delete-subgraph` because the registry row it resolves through is exactly what's gone). Deliberately not a user-facing operation — no new API surface.

## Changes

**Graph API — lock-guarded artifact deletion (`graph_api/routers/databases/management.py`, `client/client.py`)**

- `DELETE /databases/{id}` now refuses a `-wip`/`-prev` name unless it can acquire the base's materialization lock — the same lock every blue-green build holds from before its first step to after its swap. In-flight build → 409; acquire-not-check closes the check-then-unlink race.
- The materialize flow deletes its own leftover WIP while already holding that lock (step 1 cleanup and both failure paths), so the endpoint honours the `X-Materialization-Lock-Token` passthrough — the same protocol `swap_database` already uses — and the three self-cleanup call sites in `operations/extensions/materialize.py` now pass their token. Without this, the guard would have 409'd every build's own cleanup. This is the subtlest part of the PR and worth a close look.
- Valkey-unreachable degrades open, mirroring the swap endpoint: in that state builds also run unlocked, so both sides accept the same residual, and the active database is never the delete's target either way.

**Reclaim sweep (`operations/graph/storage_reclaim.py`)**

- Reuses the registry pass the storage breakdown uses (`_label_orphans`), so it deletes exactly what `/usage` reports as `transient`/`orphan` — never anything attributed to a live database. A registry read failure leaves items unlabelled and orphans untouched: the sweep fails safe.
- An orphan's estate (database + vector index + staging file) folds into one delete; a transient delete passes `preserve_duckdb=True` so the base's staging and vectors are never collateral.
- Per-item failures (409 from the lock, 404 for a fragment with no `.lbug` anchor) are recorded as skipped and don't abort the sweep.

**Dagster job (`dagster/jobs/storage_reclaim.py`, `definitions.py`)**

- `daily_storage_reclaim_job` at 5:30 UTC (after backup cleanup), mirroring the `backup_cleanup` shape. Sweeps active parent user graphs; subgraphs ride the parent's `{id}_*` scan.
- `STORAGE_RECLAIM_ENABLED` SSM flag read per run is the runtime kill switch; schedule is STOPPED in dev.

**Known limitation** (documented in the module): an orphan whose `.lbug` is gone but whose vector index or staging file lingers is skipped — the Graph API delete anchors on the database file. Reported in `skipped`, not silently ignored.

## Breaking Changes

None. No API surface changes — the only OpenAPI delta is description text on `StorageItem` noting the daily reclaim job. No SDK action needed.

## Testing

`just test-all` run this session — full suite green (12,264 passed) plus dbt, ruff, format, basedpyright, cf-lint. New coverage: the 409-under-lock refusal, the lock-token passthrough (the build's own cleanup must not 409 against itself), lock release, non-transient deletes never touching the lock, orphan-vs-transient targeting with estate folding, `preserve_duckdb` per type, registry-failure fail-safe (orphans untouched), dry-run, per-item skip-and-continue, and the job/schedule registration.

### DIFF
--- a/robosystems/dagster/definitions.py
+++ b/robosystems/dagster/definitions.py
@@ -87,6 +87,10 @@ from robosystems.dagster.jobs.shared_repository import (
   shared_replicas_refresh_job,
   shared_repository_refresh_replicas_job,
 )
+from robosystems.dagster.jobs.storage_reclaim import (
+  daily_storage_reclaim_job,
+  daily_storage_reclaim_schedule,
+)
 from robosystems.dagster.resources import (
   DatabaseResource,
   GraphResource,
@@ -204,6 +208,7 @@ all_jobs = [
   invoice_subscription_renewal_job,
   # Platform: Infrastructure
   daily_backup_cleanup_job,
+  daily_storage_reclaim_job,
   hourly_auth_cleanup_job,
   weekly_health_check_job,
   instance_health_check_job,
@@ -245,6 +250,7 @@ all_schedules = [
   monthly_usage_report_schedule,
   # Platform: Infrastructure
   daily_backup_cleanup_schedule,
+  daily_storage_reclaim_schedule,
   hourly_auth_cleanup_schedule,
   weekly_health_check_schedule,
   instance_health_check_schedule,
--- a/robosystems/dagster/jobs/storage_reclaim.py
+++ b/robosystems/dagster/jobs/storage_reclaim.py
@@ -1,0 +1,133 @@
+"""Dagster storage reclaim job.
+
+Daily sweep deleting the instance storage the breakdown labels reclaimable:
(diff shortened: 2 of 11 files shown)


jfrench9:
@claude please review this PR

PR #1118, opened 2026-08-09
Cross-graph share controls: recipient delete, block, and sender revoke

jfrench9:
Adds the authorization controls that cross-graph report sharing was missing on both sides.

Report sharing copies a published report from one tenant's schema into another's, across the org boundary, authorized capability-style: possession of an unguessable `graph_id` is sufficient, because the only way to hold one is that the subscriber handed it over. That model is deliberate — an investor is not in the company's org, is not entitled to the ledger, and *is* entitled to quarterly reports — but it is only sound if the recipient can refuse.

## What's here

**Recipient delete.** A copy is written with the sender's user id in `created_by`, so the owner rule could never match anyone in the receiving graph. An admin of the receiving graph can now delete a report that arrived from elsewhere. Scoped to copies — a graph admin still can't delete a colleague's own report, and the filed/archived immutability rule is untouched.

**Recipient block.** A new per-tenant `blocked_source_graphs` table (migration `0028`, fanned out to every existing schema), consulted in the target's session before anything is written. Optionally purges reports already received from that source; reports the recipient authored are never eligible. Blocking is idempotent and preserves the original `blocked_at`. A blocked sender is told rather than silently dropped — they already had a relationship, so a bounce is more honest than a shadow ban and stops them retrying. The recipient's private note is never disclosed.

**Sender revoke.** `revoke-report-share` withdraws a delivered copy and stamps `ReportShare.revoked_at` — a column that had been declared since the feature shipped and never written or read. Scoped to one recipient, so withdrawal is always deliberate. A recipient who already deleted the copy is not an error. The linked entity is left in place so an investor's declared holding survives.

## Two gaps found while building

Neither was in the original design, and both made the difference between a control and a gesture:

1. **Nothing marked the graph stale on deletion.** Materialization is a full rebuild from OLTP, so a removed row leaves the recipient's graph only when the graph is marked stale — and `delete-report` did so nowhere. Delete, purge and revoke now each mark the right graph, conditionally where a no-op is possible.
2. **Sharing never marked the *target* stale either**, so delivery depended on the recipient happening to have unrelated ledger activity. One line in a handler already being edited, so it's folded in.

## Surface and compatibility

Three new operations and one new GraphQL field (`blockedSourceGraphs`); no removals, no changed shapes. Additive, so this is a client minor on both SDKs — branches are up in `robosystems-python-client` and `robosystems-typescript-client` and should merge after this deploys, not before.

## Tests

25 across the surface. 17 mock-based cover the authorization rules; 8 DB-backed run against two real tenant schemas and prove the copy lands in the recipient's schema and nowhere else, that a block keeps it out, that a purge spares native reports, and that blocks don't leak across schemas. Two pre-existing `_share_to_target` tests needed a patch for the new block check.

Full suite green: 12393 passed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -176,19 +176,22 @@ jobs:
         run: |
           uv lock --locked
 
-      - name: Create the extensions database
+      - name: Create and migrate the extensions database
         run: |
           # The platform DB (robosystems_test) comes from the service
           # container's POSTGRES_DB; the extensions DB is the second of the two
-          # databases this repo runs and has to be created by hand. DB-backed
-          # extensions tests provision their own tenant schemas from the model
-          # metadata, so an empty database is all they need — no migrations.
+          # databases this repo runs, and CI never made it — so every DB-backed
+          # extensions test skipped itself, silently, forever.
+          #
+          # Migrating is what makes them worth running: 0002 seeds the taxonomy
+          # library from the JSON-LD frameworks, which is exactly what the
+          # library-backed suites assert against. ~25s on a cold database.
           echo "🗄️  Creating the extensions database..."
           PGPASSWORD=postgres psql -h localhost -U postgres -d postgres \
             -c "CREATE DATABASE extensions" \
             || echo "extensions database already exists"
-          PGPASSWORD=postgres psql -h localhost -U postgres -d extensions \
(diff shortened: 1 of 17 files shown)


jfrench9:
@claude can you review

PR #1122, opened 2026-08-09
feat(examples): RoboInvestor end-to-end demo, and the three defects it found

jfrench9:
## Summary

Adds the first end-to-end demo for RoboInvestor, which is also the first that crosses a graph boundary — and fixes the three defects it found on the way. `ref/roboinvestor.md` §6 names the absence of exactly this demo as the reason a whole-domain staleness defect survived six months: nothing in the development loop ever produced a pure-investor graph, so nothing ever exercised the cross-graph seam.

The demo ends in 31 hard assertions and exits non-zero on any failure, so it doubles as a pre-release gate.

### The demo — `examples/roboinvestor_demo/`

Meridian Ventures Fund I holds five private instruments (Series A preferred, a bridge warrant, a post-money SAFE, LLC units, and a seed position disposed mid-run). Two are issued by Cadence Labs, which keeps its books on this platform, so the fund pre-associates them by `source_graph_id` before any link exists. Cadence then shares its filed annual report into the fund's graph, which creates the linked entity, resolves both securities, and makes the traversal executable:

```
Portfolio → Position → Security → Entity → Report → Fact
```

- The issuer is the existing `saas_startup` scenario, provisioned inline on first run and reused after. `saas_startup_demo.main` now takes an `argv` so the caller's flags don't leak in as scenario flags.
- The investor graph is provisioned with `roboledger` as well as `roboinvestor`, because `add-publish-list-members` rejects any target that doesn't declare it. An investor keeps no books, so that predicate really wants to be "target can receive reports" — provisioned rather than worked around, so the constraint stays visible.
- Two assertions exist specifically because their absence went unnoticed: that every RoboInvestor write marks its graph stale, and that the stale *reason* came from a RoboInvestor operation rather than a ledger one that would mask it on a dual-schema graph.
- Everything goes through the HTTP API via the SDK facades. `_reset.py` is the sole direct-DB path and deliberately has no product surface.

### Cross-graph share: carry the sender's concepts — **please review closely**

A shared report copies facts, and every fact cites a concept. Library concepts resolve on both sides (`copy_library_into_tenant` gives every tenant the same deterministic UUID5 ids); the sender's own reporting extension does not — those are `elem_*` ULIDs minted in the sender's schema alone. `facts.element_id` carries no foreign key, so the share wrote them without complaint.

It surfaced two steps later, as everything: LadybugDB rejects an edge whose endpoint has no primary key, blue/green answers any table error by abandoning the whole WIP database, so one unresolvable concept stopped the recipient's entire graph from materializing — their own portfolios, securities and positions included. Reproduced with 6 of 85 shared facts; the recipient's graph came back `node_count: 0`. A disclosure note or a custom metric is enough to trigger it, which makes it the ordinary case.

- `_ensure_shared_elements` copies the sender's `reporting_extension` taxonomies alongside the report. **This widens what crosses the graph boundary on a share** — previously facts and the parent entity's metadata, now the concepts those facts cite. It stays inside an authorised share (the sender chose the recipient via a publish list) and is scoped to reporting extensions only; a fact citing the sender's chart of accounts is logged and skipped rather than dragging a foreign CoA across. Worth a second opinion on the boundary call.
- Copies are stamped `source='linked'`, deliberately outside `COA_SOURCES`, so the recipient can read the sender's concepts without the sender's revenue accounts appearing in the recipient's own chart of accounts. Mirrors `Entity.source='linked'`.
- The unit copied is the taxonomy, not the individual element: elements carry a self-referencing `parent_id`, and the abstract head a note hangs off is typically cited by no fact at all.
- `FACT_HAS_ELEMENT` now inner-joins `elements`, so any residual dangling reference drops one edge instead of a whole graph. The write paths own not creating the dangle; the materializer should not answer one bad id with total loss.

**Migration `0029` widens `check_element_source` to admit `'linked'` and must be applied before the code that writes it** — `just migrate-up extensions`.

### Forecast: the stale-tail sweep deleted its own levers

`_invalidate_stale_tail` scoped by `scenario_id` alone, on the stated assumption that "every set carrying it is produced here". The lever FactSet is authored by `_write_lever_fact_set` at create time, carries the same `scenario_id`, and spans the full horizon — so any run asked for fewer months than the horizon swept away the very assertions the next run reads, and the block answered every later `compute-forecast` with "has no lever FactSet — the block is corrupt", permanently. Now excluded by the same pair `_load_lever_fact_set` identifies it with. Regression test added.

### Demos: typed SDK reads, and a reset that predates `fact_dimensions`

- Eleven call sites still read 1.0-client GraphQL results as dicts (`list_elements`, `list_taxonomies`, `list_structures`, `list_information_blocks`, `get_report_package`). Each raises `AttributeError` on first contact, so every demo that maps a chart of accounts died at its mapping step: `demo-roboledger`, `demo-coffee-roaster`, `demo-saas-startup`, and both Seattle Method demos. An AST sweep over `examples/` is now clean.
- `fact_dimensions.dimension_id` is `ON DELETE RESTRICT` and both scenario resets drop `dimensions` before `facts`, so a re-run over any graph whose forecast had produced scenario facts died on the FK. Latent until now, because the forecast step was unreachable behind the item above.

## Breaking Changes

None. No GraphQL schema, operations envelope, or REST shape changed, so no client SDK regen is required by this PR.

One additive behaviour worth noting for consumers: elements can now carry `source='linked'`, so recipients of a shared report will see the sender's extension concepts in `list_elements`. `source` is an open string field and chart-of-accounts reads filter on `COA_SOURCES`, which excludes `'linked'` — so CoA, account rollups and fact-grid CoA resolution are unaffected.

## Testing

- `just test-all` — **12,408 passed, 23 skipped, 0 failed** (291s), plus ruff, format, basedpyright and cf-lint all clean.
- `just demo-roboinvestor` — **31/31 checks, exit 0**, against a local stack. The traversal returns Cadence Labs' reported facts against the fund's holdings.
- `just demo-roboinvestor --skip-share` — 22/22, isolating the portfolio surface from the handshake.
- `just demo-roboinvestor --reload-issuer` — full issuer ledger rebuild, confirming the forecast and reset fixes on the path that previously died.
- `just migrate-up extensions` — 0029 applied cleanly against the local extensions database.

Both failure modes were reproduced before fixing and confirmed gone after: the recipient graph at `node_count: 0`, and the 422 from a forecast whose lever set had been swept.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -2,8 +2,9 @@
 
 Runnable demos that load real and synthetic data into RoboSystems and query it
 back. Each one is a working reference for a different slice of the platform:
-double-entry accounting on a graph, SEC XBRL filings, a custom domain schema,
-and cross-taxonomy financial reporting.
+double-entry accounting on a graph, a venture portfolio with a cross-graph
+report share, SEC XBRL filings, a custom domain schema, and cross-taxonomy
+financial reporting.
 
 ## Prerequisites
 
@@ -24,6 +25,7 @@ just demo
 
 # Or pick one
 just demo-roboledger
+just demo-roboinvestor
 just demo-custom-graph
 just demo-sec --ticker NVDA --year 2025
 ```
@@ -113,6 +115,41 @@ Three things differ off-local, all deliberate:
 - **The graph must already exist.** Provision it the way a customer would
   (checkout, or `POST /v1/graphs` with a payment method on file) and pass its
   id; the runner will not create one on a deployed environment.
+
+### RoboInvestor — a venture portfolio, and a report that crosses graphs
+
+`examples/roboinvestor_demo/` · [walkthrough](roboinvestor_demo/README.md)
+
+The investment side, and the only demo that spans two tenants. Meridian
+Ventures Fund I holds five private instruments — a Series A preferred, a
+bridge warrant, a post-money SAFE, LLC units, and a seed position that
+gets disposed mid-run. Two of them are issued by Cadence Labs, which keeps
+its books on this platform, so the fund declares the relationship with
+`source_graph_id` before any link exists.
+
+Cadence then shares its filed annual report into the fund's graph. That
+share creates a linked entity, resolves both pre-associated securities,
+and makes one query possible:
+
+```
+Portfolio → Position → Security → Entity → Report → Fact
+```
+
+A private holding joined to its issuer's reported financials — the thing
+a single-tenant product cannot offer.
+
+```bash
+just demo-roboinvestor                  # provisions both graphs, runs everything
+just demo-roboinvestor --skip-share     # portfolio surface only, no handshake
+just demo-roboinvestor --revoke         # also withdraw the share at the end
+just demo-roboinvestor --dry-run        # preview the portfolio, write nothing
+```
+
+The issuer is the `saas_startup` scenario, provisioned inline on the first
+run and reused afterwards. The run ends with a hard validation pass and
+exits non-zero if any invariant fails — including that every RoboInvestor
+write marked its graph stale, which is the check that materialization
+depends on.
 
 ### SEC — public company financial data
 
--- a/examples/_scenario/reset.py
+++ b/examples/_scenario/reset.py
@@ -66,6 +66,10 @@ def reset_demo_state(graph_id: str) -> None:
     session.execute(text("DELETE FROM event_dimensions"))
     session.execute(text("DELETE FROM events"))
     session.execute(text("DELETE FROM agents"))
+    # fact_dimensions.dimension_id is ON DELETE RESTRICT, and facts are not
+    # wiped until further down — so the dimensions delete below fails on any
+    # graph whose forecast ran (scenario facts carry a `scenario` Dimension).
+    session.execute(text("DELETE FROM fact_dimensions"))
     session.execute(text("DELETE FROM dimensions"))
 
     # 4. Tenant-origin associations. The library's immutability triggers
--- a/examples/_scenario/runner.py
+++ b/examples/_scenario/runner.py
@@ -670,12 +670,12 @@ def create_mappings(
   offset = 0
   while True:
     page = client.list_elements(graph_id, source="rs-gaap", limit=1000, offset=offset)
-    items = (page or {}).get("elements", [])
+    items = page.elements if page else []
     if not items:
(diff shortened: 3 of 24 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Thanks — all three landed in `6248dd13`, and re-reviewing against the database rather than the diff turned up three more of the same class.

**Your findings**

1. **Fail-open on transient read** — agreed, and fixed by going the other way: the read now propagates rather than being caught. `_share_to_target`'s existing handler converts it into a per-recipient `ShareResultItem(status="error")` with the target transaction rolled back, so one recipient's failure neither corrupts their data nor kills the batch — and the sender is told instead of silently receiving a partial delivery. You were right that the inner join kept it non-catastrophic; it was still the original defect behind a rarer trigger.
2. **Mixed-batch warning** — fixed in the code rather than the doc. The uncovered concepts are now named, not just counted when the set is empty.
3. **N+1 existence checks** — batched, matching the pattern ten lines above.

**Three more, found while verifying yours**

4. **`order_by(Element.depth)` was a no-op.** The comment claimed it gave parent-before-child. `depth` is `NOT NULL DEFAULT 0` and nothing populates it — all 222 elements in the reference tenant carry 0 — so inserts landed in heap order and the tests passed on luck. Replaced with a two-pass insert (parentless, then wire), which removes the ordering dependency instead of fixing the sort.
5. **`parent_id` was unvalidated** where `parent_taxonomy_id` was carefully nulled if unresolvable — inconsistent by accident. Pass two now sets only parents that resolve.
6. **A `qname` collision failed the entire share.** `idx_elements_qname` is UNIQUE tenant-wide, so a second sender using the same namespace prefix would bounce a share neither party controls. The colliding concept is skipped with a warning; its facts land without a concept edge, which the inner join absorbs.

Taxonomies got the same two-pass treatment for symmetry — `parent_taxonomy_id` is the same self-FK shape.

Verified against the database rather than the exit code: the in-batch sibling parent and the cross-taxonomy library parent both resolve, and both copied taxonomies wire to the recipient's own `rs-gaap v1`. Demo 31/31, full suite 12,408 passed.

Your note that `view`/`block` are non-Optional in the generated models is correct — I left the guards as cheap insurance against a schema regen, not because they're load-bearing.

PR #1125, opened 2026-08-10
fix: nine defects surfaced by an end-to-end demo suite run

jfrench9:
## Summary

A full local reset and an end-to-end run of every demo turned up nine defects. They share a shape: each is invisible on a cold, first-time run, so the test suite passed straight over all of them. Three only appear once a cache is warm or a container is loaded, four are documentation describing behaviour the code does not implement, and two are demos reporting success over failed steps — which is why the others survived earlier runs unnoticed.

All ten demos now pass end to end, and the two Seattle Method demos are genuinely re-runnable, which their docs have always claimed.

## Changes

**Warm-path defects (`robosystems/middleware/auth/`, `robosystems/routers/graphs/query/`, `robosystems/operations/graph/`)**

- **Repository access cached its own denial.** `validate_api_key_with_graph` caches the allow/deny answer per key and graph, and nothing invalidated it when access was *granted* — so a denial recorded before a user subscribed outlived the subscription for the cache TTL, and the subscriber was refused. `invalidate_user_graph_access` already existed for exactly this and had no callers anywhere; `grant_access` / `revoke_access` now call it.
- **`GET /v1/user` returned 500 on a warm cache.** `email_verified` was missing from all five cached-user payload builders and all three reconstructions. The column has `default=False`, but column defaults apply on insert and a cache-built `User` never flushes, so it stayed `None` and `UserResponse` rejected it. **Worth a close look:** defaulting the missing field would have been worse than the crash — it reports `False` for a verified user, silently downgrading an authorization input — so a payload lacking the field is treated as stale and falls back to the database, which re-caches it in full. Self-healing, no cache flush needed on deploy. The four duplicated payload dicts are now one serializer, since building them in four places is how the field went missing from all four.
- **Capacity rejection surfaced as 500 and tripped the circuit breaker.** The Graph API correctly returns 503 with a reason; the client maps that to `GraphTransientError`, which fell through to the generic handler. That turned a resource ceiling into a self-sustaining outage — the rejections opened the breaker and every later query failed even once pressure cleared. Now 503 + `Retry-After`, and 503 never counts as a breaker failure.

**Committed writes reported as failures (`robosystems/operations/event_block/`, `robosystems/operations/roboledger/commands/taxonomies.py`)**

- **`create_event_block` reported a committed write as a 500.** The envelope was built *after* `session.commit()`, which expires the instance, so reading `event.id` triggered a reload that could come back empty → `ObjectDeletedError` → 500 for a write already in the ledger. A client retrying a 5xx then duplicates it. All four sites now build the envelope while the instance is loaded, then commit.
- **Repeat writes returned an opaque 500 instead of 409.** `create-event-block` (`source`, `external_id`) and `create-mapping-association` (`uq_association_structure_elements_type`) let unique-constraint violations escape as unhandled `IntegrityError`. That defeats the point of `idx_events_source_external`, which exists so re-delivering an upstream record is safe — a connector could not tell "already ingested" from a server fault, and 5xx reads as *retry* to most HTTP stacks. Both now pre-check and raise a typed error → 409, ordered ahead of the broad `ValueError: 422` in their error maps.

**Served contract corrected (`robosystems/models/api/`, `robosystems/routers/graphs/operator/`, `bin/setup/`)**

- **SQL named parameters could not be supplied.** `SqlStatementRequest.sql` tells callers to use `$param_name`, but `parameters` was `list[Any] | None` — a list only binds positional placeholders, so the documented form always failed. DuckDB binds named parameters from a dict; `parameters` now accepts either. The Cypher path already took `dict[str, Any]`, so this also removes an inconsistency between the two query surfaces.
- **The operator endpoints advertised three operators that do not exist.** Both descriptions named `financial`, `research`, `rag` with a per-mode credit table. The registry ships `cypher` and `mapping`, and credits are charged on token usage, not a fixed price per mode. Rewritten against the live registry, now pointing at `GET /v1/graphs/{graph_id}/operator` rather than enumerating.
- `live-financial-statement` omitted `cash_flow_statement`, though `LIVE_STATEMENT_TYPES` and the MCP enum both accept it.
- **Bootstrap seeded two retired SSM flag names.** `EXTENSIONS_ENABLED` and `LEDGER_ENABLED` are read nowhere — the code reads `ROBOLEDGER_ENABLED`, `ROBOINVESTOR_ENABLED`, `EXTENSIONS_GRAPHQL_ENABLED`. `EXTENSIONS_ENABLED` is worse than dead: it is a derived property, so it can never be settable, and an operator setting it to `false` got no effect while extensions stayed on. A fresh fork got two inert parameters and none of the three live toggles.

**Demos reporting success over failure (`examples/`)**

- `custom-graph` exited 0 with all six count checks and all seven presets failed; `sec-demo` did the same across fifteen presets and the document search. Both now fail the run.
- The GL ingests appended API failures to the same `warnings` list as benign payload notes, so a lost write hid behind a note — World Online reported "3388 ok, 3 warnings" and exited 0 while one entry never posted. Failures are counted apart and abort before the reconciliation, since reconciling a ledger with holes reports a lost write as a modelling difference.
- **`JE-225` had never posted, on any run.** Both its source lines carry `Amount=0` and the API correctly rejects a line item with no non-zero debit or credit. It went unnoticed because a zero-amount entry contributes nothing, so the reconciliation still matched 18/18 — green because it wasn't looking. Now detected from the payload before posting and reported as a source data-quality note.
- `memory_subgraph` wrote semantic memories to the subgraph, which the API refuses by design (403). It printed four 403s, declared success, and pointed the reader at a recall endpoint that would never answer. Now runs against the parent graph, as the error message instructs.
- Five Seattle Method / World Online steps called `.get()` on ariadne-generated GraphQL models, which are Pydantic with attribute access. The first aborted the demo; the rest would have failed one step at a time.

**Local dev (`compose.yaml`)** — graph-api raised 2g → 6g. It runs `ladybug-standard`, whose config assumes a dedicated 4GB instance holding one database, while `LBUG_DATABASES_PER_INSTANCE` lets it hold every local graph. At 2g it fit exactly one database and then rejected every query — which is what exposed the 500-vs-503 mapping above.

## Breaking Changes

None.

Three changes are SDK-visible but additive, all in the **generated tier** — no facade, root export, or symbol on the integration template's emit path is touched, so this rides a client minor rather than a major:

- `SqlStatementRequest.parameters` widens `list[Any] | None` → `list[Any] | dict[str, Any] | None`. Purely additive; positional binding is unchanged and verified.
- `create-event-block` and `create-mapping-association` gain a `409` response. New status on an existing operation; no request or success-response shape changes.
- Description text corrected on the two operator endpoints and on `live-financial-statement`'s `statement_type`.

Worth an SDK regen to pick these up, but nothing here requires coordination.

## Testing

- `just test` — **12,430 passed, 23 skipped, 101 deselected** (final run, after all commits).
- `just test-code` — ruff, format, basedpyright, cf-lint all pass.
- **Demo suite, run end to end against a freshly reset local stack**: roboledger (42/42 checks, 10,379 rows, report filed, SHACL conforms + Arelle valid), custom-graph, sec (NVDA 2025 — 4 quarters loaded, all presets + document search), coffee-roaster (16,018 rows), saas-startup (16,542 rows), roboinvestor (31/31, including the cross-graph `Portfolio → Position → Security → Entity → Report → Fact` traversal), seattle-method (18/18 exact match, |Δ| = $0.00, balance sheet balances at $14,450.00), world-online (22/23 — the one mismatch is flagged upstream data quality — trial balance balances).
- **Re-run verified**, since idempotency is the documented contract: seattle-method and world-online both exit 0 on a second run against their existing graphs, skipping what already exists and reporting what they skipped.
- **Fixes verified by observed behaviour, not by reading the diff**: repository-access lockout reproduced as 403 → subscribe → 200; `email_verified` confirmed against the database row rather than trusting the response; duplicate event returns 409; named SQL binding confirmed to actually filter (`$ind` → `Aerospace` → `Aerospace Innovators 6`), not merely parse.
- The two new envelope-ordering tests were confirmed to **fail against the previous ordering** before being kept, so they detect the regression rather than merely passing.

### DIFF
--- a/bin/setup/README.md
+++ b/bin/setup/README.md
@@ -8,7 +8,7 @@ Scripts for bootstrapping and configuring RoboSystems deployments. These handle 
 | -------------------- | ------------------------------ | ------------------- | --------- |
 | `bootstrap.sh`       | Complete first-time setup      | AWS SSO, GitHub CLI | 5-10 min  |
 | `aws.sh`             | Secrets + SSM parameters       | AWS credentials     | 1-2 min   |
-| `gha.sh`             | Configure ~80 GitHub variables | GitHub CLI          | 2-3 min   |
+| `gha.sh`             | Configure GitHub variables     | GitHub CLI          | 2-3 min   |
 | `bedrock.sh`         | Local AI development setup     | AWS credentials     | 1 min     |
 | `localstack-init.sh` | Local AWS emulation            | Docker (automatic)  | N/A       |
 | `postgres-init.sh`   | PostgreSQL databases           | Docker (automatic)  | N/A       |
@@ -263,7 +263,7 @@ just setup-gha
 7. ECR repository name
 8. Optional: RoboLedger/RoboInvestor app URLs
 
-**Variables Set** (~80 total):
+**Variables Set** — run `just gha-list` for the live set.
 
 #### Core Configuration
 
--- a/bin/setup/aws.sh
+++ b/bin/setup/aws.sh
@@ -201,12 +201,21 @@ function create_ssm_feature_flags() {
         "DIRECT_GRAPH_MATERIALIZATION_ENABLED=true"
         "EMAIL_VERIFICATION_ENABLED=false"
         "FACT_GRID_ENABLED=false"
-        "EXTENSIONS_ENABLED=false"
+        # Extensions surfaces are gated per product domain. There is no
+        # settable EXTENSIONS_ENABLED: it is a *derived* property in
+        # config/env.py (ROBOLEDGER_ENABLED OR ROBOINVESTOR_ENABLED), so
+        # seeding that name — or the retired LEDGER_ENABLED / INVESTOR_ENABLED
+        # — produces a parameter nothing reads, and an operator who sets it to
+        # false gets no effect while extensions stay on. Seed the three names
+        # the code actually reads. Code defaults are true; these start false so
+        # a fresh environment opts in per domain.
+        "ROBOLEDGER_ENABLED=false"
+        "ROBOINVESTOR_ENABLED=false"
+        "EXTENSIONS_GRAPHQL_ENABLED=false"
         # Kill switch for graph_usage_monitor_sensor's 80%/100% storage emails.
         # Seeded at its code default purely so it is discoverable in
         # `just ssm-list <env> features` — a kill switch you cannot find is not one.
         "GRAPH_USAGE_ALERTS_ENABLED=true"
-        "LEDGER_ENABLED=false"
         "LOAD_SHEDDING_ENABLED=true"
         "MCP_AUTO_LIMIT_ENABLED=true"
         "MCP_SEMANTIC_MEMORY_ENABLED=false"
--- a/bin/setup/bootstrap.sh
+++ b/bin/setup/bootstrap.sh
@@ -878,7 +878,7 @@ setup_secrets_and_variables() {
     echo "  AWS Secrets Manager - Application secrets & feature flags"
     echo "                        (required for deployment, safe to re-run)"
     echo ""
-    echo "  GitHub Variables    - ~80 variables for custom domains, scaling,"
+    echo "  GitHub Variables    - ~185 variables for custom domains, scaling,"
     echo "                        instance sizes, etc. (optional, has defaults)"
     echo ""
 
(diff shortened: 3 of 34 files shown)


jfrench9:
@claude can you review

PR #1126, opened 2026-08-10
fix(reports): make a shared report render for its recipient

jfrench9:
## Summary

A cross-graph report share delivered its facts and then hid them. The recipient saw a report header — correct name, period, "Shared report from …" provenance — over an empty **"No statements available for this report yet."** while their schema held all 85 facts. This fixes the delivery so a shared report renders for the tenant it was sent to, and carries the sender's **holon** across with it so the recipient views the publication the sender actually made.

### The render defect — `operations/roboledger/commands/reports.py`

`_share_to_target` collapsed the sender's N FactSets into a single set with `structure_id=None`, on the stated premise that a source-graph structure id "is meaningless in the target". That premise is false for the structures a report's statements actually use: the taxonomy library is seeded into every tenant schema by `provision_tenant_schema` under deterministic UUID5 ids, so the sender's balance sheet *is* the recipient's — same id, same presentation arcs. (Measured on a live pair: 55 of the issuer's 66 structure ids exist verbatim in the recipient, including all four statement structures.)

Every report read then dropped those facts, because all three resolve a FactSet's Structure:
- `get_report_package` inner-joins `structures` on `fact_sets.structure_id`
- `get_statement` filters on `structures.block_type`
- `load_fact_set_by_id_for_structure` requires the pin to match

FactSets now travel as themselves — one target set per source set, carrying `structure_id` when the id resolves on the far side. Facts whose structure is tenant-local (a `struct_*` ULID, e.g. a custom disclosure) pool on a structure-less set: delivered, queryable, and materialized into the recipient's graph, but not renderable as a statement until a structure-aware import exists. Nothing is dropped.

`share_report` now selects the FactSet rows alongside the facts and passes both; `source_facts` carries `fact_set_id`.

### Holon delivery — same file, plus `reads/reports.py`

The share now copies the sender's holon (and flat JSON-LD) into the recipient's bundle prefix, re-keyed to the recipient's graph + report id, and stamps `bundle_url` / `generation_count`. The holon is the report as `#scene` / `#boundary` / `#projection` named graphs — self-contained, renderable outside the system, and omitting `#lineage` by construction, so the books never travel with the report. Copying the sender's *object* rather than re-deriving one from the recipient's copied rows makes the recipient's view byte-identical to what the sender published.

The copy is best-effort and logs on failure: the rows are the load-bearing half, since they are what materializes into the recipient's graph and carries the cross-graph traversal. An object-store fault degrades the renderers rather than failing the delivery.

**Reviewers may want to look closely at this one:** `get_report_download_url` no longer gates *every* flavor on `bundle_url`. That column names the flat JSON-LD object, which the holon path never reads — it serves its own cached object or rebuilds from rows. The holon branch now asks about publication status directly. Without this the copied holon was present in S3 and unreachable. A negative-control test covers the case the old gate was incidentally handling (an unpublished report must still have no holon).

### Receiving is not authoring — `graphql/resolvers/_common.py`, `resolvers/ledger.py`, `commands/reports.py`

`provision_tenant_schema` creates every tenant table and seeds the library regardless of a graph's `schema_extensions`, so an investor-only tenant can hold a report perfectly well. Two gates assumed otherwise:

- `_share_to_target` refused any target without `roboledger`. Narrowed to "has an extensions tenant at all" (`roboledger` **or** `roboinvestor`), so a share still cannot land in a graph that never opted into the OLTP surface.
- The GraphQL report **reads** (`reports`, `report`, `reportPackage`, `reportDownloadUrl`) gated on `roboledger`. They now accept either extension via a new `require_any_extension` / `open_extensions_session_for_any` pair. Every other field on the resolver — and every write path — stays `roboledger`-only.

The effect is that a fund does not have to provision a ledger it will never post to in order to read what its portfolio companies send it.

## Breaking Changes

None.

No GraphQL schema change — no fields, arguments, or types were added, removed, or retyped — so no client SDK regen is required, and nothing here touches the stable-tier facades or the integration template's emit path. Three behavior changes worth noting for release notes:

- Report reads and share targets now accept `roboinvestor`-provisioned graphs. This widens access to data already resident in the caller's own tenant schema; it grants no cross-tenant reach.
- `reportDownloadUrl(format: HOLON_JSONLD)` now succeeds for a published report with no flat bundle. The error code on failure is unchanged (`REPORT_BUNDLE_NOT_AVAILABLE`); its message differs for the holon flavor.
- The share-rejection message for an ineligible target graph changed. `ShareResultItem`'s shape is unchanged.

## Testing

- `uv run pytest tests/operations tests/graphql tests/routers` — **4663 passed, 5 skipped**.
- Post-merge with `main` (PR #1125): `tests/operations/roboledger tests/graphql` — **905 passed**.
- `just test-all` was started locally but not run to completion; CI is the gate for the full suite.

New coverage:
- `test_share_controls_db.py` — two DB-backed two-tenant cases: a library structure survives the boundary and lands renderable; a tenant-local structure falls back without dropping facts. Both were confirmed to fail against the previous behavior (not merely against the new signature — the old collapse was simulated to check the assertion actually bites).
- `test_reports.py` — the sender's artifacts are re-keyed under the recipient's prefix and `bundle_url` is stamped.
- `test_report_download.py` — a published report serves a holon without a flat bundle; an unpublished one still does not.

End-to-end against the live stack, via `just demo-roboinvestor` — **31/31 checks**, including `Portfolio → Position → Security → Entity → Report → Fact resolves`, confirming the LadybugDB traversal is intact. On the recipient's copy afterwards:

- `reportPackage` returns 4 items — Balance Sheet 30 facts, Income Statement 28, Cash Flow 14, Equity 5, in display order.
- Sender and recipient holon objects share ETag `177466cf…` — byte-identical.
- `reportDownloadUrl(HOLON_JSONLD)` resolves for the shared copy.

### DIFF
--- a/.github/release-notes/v1.8.0.md
+++ b/.github/release-notes/v1.8.0.md
@@ -3,7 +3,8 @@ backups that run on a schedule, carry the right contents, and can be opened — 
 that had been advertised without being finished, subgraphs and cross-graph report sharing, were
 either completed or given the controls they needed. The last of those was then run end to end for
 the first time, which is how three defects in it — and around it — were found and fixed before a
-customer met them.
+customer met them. Looking at what the recipient of a shared report actually sees found a fourth,
+which the run could not: the data arrived correctly and rendered as an empty page.
 
 Covers everything since v1.7.0; the 1.7.x patches carried generated changelogs.
 
@@ -90,6 +91,8 @@ integrator will see them nowhere else:
   constant zero. API-usage metering is worth building; publishing zeros is not a substitute for it.
 - Reconciliation responses use `isReconcilingItem`; storage keeps its mechanical column names.
 - New: the report share controls, `update-graph-metadata`, and the graph-scope field on API keys.
+- New: blocking a sender with `purge` returns `purged_report_ids` alongside the count, so a caller
+  can see exactly which reports were removed rather than only how many.
 
 ## Writes that never reached the graph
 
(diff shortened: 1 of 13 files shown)


jfrench9:
@claude can you review

Issue #898, opened 2026-07-21
generate_rollup_rules.py out of sync with calc JSON-LD shape (regen wipes all RollUp rules)


jfrench9:
## Summary

`robosystems/taxonomy/scripts/generate_rollup_rules.py` is out of sync with the
current `rs-gaap-calculations/v1` JSON-LD shape. Re-running the documented regen
command produces **0 rules** and would overwrite the committed 15-rule
`rs-gaap-rollup-rules/v1` artifact with an empty package — silently disabling
every L2 calc-arc RollUp verification rule at the next seed/reset.

## Details

`build_rollup_rules()` reads arc nodes by the keys `arcFrom` / `arcTo` /
`arcAssociationType` / `arcRoleUri` / `arcWeight` / `arcOrder`
(`generate_rollup_rules.py:116-124`). The live source
`frameworks/rs-gaap/packages/rs-gaap-calculations/v1/taxonomy.jsonld` moved to
`from` / `to` / `associationType` / `role` / `weight` / `order` (2026-06-06,
before v1.6.5 — the loader `taxonomy/loader.py` reads this current shape).
Regenerating in-memory against the real source returns 0 rules vs the 15 in the
committed artifact.

CI never catches this: `tests/taxonomy/test_rollup_rules_seed.py` exercises
`build_rollup_rules` only with hand-built fixtures still in the old `arcFrom`
shape, so it never touches the real file.

## Impact

Pre-existing drift; **not** a v1.6.6 runtime risk (runtime uses the committed +
seeded artifact, never regenerates). The hazard is operational: anyone running
`uv run python -m robosystems.taxonomy.scripts.generate_rollup_rules` wipes the
rules. v1.6.6 ships a docstring guardrail on this script (chore branch) so the
"never needs regeneration" line no longer masks the trap.

## Fix options

- Update the key reads to the current shape (`from`/`to`/`associationType`/
  `role`/`weight`/`order`) **and** add a regen test that runs against the real
  `_CALC_SOURCE` and diffs the committed `@graph` (so CI catches future drift), or
- Deprecate/retire the script and remove the "regenerate via …" instruction from
  the package description.

Found during the v1.6.6 pre-release review.

MHa55anY:
Hey @jfrench9 Happy to take this if nobody is already on it.

Plan:
[...]

PR #1135, opened 2026-08-10
fix(taxonomy): repair the rs-gaap package references the Default Style split orphaned

jfrench9:
Closes the repo-side half of #898. The issue's diagnosis was right that
`generate_rollup_rules` is out of sync with the calc JSON-LD; investigating it
turned up three things the issue doesn't record, and they change the fix.

## The generator can't be repaired into something worth keeping

It groups calculation arcs by role URI and emits one RollUp rule per Network.
`12d8efba` (2026-06-04 23:32) collapsed the per-subtotal role URIs into three
statement-level Networks — `BalanceSheet` (8 parents), `IncomeStatement` (9),
`CashFlowStatement` (4). So the current failure isn't the empty package the
issue describes: the guard at `build_rollup_rules` raises `ValueError` on every
one of those Networks. Correcting the six key reads alone doesn't produce
rules, it produces an exception.

Regrouping by parent is straightforward, but the rule `@id`s aren't
recoverable that way. `bs-assets`, `is-cor`, `cf-netchange` are slugs of the
retired per-subtotal role URIs and are hand abbreviations — nothing derives
them from `CostOfRevenue` or `CashAndCashEquivalentsPeriodIncreaseDecrease`. A
rebuilt generator would carry a hardcoded 21-entry map of them, which is the
same maintenance burden as the 21 rules, minus the clarity. Retired, with the
reasoning recorded in `taxonomy/scripts/__init__.py`.

## The break let a second change land on top of it

Nothing held the artifact against its source, and the artifact's last
regeneration (`a16bc5d9`) was six hours before the regroup. `bfa0ebc7`
(2026-06-06) then split the combined Cash, Intangibles-incl-Goodwill,
AP-and-Accrued and SG&A leaves and added current-portion-of-LTD,
restructuring/impairment and FX/debt-extinguishment lines — leaving six rules
enumerating concepts the calc DAG no longer arcs. Rebuilt from the live source,
preserving each rule's `@id`, target, message, severity and origin. Five are
`bfa0ebc7`'s splits; the sixth (`is-rev`) is a formatting normalization, since
the committed file predates the shared `build_rollup_expression` and omitted
the parens around a single-child RHS.

## The stale enumerations were not entirely inert

The engine derives RollUp children from live calc arcs, which is why this
never surfaced. But `_evaluate_rollup_arc_derived` returns `None` when a parent
has no calc children in the merged DAG, and `engine.py` then falls back to the
frozen enumeration — where a replaced concept binds nothing, scores 0, and
reports a false failure against a real subtotal.

## Test

The generator's unit tests used fixtures still in the pre-2026-06-04 `arcFrom`
shape, so they passed against a builder that could not run on the real file.
Replaced with a drift gate that reads the live calc source: rules must cover
exactly its calculation parents, and each rule's children and expression must
match its arcs.

Verified it bites — removing a rule fails with
`parents in calc source with no rule: ['rs-gaap:GrossProfit']`, and truncating
a child list fails naming the rule. It subsumes the previous hardcoded count of
21, which is dropped so the count can't rot when the calc DAG legitimately
grows.

`just test-code` clean; `tests/taxonomy` + `tests/operations/taxonomy_block`
(491) and `tests/operations/information_block` (510) pass.

## Not included

Propagation to provisioned tenants. `rules` is a per-tenant copied table and
the provisioning copy is `ON CONFLICT DO NOTHING`, so this reaches existing
prod tenants only through a package-pinned `resync_library_into_tenant`
migration — and no migration has ever pinned `rs-gaap-rollup-rules`, so the
package has never been re-propagated since provisioning. Worth folding into the
next resync migration rather than carrying one on its own, given the fallback
path above is the only exposure.


jfrench9:
## Second commit: the same root cause, a second casualty — found in live data

Investigating a report of "bad rules" on a production entity graph confirmed
this PR's first commit against real books, and turned up a second, larger
instance of the same defect class.

### The rollup fix, confirmed live

Three RollUp rules were failing on that tenant's balance sheet —
`$AssetsCurrent`, `$AssetsNoncurrent` and `$LiabilitiesCurrent`. In every case,
across both period-ends, the reported residual equalled **exactly** the balance
of the concept the stale enumeration fails to name:

| Rule | Concept the stale enumeration omits |
|---|---|
| `$AssetsCurrent` | `CashAndCashEquivalentsAtCarryingValue` |
| `$AssetsNoncurrent` | `IntangibleAssetsNetExcludingGoodwill` |
| `$LiabilitiesCurrent` | `AccountsPayableCurrent` |

The tenant's element set is post-split and does not contain the combined
concepts the rules name, so each binds nothing, scores 0, and the subtotal
comes up short by precisely the balance that should have counted. Six exact
matches across three rules and two period-ends — not a coincidence.

This also corrects an assumption in the first commit's message. I had reasoned
the fallback path was near-unreachable because the keep-critical policy retains
calc arcs. It is not — it has been firing on live books since at least
2026-07-07.

### The second casualty: cash-flow derivation arcs

`bfa0ebc7` orphaned more than the rollup rules. Four `derivation` arcs still
keyed on the replaced combined leaves, so the flows they derive could never
fire. On the same tenant a long-term debt issuance rendered with **financing
activities at zero**, its proceeds absorbed into
`IncreaseDecreaseInOtherOperatingCapitalNet` — operating cash flow overstated
by the full amount of the issuance, and the statement still tied.

That is the worst property of this defect class: it is conservation-preserving.
The money goes somewhere, the totals foot, and nothing errors.

### Guard

Both gates are verified to fail when they should, not merely to pass now. The
reachability gate derives its notion of "mappable" from the live calc DAG plus
the suggester's own `RS_GAAP_SYNTHESIZED_DETAIL_ALLOW`, so it can't drift from
what the renderer actually accepts.

### Still outstanding

Neither commit reaches provisioned tenants. `rules` and `associations` are
per-tenant copied tables and the provisioning copy is additive, so both fixes
need a package-pinned `resync_library_into_tenant` migration — and none has
ever pinned `rs-gaap-rollup-rules` or re-propagated `rs-gaap-calculations`.

jfrench9:
## Migration `0030` + a deep-dive audit for more of this defect class

### Deep dive: no further instances

Audited the framework for the same shape — a reference that survives a
concept split, resolves to nothing, and degrades silently rather than erroring.
Six checks, all clean:

| Check | Result |
|---|---|
| Cross-package referential integrity (all 16 rs-gaap packages) | 50 undeclared refs, **all pre-existing curation artifacts** — equivalence pointers and uncurated `*Member` dimension members. Not the split class. |
| Hardcoded `rs-gaap:` qnames in Python (69 distinct) | No defects. `_CASH_ANCHOR_QNAMES` contains pre-split entries but is a defensive **superset** and includes the live leaf, so anchoring works. |
| Rule operands across `rs-gaap-rules` / `rs-metric` / `rs-driver` / `rs-gaap-rollup-rules` | Clean. One dormant hit (`ci-identity` → unwired OCI concepts) which `EqualTo` **skips** on unbound operands — no misstatement. |
| `tenant-exclude` vs the calc DAG | **Zero** calc leaves or parents excluded. The only excluded derivation target is `AccruedIncomeTaxesCurrent` — corroborating the `KNOWN_DORMANT` entry, which genuinely can never be mapped. |
| `_CANONICAL_ROOTS` reachability pins | All four are calc parents. |
| Presentation ↔ calculation coverage | Perfect both directions: no calc leaf unpresented, no parent unshown. |

So the two repairs in this branch appear to be the whole of it. Worth stating
as a negative result rather than leaving implied.

### Migration

Neither repair reaches anything provisioned, so `0030` pins both packages.
Two things resync cannot do, handled explicitly:

- it **never deletes** (retirement propagation is out of scope), and
- an association's deterministic id keys on `to_element_id`, so a repointed arc
  is a **new row, not an update** — untouched, a tenant would hold both the pre-
  and post-split arc for the same flow and double-count it.

The five orphaned association ids are deleted by exact id (not by shape, so the
statement cannot reach a tenant-authored arc); replacements arrive as inserts.
Rule ids key on `standard:local_id`, preserved by the repair, so the six rules
update in place — though public needs an explicit `UPDATE`, since
`create_library_rules` skips existing rows.

### Verified locally across 8 tenant schemas

| | public | tenant |
|---|---|---|
| derivation arcs | 18 (was 19) | 17 |
| stale ids remaining | 0 | 0 |
| replacements present | 4 | 4 |
| duplicate (structure, from, to, type) groups | 0 | 0 |
| rollup rules with a pre-split operand | — | 0 |

Tenants delete three of five rather than five because `LongTermDebtNoncurrent`
is in the tenant-exclude set, so the two arcs targeting it were never copied
down. Checked rather than assumed — the first duplicate count looked alarming
at 70 until `structure_id` was included in the grouping, which is the correct
key; public shows 378 by the same loose measure.

jfrench9:
@claude can you review

PR #1139, opened 2026-08-11
refactor(graph): move container refresh onto the instance

jfrench9:
## Summary

Moves the per-instance graph container refresh off the GitHub Actions runner and onto the instance, as a single script (`refresh-graph-container.sh`) an operator can run by hand over SSM. This is phases 1–2 of `specs/data-plane/fleet-refresh-control-plane.md`; the GHA matrix is unchanged, so this is a prerequisite for the later tag-targeted fan-out rather than the fan-out itself.

Writing down what a refresh actually needs surfaced six live defects in the existing path, all of the same shape — a fact duplicated in two places and drifted apart. Two of them (a defeated image-tag pin, and replicas losing volume mounts on restart) were silent and would not have shown up as failures.

## Changes

**New instance-side entry point** — `bin/userdata/common/refresh-graph-container.sh`

- Sources `/etc/environment`, waits out in-flight destructive ops, pulls before stopping, no-ops when the image digest is unchanged, delegates the swap to `run-graph-container.sh`, prunes.
- The busy-counter wait keeps all three fail-open escape hatches (negative counter → idle, stale heartbeat → crashed, missing row → proceed). These are deliberate — `instance_busy` is a coordination signal, not a guard (`ref/data-plane.md` §65) — and the script says so, because a future edit tightening one would look like a bug fix.
- Baked into both writer and replica userdata. Both upload paths already glob `bin/userdata/common/*.sh`, so no upload change was needed.

**Six defects fixed**

1. `aws ssm wait command-executed` is `delay:5 × maxAttempts:20` = **100s**, shorter than a pull plus restart. On expiry it reported failure for a refresh that was still succeeding, and re-running the deploy cycled the container a second time. Replaced with a bounded poll whose budget derives from `max-wait-minutes`.
2. The collector read `WriterTier` with a `describe-tags` call **per instance, inside the loop**, though `describe-instances` had already returned it. Now one `--query`.
3. `graph-maintenance.yml` queried `LadybugRole=shared-writer` — a value nothing sets (all writer tiers are `LadybugRole=writer`, distinguished by `WriterTier`), so it matched nothing on every run. Inert; shared writers were already covered by the query above it.
4. `graph_volume_monitor` filtered `LadybugRole` on `shared_master`/`shared_replica`, which are `NODE_TYPE` values. Also inert, but a trap: it would have silently pulled replicas into the volume-expansion loop that `discover_replica_instance_ids` exists to keep them out of.
5. **Silent.** `ladybug-replica.sh` exported nine mount/profile variables at boot and persisted none to `/etc/environment`, so any restart fell back to defaults and the container came back with **no Lance and no staging mount** — while `LANCE_INDEX_PATH` still pointed into the missing one. The database path survived only because its derivation is relative. This affected the 5-minute health-check cron too, not just refreshes, so a crashed replica was being auto-restarted degraded and unattended.
6. **Silent.** The refresh pulled `${ECR_URI}:${environment}` rather than the `ECR_IMAGE_TAG` the instance recorded. Usually identical — but shared replicas can be pinned to a build tag during a storage-format-breaking engine upgrade, specifically so a boot cannot pull a new engine before the new-format `sec.lbug` is published (`deploy-graph.yml:57`). A refresh in that window defeated the pin during the upgrade it protects.

Also: `refresh-graph`'s `timeout-minutes` was 15 while `max_wait_minutes` defaulted to 30, so a busy instance could never be waited out — now derived from the same input, as are the SSM `executionTimeout` and the poll budget.

**Container-name mapping consolidated** — worth a close look

`CONTAINER_NAME` had **four** derivations: `run-graph-container.sh`, the composite action, `graph-health-check.sh`, and `graph-lifecycle.sh`. The first is now the sole owner and answers `--print-container-name`; the other three ask it. `grep -rn 'graph-api-shared' bin/ .github/` is the cheap invariant.

**Un-cycled instances skip rather than fail** — worth a close look

Common userdata scripts are downloaded from S3 **at boot**, so uploading a new script does not place it on a running instance. Two distinct outcomes, both a GHA `::warning::` with the container left alone:

| Exit | Meaning | Result |
| --- | --- | --- |
| `4` | script not on the instance yet | `skipped-no-script` |
| `3` | script present, `/etc/environment` predates the completeness contract (`GRAPH_ENV_SCHEMA`) | `skipped-stale-env` |

Any other non-zero exit stays a hard failure. Exit 4 comes from an explicit `-x` test in the dispatched command, deliberately **not** inferred from bash's 127 — a missing `docker` inside the script also exits 127, and that is a real failure that must not be downgraded to a warning.

**Consequence for review:** until the graph fleet cycles (or `/etc/environment` is backfilled with a one-shot SSM command), a graph refresh will report `skipped-*` on every instance and change nothing. That is the intended transitional behavior, not a regression — but it means the instance-side path is unproven on a real instance until then, which is why this PR stops at phase 2 and does not retire the existing GHA path.

**Composite action reduced to a trigger** — `refresh-graph-containers`

Its inlined busy-wait, pull/stop sequence, separate health-check step, and separate prune step are gone. Its `aws-account-id` and `health-check-timeout` inputs are removed, along with the now-dead `graph_health_check_timeout` plumbing in `service-refresh.yml`; `prod.yml` and `staging.yml` are updated in the same commit so the reusable-workflow call stays valid.

## Breaking Changes

None.

No application code, GraphQL schema, operations envelope, or REST shape is touched — the diff is CI/CD workflows, EC2 userdata scripts, one Lambda's tag filter, and a test. No SDK regen is implied for either client.

The removed `graph_health_check_timeout` input is an internal reusable-workflow parameter, not a published surface; both in-repo callers are updated in the same commit.

## Testing

- `just test-all` — passed (unit suite, dbt, ruff, format, basedpyright, cf-lint).
- `just test-code` — re-run after the formatter touched the new test file; passed.
- `tests/infrastructure/test_userdata_refresh.py` — **new**, 21 cases, passing. Drives the real script as a bash process against stubbed `aws`/`docker` binaries, with no test hooks added to the production script. Covers every fail-open branch of the busy-counter wait, the digest-skip matrix (including that a digest match on a *stopped* container still restarts), failure propagation, and the pinned tag. It also pins the exit-code contract between the script and the composite action, which have no other coupling — if those drift, a stale instance silently goes from a warning back to a red deploy.
- Both `aws --query` expressions validated against fixtures via `jmespath`; the SSM `--parameters` shorthand validated against the real `awscli` shorthand parser (the embedded `=` in the command string parses correctly).

**Not verified on a real instance.** The script has never run on an EC2 host — that requires a staging deploy plus one hand-run SSM invocation, which is the spec's own phase-2 acceptance criterion and the next step after this merges.

### DIFF
--- a/.github/actions/refresh-graph-containers/action.yml
+++ b/.github/actions/refresh-graph-containers/action.yml
@@ -1,5 +1,10 @@
 name: Refresh Graph Container
-description: Refreshes a single graph API container in-place without replacing the instance
+description: >-
+  Refreshes a single graph API container in-place without replacing the instance,
+  by invoking the instance-side refresh script over SSM. All per-instance logic —
+  the busy-counter wait, the pull, the digest check, the swap, the health check —
+  lives in /usr/local/bin/refresh-graph-container.sh on the instance, not here.
+  This action is only the trigger and the result reporter.
 
 inputs:
   environment:
@@ -18,16 +23,9 @@ inputs:
     description: 'AWS region'
     required: false
     default: 'us-east-1'
-  aws-account-id:
-    description: 'AWS account ID for ECR URI'
-    required: true
-  health-check-timeout:
-    description: 'Seconds to wait for health check after update'
-    required: false
-    default: '30'
   max-wait-minutes:
     description: >-
-      Maximum minutes to wait for in-flight destructive operations
+      Maximum minutes the instance waits for in-flight destructive operations
       (materialization, SEC staging, extensions mat, bulk table creates)
       to complete before cycling the container. Polls every 60 seconds.
     required: false
@@ -44,7 +42,10 @@ inputs:
 outputs:
   status:
     description: 'Update status (success/failed)'
-    value: ${{ steps.update.outputs.status }}
+    value: ${{ steps.refresh.outputs.status }}
+  result:
+    description: 'Whether the container was restarted (updated) or already current (no-op)'
+    value: ${{ steps.refresh.outputs.result }}
   instance-id:
     description: 'Instance ID that was updated'
     value: ${{ inputs.instance-id }}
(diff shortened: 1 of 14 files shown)


jfrench9:
@claude can you review

PR #1141, opened 2026-08-11
feat(graph): retire the per-instance refresh matrix for a fleet command

jfrench9:
## Summary

Completes `specs/data-plane/fleet-refresh-control-plane.md` (phases 3–5) and tears out the old refresh method. The graph container refresh no longer enumerates instances into a GitHub Actions matrix with a runner job each — it issues **one tag-targeted SSM command per node-type group**. That removes a hard 256-job matrix ceiling (`databases_per_instance: 1` makes customer count equal instance count) and stops paying runner-minutes to sleep through a busy-wait that now runs on the instance.

Phases 1–2 landed in #1139, were repaired in #1140, and were validated on a live staging writer. This is the rest.

| | before | after |
|---|---|---|
| 250 instances | ~250 runner jobs, ~50 min | 1 API call, 1 runner job |
| 1000 instances | **impossible** — matrix cap | 1 API call, 1 runner job |
| blast-radius halt | none | `--max-errors` budget |
| failure notification | red job, if watched | SNS page, always |

## Changes

**Phase 3 — replica tag** (`cloudformation/graph-ladybug-replicas.yaml`)

`LadybugRole=replica` alongside the existing `NodeType=shared_replica`, which stays because the app reads it. The value **must not** be `shared_replica`: that would match `graph_volume_monitor.discover_lbug_instances`, whose callers drive volume expansion, and pull the replica fleet into a loop `discover_replica_instance_ids` exists to keep it out of.

**Phase 4 — the Lambda + paging** (`bin/lambda/graph_container_refresh.py`, `cloudformation/graph-infra.yaml`)

- `start` / `status` actions. Dispatch and aggregation only — the 30-minute wait is on the instance, which is exactly why this doesn't need to outlive it (a Lambda that waited would exceed the 15-minute ceiling and pay to idle).
- `ssm:SendCommand` scoped to the document **and** to instances carrying this environment's `Environment`+`Service` tags, so the role can't run shell on arbitrary instances.
- Rate-control defaults live in the Lambda, not just the caller, so a hand-rolled invocation can't hit the fleet at SSM's default concurrency of 50. Deliberately conservative (10%) — the fleet has never been pulled at high concurrency, so raise on evidence.
- EventBridge rule on SSM invocation status-change → `Failed`/`TimedOut` → `InfraAlertTopic`. **Failures now page whether or not anyone is watching the run**, which the old design never did at any fleet size.

**Phase 5 — the collapse** (`service-refresh.yml`, `graph-maintenance.yml`)

- `collect-graph-instances` and the `refresh-graph` matrix become one job that invokes the Lambda, polls to terminal, and writes a per-outcome summary table.
- `.github/actions/refresh-graph-containers/` **deleted**.
- `graph-maintenance.yml`'s serial docker-cleanup loop (one SSM command per instance, waited on in turn) becomes one tag-targeted command — same defect wearing different clothes, including the same 100-second `ssm wait` ceiling.

## Three deviations from the spec, each deliberate

**Targeting is per node-type group, not one unified expression.** SSM ANDs across target keys and ORs within one, so "writers by `LadybugRole` OR replicas by `NodeType`" can't be a single command. Splitting them also answers the spec's own open question in the affirmative — a replica restart is far less customer-visible than a writer's and wants its own rate control. Note the Lambda matches replicas on `NodeType`, **not** the `LadybugRole=replica` added in phase 3, so the fan-out works during that tag's boot-time rollout rather than silently selecting nothing.

**The skip classification had to move, not disappear.** Retiring the composite action removed the code translating exit 3/4 into a warning, and SSM records every non-zero exit as `Failed` — so my first draft of the collapse silently reintroduced the #1140 regression where an un-cycled fleet fails the deploy. `status` now buckets those two codes as `skipped` and reports `failed` separately; the workflow fails only on `failed`. The exit-code contract test moved with it, since that contract is the only coupling between the two files.

**`FORCE_RESTART` is new, and secrets rotation needs it.** `secrets-rotation.yml` refreshes containers monthly with `graph_node_types: "all"` so they re-read rotated credentials — and rotation doesn't change the image, so the digest-skip turned that into a fleet-wide no-op. Credentials come from Secrets Manager cached in-process behind a 1-hour TTL, so the restart *is* the work. This is the digest-skip's one real cost and it's worth paying for the other 99% of refreshes.

Also corrects `secrets_manager`'s module docstring, which claimed an `lru_cache` on the read path that doesn't exist. That false claim is what would lead a reader to conclude rotation never self-heals without a restart — the opposite of the truth, since an LRU would pin a rotated secret for the life of the process and the TTL is what avoids that.

## Scope added after review

`graph-asg-refresh.yml` carried the **same timeout-mismatch defect class** as the two fixed earlier on this branch, and worse: the action's busy-counter wait runs *per tier, sequentially*, each up to `max_wait_minutes`, and a run with no `tier` input iterates every writer tier. Worst case 3×30m of waiting plus a 20m completion wait against a fixed **45-minute** job cap — roughly 2.4× under, making both its "timed out waiting for idle" error and its completion-wait branch unreachable.

- Job timeout now derived by a `budget` job (`max_wait_minutes × tier_count + 20 + 10`), verified to exceed the internal budget at `max_wait` 15/30/60, with and without an explicit tier.
- Wait semantics deliberately unchanged — making the wait one global deadline rather than per-tier is arguably better, but it changes when a hand-run ops tool gives up and belongs in its own change.
- Fixed two drifts in the duplicated busy-counter block: a comment claiming a 2h stale window when `STALE_WINDOW_SECONDS` is 21600 (6h), and the missing unparseable-heartbeat warning that the instance-side implementation already had.
- Both implementations of the `instance_busy` contract now cross-reference each other and state why neither is redundant (per-instance self-check vs. ASG-wide gate from a runner — an instance cannot decide whether to replace itself), citing `ref/data-plane.md` §65.

No structural refactor of the duplication: different execution contexts, different granularity, no scale wall. Trigger to revisit is recorded — a third caller of the contract, or the ASG gate needing to run off-runner.

## Breaking Changes

None to any published surface — no application code, GraphQL schema, operations envelope, or REST shape. No SDK regen implied.

Internal to CI/CD: the `refresh-graph-containers` composite action is deleted (no consumers outside `service-refresh.yml`), and `service-refresh.yml` gains a `force_restart` input. Both in-repo callers are updated in the same commit.

## Testing

- `just test-all` — passed. `just test` explicitly: **12,509 passed**, 23 skipped.
- `uv run actionlint -shellcheck=` — clean.
- `uv run cfn-lint` — clean on both modified templates.
- **New:** 19 Lambda tests (`tests/lambda/test_graph_container_refresh.py`) covering targeting per group, command construction including the exit-4 presence guard, `executionTimeout` exceeding the busy-wait, rate-control defaults, and every branch of the skip classification — including that a real `127` still counts as a failure and isn't downgraded just for being non-zero.
- **Extended:** the shell suite to 24 cases, adding `FORCE_RESTART` overriding a digest match and the inverse (no-op preserved when the override is absent, or every deploy bounces the whole fleet).

**Not verified on real infrastructure.** The Lambda has never run against AWS, and the EventBridge→SNS path is untested end to end. Phase 5 has deleted the fallback, so this needs a staging deploy plus one `service-refresh` dispatch before prod — including confirming the SNS topic policy didn't disturb the ~20 CloudWatch alarms already publishing through `InfraAlertTopic` (a `TopicPolicy` replaces the topic's access policy wholesale, so the default owner statement is reproduced explicitly; that's the one change here with blast radius beyond the refresh path).


jfrench9:
@claude can you review

jfrench9:
@claude can you review

jfrench9:
@claude can you do a final review

PR #1143, opened 2026-08-11
feat(auth): SSO handoff hardening, auth posture endpoint, configurable auth-link targets

jfrench9:
Backend groundwork for centralized login (login-home architecture).

## Changes

- **SSO handoff hardening**: input validation on the exchange return URL, session-state re-validation at completion, aligned replay-marker TTLs, corrected fallback-path payload handling, and a truthful `redirect_url` (`{app}/login`, the entry the frontends actually use). Request/response models unchanged.
- **`GET /v1/auth/providers`** (new): reports which auth methods the deployment offers — password, OIDC, passkeys, registration — from SSM-backed flags (`PASSWORD_AUTH_ENABLED`, `SSO_OIDC_ENABLED`, `SSO_OIDC_PROVIDER_LABEL`, `PASSKEYS_ENABLED`), so one frontend build renders the right posture per deployment.
- **Auth-link email targets**: verification/reset/invitation links can target a designated login-home app (`LOGIN_HOME_APP`) with `return_to` context, behind `AUTH_EMAIL_LINKS_TO_LOGIN_HOME` (default **off** — no behavior change until flipped). Email branding is unchanged; welcome/capacity links untouched.
- **CORS**: allow `X-App-Source` so the login home can attribute requests to the originating product app for email branding.

## Tests

- Net-new coverage for `sso-exchange` and `sso-complete` (previously untested), the providers endpoint flag matrix, and email link targets under both flag states.
- Full unit suite: 12,535 passed. `just test-code` clean.

## Deploy notes

- Fully deployable on its own; the email-link flag stays off until the login-home frontend ships.
- The new endpoint lands in the OpenAPI spec — SDK regen rides a later client minor.


jfrench9:
@claude can you review

jfrench9:
Both findings addressed in 476169db:

1. Relative-path validation now also rejects ASCII tab/newline characters, with test vectors added. The same hardening went into the frontend twin validator in robosystems-core#38.
2. The completion-time session-state comparison now fails closed on unparseable input (401), matching the exchange-time behavior, with a regression test.

On the rate-limit note: `GET /captcha/config` — the pattern this endpoint copies — likewise carries no per-endpoint rate-limit dependency; both are static config reads behind the global protections, so `providers` matches the existing convention. Happy to add one to both in a follow-up if you'd rather they carry it.

PR #1148, opened 2026-08-12
Extract user provisioning into the operations layer

jfrench9:
## Summary

P0 of the enterprise SSO/SCIM arc: account creation moves out of the register endpoint into a shared operations-layer kernel, plus the schema prep it needs. No surface changes; the new tables have no consumers yet.

## Changes

- **`operations/user_provisioning.py`** (new): `provision_user()` — session-in, dataclass-out, domain exceptions. Handles all three org-attachment shapes (personal org, invited org, target org) in one atomic transaction; `register.py` now delegates to it, keeping its gates (validation, CAPTCHA, rate limits, IP protection) and email sending in the router.
- **Atomicity**: `auto_commit` threading through `User.create`, `UserToken.create_token`, `Org.create_personal_org_for_user`, `OrgLimits.create_default_limits` — a failure mid-provisioning now leaves no partial account (previously several steps committed independently).
- **Migrations** (platform, autogenerated ×2): `users.password_hash` nullable + indexed `users.external_id`; new `user_identities` and `scim_tokens` tables.
- **Password reset**: refuses accounts with no password set, behind the existing generic responses.
- **Drive-by**: auth email links skip the redundant `return_to` when the originating app is the login home.

## Tests

- New `tests/operations/test_user_provisioning.py`: all three attachment branches (including the previously unasserted personal-org + limits invariant), case-insensitive duplicate rejection, and a rollback test proving no partial account survives a mid-transaction failure.
- New passwordless-account guard tests in `test_password_reset.py`.
- Full `just test-all` green; both migrations applied cleanly to the local DB.


jfrench9:
@claude can you review

PR #1149, opened 2026-08-12
Add OIDC login for dedicated deployments

jfrench9:
## Summary

P1 (tier 1) of the enterprise SSO arc: a flag-gated OIDC login surface for IdP-governed sign-in on isolated tenant deployments. Ships upstream behind `SSO_OIDC_ENABLED` (default false); the managed platform never mounts it. Builds on the provisioning kernel + schema from #1148.

## Changes

- **`operations/oidc.py`** (new): the login kernel — connection resolver, Valkey-backed single-use flow state (SHA-256 keys, GETDEL), discovery/JWKS fetch + caching, ID-token validation, and `resolve_oidc_user` (link-only, never JIT).
- **`routers/auth/oidc.py`** (new): two browser-redirect endpoints (`include_in_schema=False`). Login mints state/nonce/PKCE and 302s to the IdP (also the app-tile initiated-login URI, with `iss` validated against the configured issuer). Callback validates the token and hands off through the existing session bridge — no JWT minted here.
- **Config**: `SSO_OIDC_ISSUER`/`CLIENT_ID` (plain), `SSO_OIDC_CLIENT_SECRET` (Secrets Manager); a new OIDC rate-limit bucket; an `OIDC_LOGIN_DENIED` audit event; boot validation that requires a full connection when enabled and refuses a deployment with no login method at all.
- **Dependency**: `authlib` (PKCE + token-endpoint client; ID-token validation rides the existing pyjwt + cryptography).

## Security posture

- **Link-only resolution**: an unlinked identity links once by email against the SCIM-provenance predicate (`external_id`), else refuses; a deactivated user is refused even with a valid token (the two-app assignment-drift case).
- **Token validation**: RS256 pinned, signature verified against the issuer JWKS keyed off the trusted configured issuer, required `iss`/`aud`/`exp`/`iat`/`sub`, `azp` checked for multi-audience, constant-time nonce compare.
- **Flow integrity**: single-use state (GETDEL), single-use bridge session, and a browser-binding HttpOnly `SameSite=Lax` cookie so the completing browser must be the one that began the flow.
- Ran `/security-review` on the branch; the two items it surfaced (calibrated `email_verified` handling that stays Entra-compatible, and the browser-binding cookie) are folded in with negative tests.

## Tests

- `tests/operations/test_oidc.py` + `tests/routers/auth/test_oidc.py`: token-validation negatives use **real RSA crypto** through the module's own JWKS path (nonce/iss/aud/exp/alg-confusion rejections), plus the link-only matrix, state single-use/replay, browser-binding, and a producer→consumer seam test proving `sso_complete` accepts the OIDC-written payload. Boot-validation tests for the new config gate.
- Full `just test-all` green.

Not SDK surface (both endpoints `include_in_schema=False`) — no client churn. No SCIM yet; that's the tier-2 follow-up.


jfrench9:
@claude can you review

PR #1150, opened 2026-08-12
Add minimal SCIM 2.0 provisioning for dedicated deployments

jfrench9:
## Summary

Tier 2 of the enterprise SSO arc: a flag-gated SCIM 2.0 surface for IdP-driven user provisioning on isolated tenant deployments. Behind `SCIM_ENABLED` (default false); the managed platform never mounts it. Completes the arc atop #1148 (provisioning kernel + `scim_tokens`) and #1149 (OIDC login).

## Changes

- **`middleware/auth/scim.py`**: per-org bearer dependency against `scim_tokens` (sha256-fingerprint lookup → active/expiry → bcrypt verify). A user JWT or API key is never accepted here; a SCIM token is never accepted by the normal auth deps. Publishes `request.state.scim_org_id`; failures answer with the SCIM error envelope.
- **`routers/scim/`** (`/scim/v2`): conformance discovery endpoints; Users `POST/GET/PUT/PATCH/DELETE` + the `userName eq` filter Okta sends before creating. Create delegates to the provisioning kernel (passwordless, IdP-asserted email, into the token's org at `SSO_DEFAULT_ROLE`); the active toggle delegates to the existing deactivate kill-switch (session invalidation + API-key revocation), tolerating both RFC and Entra PATCH shapes; DELETE maps to an idempotent deactivate, never destructive.
- **Bootstrap**: `operations/admin/scim_bootstrap.py` + admin endpoint + CLI (`scim bootstrap` / `scim revoke-token`) create the enterprise org and mint its token once.
- **Wiring**: `SCIM_ENABLED` + `SSO_DEFAULT_ROLE` config, SCIM rate bucket, SCIM audit events, `/scim` added to the no-store sensitive prefixes.

## Tenant isolation

Every endpoint is confined to the token's org: `{id}` operations gate on `OrgUser.get_by_org_and_user`, list joins+filters on the org, and a cross-org id is a 404 (no existence leak). No IDOR.

## Tests

- `tests/routers/scim/`: bearer auth (no token / bogus token / user-JWT-rejected), conformance, Users CRUD, the filter, org-scoping (cross-org read + list both blocked), and the deactivate path proving `session_version` bump + API-key revocation — including the Entra `"False"`-string PATCH shape.
- `tests/operations/admin/test_scim_bootstrap.py`: org+token mint, reuse, rotation coexistence, revoke.
- Full `just test-all` green.

## Security

Ran `/security-review`: no blocking findings. Two authorization observations (global-vs-org-scoped `active`; 409 uniqueness as a platform-wide email oracle) are benign in the single-tenant model SCIM ships in and require schema changes to address — recorded as design follow-ups, out of scope for minimal tier 2.

Not SDK surface (`include_in_schema=False`) — no client churn.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -31,10 +31,8 @@ LOG_LEVEL=INFO
 # PORT=8000
 
 ## Docker Image Configuration
-## - New users: Keep these to pull pre-built images (~10s download)
-## - Developers: Comment out to build locally from Dockerfile (~3 min build)
-## Images are cached after the first pull — run `just upgrade` to fetch newer ones
-## (works in both modes; a plain `just start` will keep running the cached image)
+## Keep these to pull pre-built images; comment out to build locally from the
+## Dockerfile. Images cache after first pull — `just upgrade` fetches newer ones.
 ROBOSYSTEMS_IMAGE=robofinsystems/robosystems:latest
 ROBOSYSTEMS_PULL_POLICY=if_not_present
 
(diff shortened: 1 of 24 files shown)


jfrench9:
@claude can you review

PR #1151, opened 2026-08-12
feat(auth): harden auth posture gating and SCIM provisioning controls

jfrench9:
## Summary

Hardening pass over the enterprise SSO/SCIM surface (follow-up to #1148/#1149/#1150), from an external design review of the merged work:

- **Enforce the \`PASSWORD_AUTH_ENABLED\` posture at the endpoints.** The flag previously drove only the \`/auth/providers\` advertisement; now every password-credential endpoint (login, register, forgot/validate/reset, password change) refuses with 403 when it is disabled. Invitations do not bypass the guard. No behavior change for deployments with the flag at its default (true).
- **SCIM rate limiting: pre-auth IP bucket + boot checks.** An IP-keyed bucket now sits in front of the per-token bucket on \`/scim/v2\`; boot validation requires rate limiting to be enabled whenever the OIDC/SCIM surface is mounted outside local dev, requires an https issuer, and restricts \`SSO_DEFAULT_ROLE\` to member/admin.
- **SCIM provisioning tokens expire by default** (365d, configurable 1–3650 via admin API/CLI; rotation stays overlap-based). \`/admin\` responses now carry \`Cache-Control: no-store\`.
- **Deactivation reports completeness.** \`User.deactivate\` returns what actually took (keys revoked, cache invalidated); the SCIM active-toggle returns a retryable 503 when side effects don't fully apply and re-asserts them on the IdP's retry instead of short-circuiting.

## Testing

- Full suite green (\`just test-all\`: 12,656 passed + ruff/format/basedpyright clean)
- New coverage: posture-guard parametrized suite, IP-bucket keying, boot-validation pass/fail pairs, token-expiry defaults, incomplete-deactivation 503 + retry convergence
- Security review run on the branch: no findings

### DIFF
--- a/main.py
+++ b/main.py
@@ -87,8 +87,16 @@ logger = get_logger("robosystems.api")
 
 # Path prefixes whose responses may contain per-user secrets (tokens,
 # API keys, billing details, org membership). These get `Cache-Control:
-# no-store` applied in the security-headers middleware.
-_SENSITIVE_PATH_PREFIXES = ("/v1/auth", "/v1/user", "/v1/billing", "/v1/orgs", "/scim")
+# no-store` applied in the security-headers middleware. `/admin` is here for
+# the SCIM bootstrap response, which carries a raw bearer token.
+_SENSITIVE_PATH_PREFIXES = (
+  "/v1/auth",
+  "/v1/user",
+  "/v1/billing",
+  "/v1/orgs",
+  "/scim",
+  "/admin",
+)
 
 
 def csp_variant_for_path(path: str) -> str:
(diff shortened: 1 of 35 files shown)


jfrench9:
@claude review

jfrench9:
Follow-up commit (f64b678f) from a second review pass over the branch:

- **Revocation completeness now reported end-to-end.** The per-key validation-cache invalidation reports success/failure through the whole chain (cache layer → `UserAPIKey.deactivate` → `DeactivationResult.key_caches_invalidated`), and the key sweep covers already-inactive rows so a retried deactivation can finish what an earlier pass left incomplete. The admin status response carries a `fully_applied` field.
- **SCIM PATCH refuses unrecognized payloads** (400 `invalidValue` when no supported operation was applied) instead of acknowledging an update it didn't perform; multi-op payloads that include a supported operation still apply it.
- **SCIM token expiry enforced in the model**: `ScimToken.create` defaults the lifetime rather than allowing a non-expiring mint, and validation refuses any pre-invariant null-expiry row.
- **Both SCIM rate buckets fail closed** when the limiter backend is unavailable, matching the existing auth-endpoint pattern.

Full suite + code quality green.

PR #1152, opened 2026-08-12
docs: realign READMEs and instruction files with the code

jfrench9:
## Summary

An audit checked all 45 tracked READMEs claim-by-claim against source. This PR corrects the seven badly stale ones plus the fourteen the audit named specific errors in — 21 doc files in all. It is **not** the complete set: the audit reported roughly twenty files carrying concrete errors and only enumerated a subset, so some minor drift remains unswept. Every finding was re-verified in source before editing, and four turned out to be wrong and were left alone.

- **`cloudformation/README.md`** was the worst by a wide margin (62 commits behind). Documents the two missing templates; replaces the tier table and every parameter table with pointers to their authoritative source rather than restating values that rot; deletes the "SSM Parameter Pattern" section, since no template creates or reads an SSM parameter (runtime Parameter Store use is now described where it actually happens); corrects the exports against source; and fixes the deploy instructions, since every stack is `workflow_call`-only and five piggyback on another stack's workflow. Net 743 → 677 lines while covering two more templates.
- **`bin/setup/README.md`** — `gha.sh` defaults (prod spot weights are 0, not 90/80), the undocumented SES identity bootstrap step, three `SHARED_REPLICAS_SPOT_*` variables that do not exist, and `postgres-init`'s `extensions` / `extensions_test` databases.
- **`frameworks/`** — the `cm` framework and the `rs-metric` / `rs-driver` packages were undocumented; four advertised reporting styles do not exist (v1 ships Default, Partnership, LLC); Style has been per-entity since extensions migration `0020`, not per-graph.
- **`middleware/`** — five wrong rows in the MCP delegation table, an admission-control section describing the *other* controller, `OTEL_TRACES_ENABLED` documented as a working variable when nothing reads it, and OIDC/SCIM absent from the auth README.
- **`admin`, `dagster`, `schemas`, `adapters/sec`, `graph_api`, `tests`, `operations`** — CLI examples that error as written, a sensor attributed to the wrong job, a class that exists nowhere in the repo, and a CI description that misstated the gate.
- **`config/defaults.py`** — its shared docstring described only one of its two admission-control consumers, which is the likely origin of the `middleware/graph` error above.
- **Tier vocabulary** — retire `Premium` / `Enterprise` for the canonical `Large` / `XLarge`, and refresh three stale price fixtures.

Separately, `fix(ci)` removes two instance-type fallbacks in the graph deploy workflows. Neither is reachable today — the sole caller always supplies the value and the config key resolves — but if either fired it would silently provision the wrong hardware for the whole replica fleet instead of stopping the deploy. The `yq` fallback is now an explicit guard, and the replicas input is required.

## Notes for the reviewer

- **`frameworks/rs-gaap/v1.json` is seed-forward only.** Its `description` is persisted (`Framework.description`), so the corrected text reaches newly seeded deployments only; already-seeded databases keep the old string until a reseed.
- **This fixes instances, not the mechanism.** Nothing here prevents recurrence — no domain owns published surfaces — so the next sizing or pricing change will reproduce this class of drift.
- Known follow-ups in `graph_api` connection-pool configuration, an unregistered MCP tool, and a credit-pricing fallback were found during the sweep and left as code changes for a separate PR. Details are recorded in the design vault rather than here.

## Testing

- `just test-all` green: **12,667 passed, 23 skipped**; ruff, format, basedpyright, cfn-lint and actionlint all clean.
- actionlint baseline unchanged at 28 pre-existing shellcheck infos, none of them at the edited lines, and clean in the `-shellcheck=` form CI actually runs.
- Workflow edits verified against the sole caller and against `yq` resolution for both environments before the fallbacks were removed.

### DIFF
--- a/.github/workflows/deploy-graph-replicas.yml
+++ b/.github/workflows/deploy-graph-replicas.yml
@@ -49,10 +49,9 @@ on:
         required: true
         type: string
       instance_type:
-        description: "EC2 instance type for replicas"
-        required: false
+        description: "EC2 instance type for replicas (authoritative source is .github/configs/graph.yml)"
+        required: true
         type: string
-        default: "r7g.medium"
 
       # Auto-scaling Configuration
       min_instances:
--- a/.github/workflows/deploy-graph.yml
+++ b/.github/workflows/deploy-graph.yml
@@ -429,8 +429,14 @@ jobs:
           echo "📊 Deployment Summary:"
           echo "  - $LBUG_COUNT Graph writers (LadybugDB)"
 
-          # Extract shared replica config from graph.yml
-          REPLICA_INSTANCE_TYPE=$(yq eval ".${ENV_KEY}.replicas[0].instance.type // \"m7g.large\"" .github/configs/graph.yml)
+          # Extract shared replica config from graph.yml. No fallback default:
+          # a silent guess here provisions the wrong hardware for the whole
+          # replica fleet, so a missing key must stop the deploy instead.
+          REPLICA_INSTANCE_TYPE=$(yq eval ".${ENV_KEY}.replicas[0].instance.type" .github/configs/graph.yml)
+          if [ -z "$REPLICA_INSTANCE_TYPE" ] || [ "$REPLICA_INSTANCE_TYPE" = "null" ]; then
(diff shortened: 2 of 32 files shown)


jfrench9:
@claude can you review

PR #1154, opened 2026-08-12
feat(auth): scope identity provisioning and linking to the deployment enterprise org

jfrench9:
## Summary

- Adds an `ENTERPRISE_ORG_ID` pin for dedicated deployments: SCIM bearer scoping, bootstrap targeting, and OIDC link-time org membership all honor it (boot warns while unpinned; enablement sequence is bootstrap → pin → restart)
- First-login identity linking now requires claim equality (`SSO_OIDC_BINDING_CLAIM`, default `sub`) against the SCIM-stamped `external_id`, rather than presence
- SCIM user creation requires a non-empty `externalId` (400 `invalidValue`)
- The SSO connection config block moves into the base deployment secret alongside its client secret (JWT_ISSUER precedent); env vars remain the local-dev path

No schema changes. Flags remain default-off; the managed platform does not mount this surface.

## Testing

- Full suite green (12,683 passed) + ruff/format/basedpyright clean
- New coverage: binding-predicate matrix (mismatch/empty/missing), org-pin refusals across bearer auth, bootstrap, and linking, externalId 400s, validation warning pair

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -135,11 +135,19 @@ INTUIT_REDIRECT_URI=http://localhost:3001/connections/qb-callback
 ## Connection to the customer IdP; enable via SSO_OIDC_ENABLED / SCIM_ENABLED
 ## in FEATURE FLAGS. Issuer is the org authorization server
 ## (e.g. https://<org>.okta.com, not /oauth2/default).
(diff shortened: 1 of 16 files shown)


jfrench9:
@claude can you review

PR #1160, opened 2026-08-13
feat(auth): passkey MFA surface — login handshake, ceremonies, enforcement

jfrench9:
## Summary

The complete passkey MFA (WebAuthn) backend — foundation + surface in one PR (supersedes #1159, whose commits this branch contains):

**Foundation**: `UserPasskey` + `UserMfaRecoveryCode` models and platform migration (`7fd78cf16c54`), the ceremony kernel (`operations/passkeys.py` — single-use flow-scoped challenges, session-in/dataclass-out contract), purpose-scoped short-lived MFA tokens (refused as session bearers), WebAuthn RP ID/origin derived from deployment config with explicit overrides, boot validation rules, `webauthn` (py_webauthn) dependency.

**Surface**:
- **Login handshake**: after password verification, an enrolled user receives `status: mfa_required` with a token redeemed at `/v1/auth/mfa/options` + `/verify` (assertion or recovery code). `AuthResponse` gains additive `status`/`mfa_token` fields (default `authenticated` — register/SSO contracts unchanged).
- **Enrollment**: `/v1/auth/passkeys/register/*` with dual principals — authenticated session (settings) or the forced-enrollment lane, where completing the ceremony completes the login. First passkey returns single-use recovery codes and triggers a notification email.
- **Passwordless login**: `/v1/auth/passkeys/login/*` — user-verified discoverable credentials.
- **Enforcement** (behind `MFA_ENFORCEMENT_ENABLED`): org owner/admin password logins without a passkey route through enrollment before a session is issued. The OIDC lane mints through the SSO bridge and is never challenged.
- **Lifecycle**: list/remove (re-auth required; last passkey of an enforced role protected), recovery-code regeneration, MFA status. Dedicated fail-closed rate buckets, six new audit event types, posture guard mirroring `require_password_auth`.

Everything stays dark: `PASSKEYS_ENABLED` defaults false, every endpoint 403s until it flips, and the login path is byte-identical with the flag off.

**Review follow-ups applied** (from the Claude reviews on #1159 and this PR): user-controlled fields in the notification email are HTML-escaped, the WebAuthn origin is boot-validated alongside the RP ID, the duplicate-credential insert race surfaces as its domain error, and the MFA token helpers gained direct unit tests.

## Testing

- 91 kernel/router tests + 12 posture-drift 403 table + 6 boot-validation + 11 MFA-token unit + 2 email-escaping tests
- Full `just test-all` green (12,788 passed)
- Security review run on the branch: no findings

[...]

PR #1164, opened 2026-08-13
fix(auth): harden passkey enrollment and MFA token redemption

jfrench9:
## Summary

Post-merge hardening for the passkey MFA surface (#1160) and the SSO org boundary, from an internal security review. The whole surface is still dark behind `PASSKEYS_ENABLED` and has not shipped, so this lands before anything is enabled. Detailed dispositions: `specs/security/mfa-2fa.md` (vault).

## Changes

- **Passkey enrollment (`routers/auth/passkeys.py`, `middleware/auth/dependencies.py`, `models/api/passkeys.py`)** — the settings lane now authenticates with a JWT session only (new `get_optional_jwt_user` dependency; API keys read as anonymous on this surface) and requires a fresh re-auth proof — current password or a `reauth`-ceremony assertion — mirroring what passkey removal already required. The forced-enrollment lane is unchanged; its token is its own freshness proof.
- **OIDC resolution (`operations/oidc.py`)** — when `ENTERPRISE_ORG_ID` pins the deployment's org, membership is now required on every resolution, not only when a new identity link is created.
- **MFA token redemption (`routers/auth/mfa.py`, `models/core/user/user_mfa_recovery_code.py`)** — the single-use gate is an atomic SET NX claim taken before the session mints, the attempt budget increments before checking, recovery-code consumption is a single conditional UPDATE, and all three fail closed on store errors.
- **Tests** — atomic-primitive unit tests, enrollment principal/proof matrix (including the API-key and no-proof refusals), and the org-boundary tests now assert membership on both resolution paths.
- Ride-along: `pytest.ini` gains `EXTENSIONS_DATABASE_URL` in the test environment.

## Breaking Changes

None for the published SDKs. The new `password`/`assertion` fields on the enrollment options request are additive (generated tier — a client minor picks them up on the next regen). Behaviorally, settings-lane enrollment now demands a re-auth proof, so robosystems-app's PasskeysCard needs the matching prompt before `PASSKEYS_ENABLED` turns on anywhere — that frontend change is the follow-up, sequenced ahead of any flag flip.

## Testing

- `just test-all`: 11,285 passed (one unrelated, order-flaky materialization test errored in a second run; it passes 3/3 in isolation and the first full run was green).
- Full auth/OIDC/SCIM/passkey surface (577 tests) green; ruff, format, and basedpyright clean on all touched files.
- `/security-review` on the branch: no findings at threshold.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -16,6 +16,7 @@ env =
     LOG_LEVEL=ERROR
     DATABASE_URL=postgresql://postgres:postgres@localhost:5432/robosystems_test
     TEST_DATABASE_URL=postgresql://postgres:postgres@localhost:5432/robosystems_test
+    EXTENSIONS_DATABASE_URL=postgresql://postgres:postgres@localhost:5432/extensions
     POSTGRES_PASSWORD=postgres
     AWS_ENDPOINT_URL=http://localhost:4566
     AWS_REGION=us-east-1
--- a/robosystems/middleware/auth/README.md
+++ b/robosystems/middleware/auth/README.md
@@ -170,8 +170,12 @@ enforcement without passkeys, or that runs OIDC/SCIM/passkeys with
 ### Passkey MFA (WebAuthn)
 
 `PASSKEYS_ENABLED` turns on the passkey surface: enrollment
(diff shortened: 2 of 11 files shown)


jfrench9:
@claude can you reviews

PR #1166, opened 2026-08-13
test(security): black-box tenant-isolation harness (R-11)

jfrench9:
## Summary

Adds a black-box authorization test harness that provisions real test tenants and fires a cross-tenant + privilege-escalation matrix at a live deployment over HTTP, classifying each response PASS / LEAK / INVALID / INCONCLUSIVE. It is the first authenticated test of **R-11 (multi-tenant isolation)** — the load-bearing claim of the product and a HIGH-inherent risk that had never been tested. The harness is opt-in (skips unless `TARGET_API_URL` is set), so it is inert in a normal `just test` / CI run and gates nothing. Also included is a small, non-prod-only auth test-support seam that unblocks the harness's vertical axis.

Run against the local stack this surfaced zero leaks — every cross-tenant probe was denied — while the oracle's own unit tests prove it can detect a leak (it caught a real bug in its first cut mid-development).

## Changes

- **Harness (`tests/security/isolation/`)** — a pytest suite, marked `isolation`/`security`/`integration`, skip-gated on `TARGET_API_URL`:
  - `_provision.py` — provisions two independent tenants (register → login → API key → create graph, both extensions), plus a second graph for tenant A, a subgraph, a graph-scoped key, and (when the flags below are on) a multi-user org role matrix.
  - `_oracle.py` — verdicts computed from API responses only, comparing the attacker's response to the owner's own (the positive control is the truth source). Catches the known leak shape of a 200 carrying another tenant's `org_name` with empty collections, which a status-code check misses.
  - `_http.py` / `_report.py` / `matrix.py` / `conftest.py` — the black-box client, JSON+text report, surface definitions, and fixtures.
  - `test_horizontal.py` — cross-tenant matrix (both directions) over REST reads, Cypher, destructive core ops (delete-graph, materialize, change-tier, backup, subgraph), MCP read/write tools, GraphQL, and the roboledger + roboinvestor extensions command surfaces.
  - `test_scoping.py` — subgraph isolation and graph-scoped API-key enforcement.
  - `test_vertical.py` — privilege escalation within one org (viewer can't write, member can't administer, org-admin has implicit graph admin).
  - `test_oracle.py` — unit tests proving the classifier flags leaks (runs in the normal suite; no live target needed).
  - `README.md` — how to run, the oracle rationale, the connection model.
- **Auth test-support seam** *(commit `c393b363` — please review this commit closely; it is the only production-code change and it touches auth)*:
  - `robosystems/config/env.py` — new `AUTH_INVITE_TOKEN_IN_RESPONSE` flag (default off) and an `expose_invite_token_in_response()` guard that **ANDs the flag with `not is_production()`**, so the token can never surface in a production response regardless of the env var.
  - `robosystems/models/api/orgs.py` — `OrgInvitationResponse` gains a nullable `token` field, populated only under the guard above and documented as test-support only.
  - `robosystems/routers/orgs/invitations.py` — sets that field on the create response when the guard allows, so automated authorization tests can complete the invite → register → role-grant flow without email interception (the raw token is otherwise emailed and stored only as `sha256`).
- **`pytest.ini`** — registers the `isolation` marker.

## Breaking Changes

None. The only API-surface change is additive: `OrgInvitationResponse` gains a nullable `token` field (null in every environment except a non-prod build with the test-support flag on). This is a generated-tier SDK change — an SDK regen opportunity that rides a client minor, not a break; no facade or integration-template symbol is affected.

## Testing

- **Isolation harness** — run against the local stack (`TARGET_API_URL=http://localhost:8000`): **31 tests green**. Horizontal axis = 58 cross-tenant probes, all denied 403 in both directions; scoping and vertical axes green; 16 oracle unit tests green. The vertical axis was exercised with the two flags on locally.
- **Code quality** — `ruff check`, `ruff format`, and `just typecheck` all clean, enforced by the pre-commit hook on every commit (0 errors).
- **`just test-all` (full unit suite)** — not run this session. The harness is opt-in and skips without `TARGET_API_URL`, so it does not change the default suite; the oracle unit tests (`test_oracle.py`) do run in it.

## Certification

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -230,6 +230,10 @@ SECURITY_AUDIT_ENABLED=false
 
 ## Organization
 # ORG_MEMBER_INVITATIONS_ENABLED=false
+# Test-support only (dev/staging): returns the raw org-invite token in the create
+# response so the tenant-isolation harness can provision a multi-user org. Ignored
+# in production (fail-closed). Do not enable in prod.
+# AUTH_INVITE_TOKEN_IN_RESPONSE=false
 
 ## Graph Operations
 # DIRECT_GRAPH_MATERIALIZATION_ENABLED=true
--- a/justfile
+++ b/justfile
@@ -142,6 +142,11 @@ test-full:
 test-cov:
     uv run pytest --cov=robosystems tests/
 
+# Run the tenant-isolation harness (R-11) against a deployment (default: local stack)
+test-isolation target="http://localhost:8000":
+    TARGET_API_URL={{ target }} uv run pytest tests/security/isolation -m isolation \
+        -o addopts="-v -ra" -s
+
 # Run dbt models and tests for an adapter
 test-dbt adapter tmpdir=`mktemp -d`:
     DBT_DUCKDB_PATH="{{ tmpdir }}/{{ adapter }}.duckdb" uv run dbt build \
--- a/pytest.ini
+++ b/pytest.ini
@@ -9,6 +9,7 @@ markers =
     unit: marks unit tests
     slow: marks tests that take longer to run
     security: marks security-related tests
+    isolation: black-box tenant-isolation harness; skipped unless TARGET_API_URL is set (see tests/security/isolation/)
     real_retry_delay: opts out of the zeroed Graph API client backoff (see tests/graph_api/conftest.py)
 
 env =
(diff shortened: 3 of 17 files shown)


jfrench9:
@claude can you review

PR #1168, opened 2026-08-13
fix: purge documents on offboarding + apply org graph limit at checkout

jfrench9:
## Summary

Two pre-flight hardening fixes ahead of a cradle-to-grave tenant lifecycle drill. The first completes data deletion so an offboarded graph leaves no document residue behind; the second applies the organization graph limit at the point of sale, consistent with the graph-creation route.

## Changes

**Graph deprovision — complete document deletion** (`operations/graph/deprovision_service.py`, `config/deprovisioning.py`)

- `Document` rows are now deleted in `_clean_pg_records`. They previously survived teardown: `documents.graph_id` cascades on the `graphs` row, but that row is *soft*-deleted, so the cascade never fired.
- New best-effort `_purge_search_index` step removes the graph's content from the shared OpenSearch index, guarded on `SEMANTIC_SEARCH_ENABLED` so a search-disabled deployment is a clean no-op rather than a status downgrade.
- `DeprovisionResult` now reports `documents_deleted` and `search_purged`.
- Documented `retention_days` on `DeprovisioningConfig` to record that it is the suspend→deprovision sensor window (`dagster/sensors/graph_lifecycle.py`), distinct from the backup-hosting window — the two must not be conflated.

**Billing — harden the graph checkout precondition** (`routers/billing/checkout.py`)

- Apply the organization's graph limit at checkout, before creating the checkout session, so the same limit the graph-creation route enforces is checked at the point of sale. Reuses the existing `OrgLimits` helper so the two paths cannot disagree.
- Scoped to `resource_type == "graph"`; repository purchases are unaffected.

**Tests**

- `test_deprovision_search_purge.py` (new): the purge on the enabled / disabled / failure paths.
- `test_deprovision_service.py`: asserts `Document` rows are gone after teardown; a module-level stub keeps the purge off a real cluster.
- `test_checkout.py`: a refusal-before-payment test (asserts no Stripe customer is created), plus an allow-by-default patch so the existing flow tests exercise the path below the gate.

## Breaking Changes

None. No public API shape changes — the checkout refusal reuses the existing `403` response, and the new `DeprovisionResult` fields are additive and internal. No client SDK regen required.

## Testing

- `just test` (full unit suite): **12,838 passed, 38 skipped, 0 failed**.
- `just test-code` (ruff + format + basedpyright + cf-lint): all clean.
- Pre-commit hooks passed on both commits.

## Certification

### DIFF
--- a/robosystems/config/deprovisioning.py
+++ b/robosystems/config/deprovisioning.py
@@ -7,6 +7,11 @@ from dataclasses import dataclass, field
 class DeprovisioningConfig:
   """Configuration for graph deprovisioning behavior."""
 
+  # The suspend → auto-deprovision grace window. A canceled subscription
+  # suspends the graph; suspended_graph_deprovisioning_sensor
+  # (dagster/sensors/graph_lifecycle.py) only tears it down once
+  # ends_at < now - retention_days, unless the cancellation was IMMEDIATE.
+  # Load-bearing — not the backup hosting window (that is backup_hosting_days).
   retention_days: int = 7
   require_final_backup: bool = True
   backup_delay_hours: int = 24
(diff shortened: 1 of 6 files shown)


jfrench9:
@claude can you review

PR #1175, opened 2026-08-15
fix: four silent-failure defects — disk alarms, fact dedup, SEC manifest, cache invalidation

jfrench9:
## Summary

Four unrelated defects surfaced by a design-doc audit of the forward-work corpus. They share one shape: each is a mechanism that reads as working — a declared alarm, a dedup pass, a documented capability, an invalidation sweep — and does nothing. None of them fails loudly, so none had been noticed.

### CloudWatch — the graph disk alarms watched nothing (`cloudformation/graph-volumes.yaml`, `graph-ladybug-replicas.yaml`)

All four `DISK_USED_PERCENT` alarms declared a `MetricStat` carrying a single `path` dimension. The CloudWatch agent publishes that metric with six (`path`, `InstanceId`, `AutoScalingGroupName`, `InstanceType`, `device`, `fstype`), and a `MetricStat` must match the dimension set **exactly** — so the alarms matched no metric stream anywhere in the graph fleet and sat `OK` on `"no datapoints were received"` for the life of the stack.

Each is now a Metrics Insights query, which aggregates across dimension sets while keeping the path filter. Two fleets, two treatments — this is the part to review:

**Data volumes (`/mnt/ladybug-data`) — writers and shared master.** Thresholds moved to **75 / 85 / 90**. The path filter is load-bearing: this ladder feeds `VolumeMonitorFunction`, which sweeps the volume registry and expands anything past `EXPANSION_THRESHOLD=0.8`, and only attached, registered data volumes are expandable. Live value is **1.09%**, so none of these fire.

**Root volumes (`/`) — shared replicas.** These hold the database on a launch-template device with `DeleteOnTermination`, no autoscaling, absent from the volume registry, so the expansion ladder cannot help them. They get **one notify-only alarm at 90%** in the replicas stack rather than a ladder: the remedy is raising `SHARED_REPLICAS_ROOT_VOLUME_SIZE_<ENV>` and cycling the fleet — one action, so one page.

⚠️ **The replica alarm will be `ALARM` on deploy.** Live value is **89.97%**. That is the intended signal, not a false positive, and it clears when the volume is resized. Sizing is a separate, deliberate decision and is not part of this PR.

Notes on the mechanics, for reviewers:

- Metrics Insights rejects `OR` in `WHERE` and an alarm permits only one Insights query, which is why the two paths are two alarms rather than one.
- Math over a single Insights query *is* permitted — that is what keeps `RATE()` working for `RapidDiskGrowthAlarm`.
- **Validated live against prod with `get-metric-data` before commit**, and the rendered YAML re-checked after quote-escaping. `cfn-lint` does not catch dimension mismatches, so it passes either way.

### SEC statement reads — fact dedup dropped the period type (`operations/roboledger/views/financial_statement_query.py`)

`deduplicate_facts` keyed on `(qname, end_date)` while projecting `duration_type` at the query level and never using it. An annual and a quarterly fact legitimately share a qname and an end date — Q4 and FY both land on the fiscal year end — so the pair collapsed them into whichever row came back first. `ORDER BY end_date DESC` does not break that tie, which made the survivor not merely arbitrary but **unstable between otherwise identical calls**.

Now keyed on `(qname, end_date, period_type, duration_type)`. The sibling path `fact_query.deduplicate_facts` already keyed on the full period identity; this brings the two into line.

**Behavioral note for reviewers:** this is a response-content change, not a schema change. `financial-statement-analysis` (REST view and MCP tool) can now legitimately return both a Q4 and an FY row where it previously returned one. Any consumer assuming one row per `(qname, end_date)` was relying on the defect.

### SEC manifest — an advertised subgraph with no serving path (`adapters/sec/manifest.py`)

`agent_instructions` ships to every connected MCP client and told users to switch to the `sec_historical` subgraph for older periods. That read cannot succeed: the replica stack serves `sec` only, shared-master reads are disabled, and the master runs at desired 0. Its snapshot is also weeks stale and it has no graph-registry row.

Removed rather than left as a documented dead end in front of new users. A comment records the condition for restoring it: a read against that graph actually succeeding.

### Auth cache — invalidation could not match its own entries (`middleware/auth/cache.py`)

`invalidate_user_data` parsed entries with `json.loads` while the write path stores them via `_encrypt_cache_data`, so the sweep could not match any entry and the surrounding `except` swallowed the result. Both the API-key and JWT loops were affected. They now route through `_decrypt_cache_data`, skipping entries that will not decrypt so a single bad entry cannot abort the sweep.

The method had four production callers and **no test**. Adds two regression tests, both confirmed to fail against the previous code.

Checked and deliberately unchanged: the two other `json.loads(cached_data)` sites (`get_cached_graph_access`, `get_cached_jwt_graph_access`) are correct — that cache family is written with plain `json.dumps` and is not encrypted.

## Breaking Changes

**None.** No REST shape, GraphQL schema, or operations-envelope change; the OpenAPI spec is unchanged, so no client regen is required and neither SDK tier is touched.

The statement-read dedup fix changes response *content* (see the behavioral note above) without changing response *shape*. Flagging it here for visibility rather than because it is a contract break — the previous behavior was non-deterministic and could not have been depended on safely.

## Testing

Run this session, all from a local environment:

- `just test` — **12,849 passed, 38 skipped, 0 failed** (5m14s)
- `just test-code` — clean (ruff, format, basedpyright, cf-lint: `0 errors, 0 warnings, 0 notes`)
- `just cf-lint graph-volumes` — clean
- Targeted: `tests/middleware/auth/` + `tests/operations/roboledger/views/` — 318 passed
- **Metrics Insights expressions validated live** against the prod `RoboSystems/Graph/prod` namespace with `cloudwatch get-metric-data` before commit, including the `RATE()` form. This is the check that matters here, because `cfn-lint` is blind to dimension mismatches.
- The two new auth-cache tests were **reverted against the original implementation and confirmed to fail**, then restored.

Not run: deployment of the CloudFormation change (`graph-volumes` deploys through its own workflow, not with the service release).

## Certification

### DIFF
--- a/cloudformation/graph-ladybug-replicas.yaml
+++ b/cloudformation/graph-ladybug-replicas.yaml
@@ -841,6 +841,35 @@ Resources:
           MetricValue: "1"
           Unit: Count
 
+  # Replicas hold the database on the ROOT volume: it is a launch-template device
+  # with DeleteOnTermination and no autoscaling, and it is not in the volume
+  # registry — so the graph-volumes expansion ladder cannot help here and this
+  # alarm is deliberately notify-only. One threshold, not a ladder: the remedy is
+  # raising SHARED_REPLICAS_ROOT_VOLUME_SIZE_<ENV> and cycling the fleet, which is
+  # a deploy, so a warning tier would only page twice for the same action.
+  #
+  # Sizing matters because the download happens at boot: a replica s3 cp's the
+  # whole sec.lbug before it can serve, so once the file no longer fits, every
+  # instance that cycles fails to come up and the daily refresh cycles all of them.
+  # The failure is a cliff, not a slope. Metrics Insights for the same reason as
+  # graph-volumes.yaml — a MetricStat cannot match a partial dimension set.
+  ReplicaRootDiskUsageAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Properties:
+      AlarmName: !Sub robosystems-shared-replicas-${Environment}-root-disk-usage
+      AlarmDescription: Shared replica root volume exceeded 90% - the database no longer has room to grow and a cycling instance may fail to download it
+      Metrics:
+        - Id: worst_root_disk
+          Expression: !Sub 'SELECT MAX(DISK_USED_PERCENT) FROM "RoboSystems/Graph/${Environment}" WHERE path = ''/'''
+          Period: 300
+          ReturnData: true
+      EvaluationPeriods: 2 # 10 minutes sustained; the value is flat, so this only debounces gaps
+      Threshold: 90
+      ComparisonOperator: GreaterThanThreshold
+      AlarmActions:
+        - !Ref ReplicaAlertTopic
+      TreatMissingData: notBreaching
+
   ReplicaContainerStartAlarm:
     Type: AWS::CloudWatch::Alarm
     Properties:
--- a/cloudformation/graph-volumes.yaml
+++ b/cloudformation/graph-volumes.yaml
@@ -619,25 +619,25 @@ Resources:
     Type: AWS::CloudWatch::Alarm
     Properties:
       AlarmName: !Sub "robosystems-graph-volumes-${Environment}-disk-usage-warning"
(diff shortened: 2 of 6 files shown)


jfrench9:
@claude can you review

PR #1190, opened 2026-08-16
fix(auth): revoke cached grants together with the grant row

jfrench9:
## Summary

Hardens the revocation paths so a withdrawn grant stops authorizing at once: every surface that revokes access now clears the auth-cache entries derived from that grant in the same step, and the off-boarding sweep revokes every repository grant a member still holds. Detailed notes: `local/RoboSystems/specs/security/org-multi-user-hardening.md` (§1, the revocation cluster).

## Changes

- **`middleware/auth/cache.py`** — `invalidate_user_data` now drops each matched API key's per-graph access decisions along with its validation entry (shared helper `_api_key_cache_keys`, also used by `invalidate_api_key`).
- **`models/core/user/user_repository.py`** — `UserRepository.invalidate_access_cache()`, called after `create_access` and `revoke_access` commit; the model is the choke point for every caller. The operation-level helper in `repository_subscription_service.py` is removed in its favour.
- **`operations/billing/repository_subscriptions.py`** — `cancel_user_repository_subscriptions` also revokes any active repository grant with no live subscription behind it (a period-end cancel leaves the grant alive to `expires_at`).
- **`operations/admin/user_deletion.py`** — captures API-key fingerprints before the bulk delete and sweeps the account's API-key and JWT cache entries after commit; incomplete sweeps log at CRITICAL.
- **Tests** — new `tests/security/test_revocation_invariant.py`: seeds the cache in production shape (real `APIKeyCache` over a fake Redis) and asserts after graph-member removal, org-member removal, immediate repository cancel, the off-boarding sweep, and account deletion that no grant row and no cache entry still authorizes. All five revocation cases fail against `main`. Plus one regression test each in `test_cache_comprehensive.py` and `test_repository_subscriptions.py`.

## Breaking Changes

None. No API surface changes; nothing for the SDKs.

## Testing

- `just test-code` — clean (ruff, format, basedpyright).
- Full unit suite (`pytest -m "not slow and not integration"`) — 12,850 passed, 23 skipped.
- New invariant tests verified to fail with the source changes stashed (5 of 6; the sixth is the seed self-check).

## Certification

### DIFF
--- a/robosystems/middleware/auth/cache.py
+++ b/robosystems/middleware/auth/cache.py
@@ -856,6 +856,26 @@ class APIKeyCache:
       logger.error(f"Failed to invalidate JWT user data cache for {user_id}: {e}")
       return False
 
+  def _api_key_cache_keys(self, api_key_hash: str) -> list[str]:
+    """Every Redis key holding state for one API key: the validation entry,
+    its signature, and each cached per-graph access decision.
+
+    Revocation must drop all of them together. The per-graph entries are the
+    ones a partial sweep misses: ``validate_api_key_with_graph`` short-circuits
+    on a cached allow, so a surviving ``apikey_graph:`` entry keeps a revoked
+    key inside the graph until TTL even after the validation entry is gone.
+    """
+    api_key_cache_key = self._get_api_key_cache_key(api_key_hash)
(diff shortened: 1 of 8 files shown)


jfrench9:
@claude please review this PR

PR #1191, opened 2026-08-16
fix(graphs): resolve graph-scoped reads on the graph role, not org membership

jfrench9:
## Summary

Three graph-scoped reads that only separate from org membership once an org has a second member: the graph subscription read now resolves the graph role, minting a backup download URL for a dedicated graph requires the admin role (matching create and restore), and subgraph access/visibility follows the parent grant rather than a row of its own. Companion to #1190; both are the current build in `local/RoboSystems/specs/security/org-multi-user-hardening.md`.

## Changes

- **`routers/graphs/subscriptions.py`** — `GET /v1/graphs/{graph_id}/subscriptions` authorizes via `GraphUser.user_has_access` (explicit grant, or implicit org owner/admin) instead of `OrgUser` membership.
- **`routers/graphs/backups/download.py`** — dedicated-graph downloads pass through `verify_admin_access` before the tier-limit check; shared repositories keep their subscription-plan gate. Docstring/OpenAPI text updated to state the requirement.
- **`operations/graph/subgraph_service.py`** — `create_subgraph` no longer writes a `GraphUser` row on the subgraph id. `get_effective_role` already resolves a subgraph to its parent grant, so the row was never read for authorization and outlived the parent grant when a member was removed.
- **`routers/graphs/main.py`** — `GET /v1/graphs` lists subgraphs under every parent the caller holds a grant on, at the parent's role, and ignores any legacy row on a subgraph id (no duplicates, no different role). Reviewers: this widens *listing* to what authorization already allows — previously only the subgraph's creator saw it in this list; the three frontends filter subgraphs out of this list and use `listSubgraphs` instead.
- **`routers/orgs/main.py`** — `_visible_org_graphs` includes subgraphs of granted parents (`or_(graph_id IN granted, parent_graph_id IN granted)`).
- **`migrations/platform/versions/a40f29cc6fe7_…`** — data-only backfill deleting `graph_users` rows on subgraph ids. Skeleton from `just migrate-create`, body written by hand (autogenerate has nothing to detect); downgrade is a no-op.
- **Tests** — new `tests/routers/graphs/test_graph_read_authorization.py` (9 tests; 6 fail against `main`). Four existing dedicated-graph tests in `test_backup_download.py` repointed from the removed `get_by_user_id` patch to `GraphUser.user_has_admin_access` so the real gate is executed.

## Breaking Changes

None. Response shapes are unchanged; `isSubgraph`/`parentGraphId` are existing fields. Nothing for the SDKs.

## Testing

- `just test-code` — clean.
- Full unit suite (`pytest -m "not slow and not integration"`) — 12,852 passed, 23 skipped.
- Migration applied locally (`just migrate-up` → `a40f29cc6fe7 (head)`).

## Certification

### DIFF
--- a/migrations/platform/versions/a40f29cc6fe7_drop_subgraph_scoped_graph_user_rows.py
+++ b/migrations/platform/versions/a40f29cc6fe7_drop_subgraph_scoped_graph_user_rows.py
@@ -1,0 +1,35 @@
+"""drop subgraph scoped graph_user rows
+
+Revision ID: a40f29cc6fe7
+Revises: 7fd78cf16c54
+Create Date: 2026-08-15 21:43:34.702797
+
+"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "a40f29cc6fe7"
+down_revision = "7fd78cf16c54"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+  # Data-only backfill; autogenerate has nothing to detect.
+  #
+  # Subgraph creation used to write a GraphUser row on the subgraph id for the
+  # creator. Authorization never read it (GraphUser.get_effective_role resolves
+  # a subgraph to its parent grant), and member removal never deleted it, so it
+  # outlived the parent grant. Creation no longer writes it and the graph list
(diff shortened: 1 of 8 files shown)


jfrench9:
@claude please review this PR

PR #1192, opened 2026-08-16
fix(graphs): remove connections and register the final backup at teardown

jfrench9:
## Summary

Graph deprovisioning left two loose ends: connection records were never removed with the graph they belonged to, and the final backup taken during teardown was written to S3 but never registered, so nothing could reach it. This closes both — offboarding now removes what it should, and keeps reachable the one archive it exists to hand back.

It also corrects two S3 lifecycle rules in the user bucket that pointed at prefixes which have never existed, so pre-ingestion uploads were covered by nothing and accumulated indefinitely.

All three were found while reviewing the data deletion and retention design notes against what the code actually does.

## Changes

**`operations/graph/deprovision_service.py`**

- `_clean_pg_records` now deletes `Connection` and `ConnectionCredentials` for the graph. Credentials are removed first and by id: `connection_credentials.connection_id` carries no foreign key, so no cascade reaches it and removing the connection rows first would leave its rows with nothing to find them by. Soft-deleted connections are included, since the graph is going either way.
- `_create_final_backup` now records the backup as a `GraphBackup` row via a new `_register_final_backup`. The row is built and flushed rather than going through `GraphBackup.create` / `complete_backup` — both of those commit, and this runs inside a best-effort step of a caller-owned transaction, where an early commit would persist a partial teardown if a later step failed.
- `expires_at` follows the tier's backup hosting window, not the shorter retrieval window. The cleanup job deletes objects past `expires_at`, so a shorter value here would remove the archive early; retrieval and retention are separate promises. The object's lifetime is unchanged either way — while untracked these resolved to the 90-day maximum through the orphan sweep.
- `DeprovisionResult` gains `connections_deleted` and `backup_registered`, and the operator log line now carries `documents_deleted`, `connections_deleted` and `search_purged`. Two of those fields already existed but never reached the log, so it could not state what teardown actually removed.

**`models/core/graph/graph_backup.py`**

- New `BackupInitiator.FINAL`. Deliberately not `SYSTEM`: that initiator is filtered out of the customer listing by design, as an artifact the customer cannot act on — the opposite of this row. `FINAL` passes the existing `!= SYSTEM` filter with no change to the listing query.

**`routers/graphs/backups/download.py`**

- The tier-limit lookup now resolves deprovisioned graphs. The default lookup skips them, which returned 404 *after* both access checks had already passed. **Worth a close look:** authorization is unchanged — `verify_admin_access` still gates the endpoint on graph admin, which after teardown only an org's owners and admins hold.

**`cloudformation/s3.yaml`**

- The user bucket carried `ExpireStagingData` (`staging/`, 7 days) and `ExpireExports` (`exports/`, 30 days). Neither prefix exists in that bucket — verified against prod, which holds exactly `graph-backups/`, `report-bundles/`, `shared-repositories/` and `user-staging/`. S3 prefix matching is literal, so both rules were inert and `user-staging/` was covered by nothing.
- Replaced with one rule on `user-staging/` at 30 days. These objects are ingestion input, not a system of record — the graph holds the result. 30 rather than the old rule's 7 because that number was written against a prefix nothing ever wrote to, so it was never sized against a real upload-to-ingest gap.
- **`report-bundles/` is deliberately left uncovered**, now with a comment recording why: those objects are live report publications with an explicit application-level delete on withdrawal (`delete_report_artifacts`), and a time-based rule would remove the published artifact of a report whose row still exists. Absence of coverage is the correct state there.
- **Deployment note:** the existing objects under `user-staging/` are all older than 30 days, so the first lifecycle evaluation after deploy will expire them. The bucket is versioned, so that writes delete markers and `ExpireNoncurrentVersions` reclaims the space ~30 days later.

**`models/api/graphs/backups.py`**

- The `initiated_by` description now covers the new value, which would otherwise reach the listing undescribed.

## Breaking Changes

None.

No SDK coordination needed. `initiated_by` is typed `str` in the API model, so the new value is data rather than a schema change, and the OpenAPI shape is unchanged apart from that field's description text. `DeprovisionResult` is internal and is not part of any response model. Everything else is additive.

## Testing

- `just test` — 12,921 passed, 38 skipped, 0 failed (5m09s).
- `just test-code` — clean (ruff, format, basedpyright, cf-lint), re-run after each subsequent commit including the template change; `just cf-lint s3` passes.
- `tests/operations/graph/test_deprovision_service.py` extended: teardown now asserts zero connection rows **and** zero credential rows, and a new test asserts the final backup is registered, completed, not `SYSTEM`-initiated, and expiring at the hosting window rather than a shorter one.
- One test-hygiene fix included: the existing deprovision test used a bare `MagicMock` for backup metadata, which passes every attribute access and then fails at flush — swallowed by the best-effort backup step. It now uses values a database column accepts, so the registration path is genuinely exercised rather than silently skipped.

## Certification

### DIFF
--- a/cloudformation/s3.yaml
+++ b/cloudformation/s3.yaml
@@ -204,16 +204,27 @@ Resources:
         Status: Enabled
       LifecycleConfiguration:
         Rules:
-          # User staging data expires after 7 days
-          - Id: ExpireStagingData
+          # Pre-ingestion uploads expire after 30 days. The prefix is
+          # user-staging/, not staging/: S3 prefix matching is literal, so the
+          # two rules this replaces (staging/ 7d and exports/ 30d) matched no
+          # object that has ever existed in this bucket, and source files sat
+          # here indefinitely — including a departed customer's, since
+          # deprovisioning does not touch S3. The objects are ingestion input,
+          # not a system of record; the graph holds the result. 30 days rather
+          # than the old rule's 7 because that number was written against a
+          # prefix nothing wrote to, so it was never sized against a real
+          # upload-to-ingest gap. Note GraphFile rows outlive the object and
+          # still report s3_key, so expiry leaves that metadata dangling.
+          - Id: ExpireUserStagingData
             Status: Enabled
-            Prefix: staging/
-            ExpirationInDays: 7
-          # User exports expire after 30 days
-          - Id: ExpireExports
-            Status: Enabled
-            Prefix: exports/
+            Prefix: user-staging/
             ExpirationInDays: 30
+          # report-bundles/ deliberately has no expiration rule. Those objects
+          # are live report publications, deleted explicitly when a report is
+          # withdrawn (delete_report_artifacts in the roboledger reports
+          # commands). A time-based rule here would destroy the published
+          # artifact of a report whose row still exists. Absence of coverage
+          # is the correct state — do not "fix" it.
           # Graph backups expire after 90 days (max tier retention)
           - Id: ExpireGraphBackups
             Status: Enabled
--- a/robosystems/config/deprovisioning.py
+++ b/robosystems/config/deprovisioning.py
@@ -17,11 +17,16 @@ class DeprovisioningConfig:
   backup_delay_hours: int = 24
   # 90 days for every tier: the S3 lifecycle rule (cloudformation/s3.yaml,
   # ExpireGraphBackups: 90) deletes the object then regardless of what is
-  # promised here, and the final backup creates no GraphBackup row, so the
-  # tier-aware cleanup job cannot extend it either. This table once promised
-  # 180/365 days to Large/XLarge — hosting the infrastructure could not
-  # deliver. Extending it for real means exempting final backups from the
-  # lifecycle rule and tracking them in GraphBackup.
+  # promised here. This table once promised 180/365 days to Large/XLarge —
+  # hosting the infrastructure could not deliver.
+  #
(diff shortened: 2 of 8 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Both review findings applied in `e7ee1fed`.

**1. Transaction wedge — fixed, but not the way suggested.** The finding is correct: `_register_final_backup`'s flush is the first database write in the teardown, so on PostgreSQL a failure there aborts the transaction and every later step runs against a dead session.

The proposed remedy — `session.rollback()` in the except branch — would be worse than the bug here. `dagster/jobs/graph_lifecycle.py` opens **one session** and commits only after the loop, so a rollback would discard the completed teardowns of every graph already processed in that run. Those graphs' databases, extensions schemas and registry slots are already destroyed, so their records would be rolled back to looking live over deleted infrastructure — a worse failure than the wedge.

Wrapped the write in `session.begin_nested()` instead, which contains the failure to that row and leaves the outer transaction usable.

Regression test added, and verified non-vacuous: with the savepoint removed it fails with `PendingRollbackError`, so it holds the containment rather than describing it.

**2. Stale config comment — fixed.** `config/deprovisioning.py` still said the final backup creates no `GraphBackup` row. Rewritten to state what is now true, including that per-tier hosting windows *would* now take effect but stay capped by the lifecycle rule's 90-day ceiling.

Full suite: 12,922 passed, 38 skipped, 0 failed. `just test-code` clean.

PR #1193, opened 2026-08-16
fix(auth): retire revoked API keys from the listing and the sweep

jfrench9:
## Summary

Revoked API keys had nowhere to go. Revocation is a soft flip (`is_active = False`), the listing endpoint returned every row a user had ever created, and the hourly cleanup job only deleted keys past an `expires_at` — a column that is optional at creation and therefore null for keys minted from the UI. The result is that rounds of key provisioning leave the settings screen dominated by dead keys, permanently, with no action available on any of those rows.

This makes the listing active-only and gives the hourly sweep a second criterion so the rows are actually retired.

## Changes

**`robosystems/routers/user/api_keys.py`**

- The listing now queries `get_active_by_user_id` instead of `get_by_user_id`. Revoked keys are excluded at the query, not in the response.
- The filter is applied in the handler only. `revoke_api_key` and `update_api_key` continue to resolve keys through the unfiltered lookup, deliberately: revoking an already-revoked key must find it and re-assert the cache invalidation, which is what the endpoint's 503-retry path relies on to converge.
- The `api_keys_listed` business event dropped its `total_keys` / `inactive_keys` split, which is now constant.

**`robosystems/dagster/jobs/infrastructure.py`**

- `cleanup_expired_api_keys` → `cleanup_stale_api_keys`, now also sweeping keys revoked more than `REVOKED_KEY_RETENTION_DAYS` (30) ago, dated off `updated_at`.
- A key that lapsed by date has its cached validation entry cleared before its row is dropped. Worth a close look: the cache is keyed on the row's fingerprint, so dropping the row first would leave an entry nothing can target. A key whose cache clear fails is counted as deferred and left for the next run rather than deleted blind — the same fail-loud contract the revoke endpoint uses.
- The revoked branch skips that clear. It runs two `KEYS` patterns against the shared cache, which Valkey serves single-threaded; against a five-minute validation TTL and a thirty-day retention window, it can only ever scan for an entry that lapsed weeks earlier. The cost was bounded while it ran once per interactive revocation and would have become a loop over every stale key.
- Return value gains `deferred_count`.

Note that filtering and sweeping are independent: the listing hides a revoked key immediately, and the row survives its retention window regardless.

## Breaking Changes

None. The response model and parameters of `listUserAPIKeys` are unchanged, so no client regeneration is required — the only OpenAPI movement is an added `description` string on the endpoint, which is documentation-only and touches neither tier of the SDK contract.

API consumers reading this endpoint should know the behavior change is real even though the shape is not: revoked keys no longer appear in the response.

## Testing

- `just test` — 12,925 passed, 38 skipped.
- `just test-code` — clean (ruff, format, basedpyright, cf-lint).
- Both new tests were verified to fail against the pre-change source: the sweep test with the revoked branch removed from the predicate, the listing test with the handler pointed back at the unfiltered lookup.
- The sweep's predicate is covered against a real database (which rows survive, which are taken); the deferral and the invalidate-before-delete ordering are covered with a mocked session.

## Certification

### DIFF
--- a/robosystems/dagster/jobs/infrastructure.py
+++ b/robosystems/dagster/jobs/infrastructure.py
@@ -50,11 +50,11 @@ REVOKED_KEY_RETENTION_DAYS = 30
 def cleanup_stale_api_keys(context: OpExecutionContext, db: DatabaseResource) -> dict:
   """Delete API keys that are past their expiry or long revoked.
 
-  Each key's cached validation entry is cleared before its row is dropped:
-  the cache is keyed on the row's fingerprint, so once the row is gone a
-  surviving entry can never be targeted and would keep the key authenticating
-  until its TTL with nothing left to revoke. A key whose cache clear fails is
-  left in place for the next run rather than deleted blind.
+  A key that lapsed by date is cleared from the validation cache before its
+  row is dropped: the cache is keyed on the row's fingerprint, so once the
+  row is gone a surviving entry can never be targeted and would keep the key
+  authenticating until its TTL with nothing left to revoke. A key whose cache
+  clear fails is left in place for the next run rather than deleted blind.
   """
   with db.get_session() as session:
     now = datetime.now(UTC)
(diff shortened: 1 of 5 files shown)


jfrench9:
@claude please review this PR

PR #1194, opened 2026-08-16
fix(graph): exempt shared repositories from the orphan-registration sweep

jfrench9:
## Summary

`robosystems-graph-prod-orphaned-graph-registrations` went to ALARM on its first evaluation after deploy, for a healthy fleet. The shared master is parked to zero between SEC ingestion runs, so the instance named by the `sec` graph-registry row is absent from the instance registry most of the day — and the nightly sweep read that resting state as routing drift.

A shared repository's registry row is bookkeeping, not routing, so the stale `instance_id` is never dereferenced. This exempts those rows from the orphan check.

## Changes

**`operations/graph/infrastructure.py`**

- `cleanup_stale_graphs` no longer counts or stamps a row whose `graph_id` is a shared repository or a subgraph of one.
- The predicate is `is_shared_repository_or_subgraph` — the same one `GraphClientFactory._create_shared_repository_client` uses to take the ALB path — rather than the row's own `repository_type` field. **Worth a look:** this is the load-bearing choice. Reusing the router's predicate means the exemption and the routing decision cannot disagree about which graphs it covers, and it picks up subgraphs (`sec_historical`) without a second rule. The row field would have been self-contained but could drift from what the router actually does, and depends on the registering userdata having written it.
- The exemption is expressed as an extra conjunct on `instance_missing` rather than a `continue`, so an exempt row that still carries a marker falls through to the existing clear branch and self-heals on the next sweep. The live `sec` row carries one now; no manual registry write is needed to clean it up.
- That clear branch's log line said "instance … is back in the registry", which is not true for the exemption path. It now states the outcome without asserting a cause.

**`dagster/jobs/infrastructure.py`**

- Op docstring records the exemption alongside the existing mark-never-delete rule.

**`tests/operations/graph/test_infrastructure.py`**

- Three cases added. Two of them drive the real `sec` / `sec_historical` ids through the actual manifest registry rather than a patched predicate, so they fail if the registry ever stops recognising the deployed shared repo.
- The third asserts a user graph orphaned in the *same* sweep as an exempt row is still marked and counted — the exemption must not widen into a general suppression.

No CloudFormation change. The alarm's threshold and description are still correct; only what fed the metric was wrong. It clears on the first post-deploy sweep that publishes zero.

## Breaking Changes

None.

No SDK coordination needed. The change is confined to a scheduled Dagster sweep and its operation; no REST shape, GraphQL schema, operations envelope, or response model is touched, and the OpenAPI spec is unchanged.

## Testing

- `just test-code` — clean (ruff, format, basedpyright, cf-lint), re-run after the commit.
- `uv run pytest tests/operations/graph/ tests/dagster/` — 890 passed.
- Full `just test-all` not run in this session; the change is scoped to one operation and its Dagster op, both covered above.
- Verified against prod before writing the fix: the `sec` row points at `i-062f857aca359af01`, terminated at 13:30Z on 2026-08-15 when the shared-writer ASG went 1→0; ASG history shows the same wake/sleep cycle daily, with no wake window overlapping either sweep hour (weekly 02:00 UTC, daily 03:00 UTC). Every other registry row is clean and both shared replicas are InService.

## Certification

### DIFF
--- a/robosystems/dagster/jobs/infrastructure.py
+++ b/robosystems/dagster/jobs/infrastructure.py
@@ -287,7 +287,9 @@ def cleanup_stale_registry_entries(context: OpExecutionContext) -> dict[str, Any
   Removes entries marked deleted more than 7 days ago. Entries whose
   instance_id is missing from the instance registry are marked and counted
   (``OrphanedGraphRegistrations``), never removed — the row is a live graph's
-  routing and the instance registry drifts on ASG cycling.
+  routing and the instance registry drifts on ASG cycling. Shared
+  repositories are exempt: they route via ALB, not via their registry row,
+  and their master is parked to zero between ingestion runs.
   """
   from robosystems.operations.graph.infrastructure import InstanceMonitor
 
--- a/robosystems/operations/graph/infrastructure.py
+++ b/robosystems/operations/graph/infrastructure.py
@@ -28,6 +28,7 @@ import boto3
 from botocore.exceptions import ClientError
 
 from robosystems.config import env
+from robosystems.config.shared_repositories import is_shared_repository_or_subgraph
 from robosystems.logger import logger
 
 if TYPE_CHECKING:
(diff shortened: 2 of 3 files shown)


jfrench9:
@claude please review this PR

PR #1195, opened 2026-08-16
fix(orgs): close out an account when its last membership goes

jfrench9:
## Summary

A user belongs to exactly one organization, so removing their last membership left an account in a state nothing handled: able to authenticate, unable to reach anything, and with no way to create or join an organization from the outside. This makes that outcome explicit — the account is closed out on removal — and fixes the two things that made the state hard to act on.

Found while running the org multi-user acceptance walk against production; details in `specs/security/org-multi-user-hardening.md` (F16).

## Changes

**`routers/orgs/members.py`** — when a removal leaves the member with no organization, the account is routed through the existing deactivation switch (which also bumps the session version and revokes API keys, so no credential outlives the membership that justified it). Reversible by an admin; freeing the address for a fresh signup remains a deletion. The removal record gains the outcome, and now reports the grants actually revoked rather than the organization's graph count.

**`routers/admin/users.py`, `models/api/admin/users.py`** — the user listing inner-joined membership, so an account belonging to no organization could not appear in it. That is exactly the population support is asked to act on, and the row builder already handled a missing membership; only the join disagreed. Now an outer join, and the response carries the account's active flag so a live account is distinguishable from a closed one.

**`routers/graphs/main.py`** — graph creation raised 500 for a caller with no organization, where every other organization-resolving surface answers 403. It is a reachable state, not a fault, and the 500 counted against the error rate the alarms watch. Now 403 with the same code.

## Breaking Changes

None. The admin user listing gains a field and returns rows it previously omitted; no field changed shape or meaning. `POST /v1/graphs` returns 403 instead of 500 for a caller with no organization — the error code is unchanged, and no client branches on it.

SDK: additive on the generated tier (`is_active` on the admin user response). Nothing in the stable tier is touched, so this rides a client minor at the next regen.

## Testing

- `just test` — full unit suite, **12938 passed, 38 skipped**.
- Nine new tests. The five covering deactivation and the three covering the listing were each confirmed to **fail against this branch's parent** before the fix, so they test the behavior rather than restating it.
- One pre-existing failure (`admin/test_orgs.py::test_list_orgs_includes_max_graphs`) appears only in partial-suite runs and reproduces identically on pristine `main`; it passes alone and in full-suite order.

### DIFF
--- a/robosystems/models/api/admin/users.py
+++ b/robosystems/models/api/admin/users.py
@@ -13,7 +13,18 @@ class UserResponse(BaseModel):
   email: str
   name: str | None
   email_verified: bool
-  org_id: str
+  is_active: bool = Field(
+    default=True,
+    description=(
+      "Whether the account can authenticate. Removal from a last organization "
+      "deactivates the account, so an inactive user with no org is the shape "
+      "support sees when asked to free an email."
+    ),
+  )
+  org_id: str = Field(
+    default="",
+    description="Empty when the user belongs to no organization.",
+  )
   org_role: str
   created_at: datetime
   updated_at: datetime
--- a/robosystems/routers/admin/users.py
+++ b/robosystems/routers/admin/users.py
@@ -51,7 +51,12 @@ async def list_users(
   """List all users with optional filters."""
   session = next(get_db_session())
   try:
-    query = session.query(User).join(OrgUser, User.id == OrgUser.user_id)
+    # Outer join: a user with no org membership is exactly who support needs
+    # to find here — removal from a last org deactivates the account and
+    # leaves the email squatted, and freeing it means resolving this email to
+    # an id. An inner join hid precisely that population (the row builder
+    # below already handles a missing membership).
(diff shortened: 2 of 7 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Good catch on the deactivation edge case — fixed in `8c6c8c1`-onwards (latest commit on the branch).

Verified the failure mode as described, and it was slightly worse than reported: the audit record sits *after* the deactivation step, so a failure there also skipped `ORG_MEMBER_REMOVED` for a removal that had already committed. A privileged access change with no evidence is the exact gap this surface closed last week, reached from the other side.

Deactivation is now best-effort — on failure it logs `CRITICAL` naming the repair command, and the removal still returns 204 and is recorded with `user_deactivated: false`. Added a regression test that patches `User.deactivate` to raise and asserts 204 + membership gone + audit fired; confirmed it returns 500 with no audit event against the previous commit.

PR #1196, opened 2026-08-16
fix(forecast): anchor the walk on the seam, and keep forecasts out of the books

jfrench9:
## Summary

Closes items 2 and 3 of the forecast-hardening spec and runs item 1's reader-conformance audit. A scenario now survives a period close without being rebuilt, the verification gate is tested against a forecast that is wrong rather than only against a stubbed verdict, and the audit that was scoped as a precondition for projecting scenario facts into the graph turned up two live OLTP defects instead.

Item 1's projection flip is deliberately **not** taken here. It buys a capability nothing currently asks for, and the exclusion means the graph is limited rather than wrong; the audit was kept because it stands alone.

## Changes

**Seam-anchored balance-sheet roll** (`operations/information_block/forecast_compute.py`, `forecast.py`, `models/api/information_block.py`, `models/api/extensions/forecasts.py`)

`base_period` was doing two jobs: origin of the authored lever window, and the month the walk seeded its opening balances from. Advancing it to keep balances current forced a restatement of every lever onto a shifted window, which `update` rejects without a full re-supply — so keeping a scenario current across a close meant deleting and rebuilding it. Left un-advanced, the forward months opened on balances several closes old.

- `base_anchor: "seam" | "fixed"` on `ForecastMechanics` (default `seam`), threaded through create and update. Switching it restates nothing, unlike `base_period`.
- `_resolve_anchor_period` scans back from the horizon end for the newest month carrying **close-stamped** income-statement **and** balance-sheet sets. Whether the balance sheet is required is decided by what the *base* month carries, so a scenario that rolls one can't re-anchor onto a month that would silently drop the roll. Canonical-only matters: `create_report` consults no fiscal calendar, so a publication snapshot can sit on an open or reopened month and proves nothing about closure.
- The walk computes `anchor+1 … base+horizon`. **The window's end never moves** — a re-anchored walk computes fewer months, never later ones, or a scenario silently grows a horizon its author never asserted levers for. A scenario the close has fully overtaken computes no months, clears its now-closed stale months, and says so in its diagnostics (this raised in the first push — see the review follow-up below).
- `anchor_period` on the response and in each month's `ForecastProvenance`; a re-anchor is announced in `diagnostics` rather than applied quietly (months_computed shorter than months without a halt is a new shape).
- `_invalidate_superseded_head` sweeps months an earlier run computed that have since closed — invisible on the plan grid because actuals win at overlaps, still visible to anything reading the scenario directly. It shares one implementation with the tail sweep (`_sweep_scenario_sets`) so the lever-set exclusion is written once.
- The base-advance step is removed from the close playbook (`middleware/mcp/tools/playbook_tools.py`).

**A test that binds `verification_passed` to a forecast that is wrong** (`tests/operations/information_block/test_forecast_compute.py`)

`TestStopTheWalk` proves the walk's *response* to a verdict by injecting one; nothing proved the real verifier reaches the right verdict. `TestTheRealVerifier` runs the real `_verify_month_sets` — and through it the real rule engine — against a real extensions schema, over a month that is balanced and wrong: `Assets 200 = Liabilities 80 + Equity 120` passes the accounting equation while `Cash 100 + PP&E 50` fails the rollup. That shape is the point; the BS roll closes on a balancing cash plug, so `A = L + E` holds by construction however wrong the walk is.

The `_run` harness now models which months are closed (`closed_months` / `bs_closed_months`) — a stub answering "yes" for every month would report the whole horizon as already closed.

**Rule-engine scenario leakage** (`operations/information_block/rules/engine.py`) — worth close review

Scenario month sets carry the *statement* structure's id, by design: a scenario column renders through the same structure as its actuals. The engine has two unpinned entry points — `evaluate-rules` called with only a `structure_id` (all three narrowing arguments are documented optional) and the period close, which evaluates its statement structures without pinning a FactSet — and both scoped by `structure_id` alone.

- `_bind_variables` takes the newest fact per element when unpinned. On a graph carrying a forecast the newest fact is a month that has not happened, so a check of the books answered about the plan and wrote the verdict onto the books.
- `_load_period_balances` pins to the newest *actual* FactSet but fell back to `structure_id` scope when there wasn't one. Reachable: the walk stamps balance-sheet sets on the statement structure even where no actual balance sheet exists (the degraded, rule-driven branch), so "never reported" and "carries only forecast months" were the same state to that query. The comment there already said "no balances"; the code did something else.
- Both now scope through one shared `_actual_set_ids`. `_bind_sum_variables` gets the same filter — inert today (schedule structures carry no scenario sets) and the third place the invariant has to hold to be one.

**`has_dimensions` was a literal** (`operations/extensions/materialize.py`)

`Fact.has_dimensions` and `dimension_count` were hardcoded `false` / `0` while four graph read paths filter consolidated totals on exactly that column — so the documented contract was being upheld by the scenario exclusion upstream, not by the flag. Anything dimensioned reaching the graph would have passed straight through every one of those filters. Now derived from the `fact_dimensions` junction. Provably inert today: only scenario facts carry a junction row, and those are excluded.

**Review follow-ups** (commits `94719978`, `13cc580e`, `003fbddd`) — see the PR comments for detail: the anchor scan is bounded by one query instead of walking the whole horizon; anchor candidates must be close-stamped rather than merely present; the overtaken path sweeps and returns instead of raising (the exception rolled the sweep back); and the close playbook's account of all this was corrected, since it ships to MCP clients in the same image as the tools it describes.

## Breaking Changes

None.

Additive only on the client SDK surface — `anchor_period` on `ComputeForecastResponse`, `base_anchor` on `ForecastMechanics` / `CreateForecastRequest` / `UpdateForecastRequest`. Generated tier, so it rides a client minor; nothing in the stable tier or the integration template's emit path is touched. An SDK regen opportunity, not a coordination point.

Behaviour note for existing forecast blocks: `base_anchor` defaults to `seam`, so blocks authored before it existed re-anchor as periods close. Checked against the live Harbinger graph — all three scenarios carry `base_period: 2026-06` against `closed_through: 2026-06`, i.e. already seam-anchored, so the default is a no-op there today. Scenarios deliberately built on a counterfactual base want `base_anchor: "fixed"`, which needs no levers re-supplied.

## Testing

- `just test-code` — clean (ruff check, format, basedpyright: 0 errors, cfn-lint, actionlint).
- Full unit suite via `uv run pytest -m "not slow"` — **12,958 passed, 38 skipped**. (Note for anyone reproducing: sourcing `.env.local` sets `ENTERPRISE_ORG_ID`, which CI does not, and that alone fails 36 unrelated SCIM/OIDC tests. Run with only `TEST_DATABASE_URL` set.)
- Both new engine tests were confirmed to **fail before** the fix and pass after; each ships with a control that would catch a filter which simply drops everything.
- The seam-anchoring tests were mutation-checked: with re-anchoring disabled, 7 of the 10 fail and the 3 that pass are exactly the ones asserting unchanged behaviour. The real-verifier tests were mutation-checked twice — forcing the rollup evaluator to pass breaks the main assertion, and a binder that never binds breaks the control.
- The materializer change was verified by **executing the generated staging SQL against real DuckDB + Postgres**, not by reading it: the scenario fact stays excluded, and a deliberately dimensioned actual fact projects as `(true, 1)` where it previously read `(false, 0)`.

## Certification

### DIFF
--- a/robosystems/middleware/mcp/tools/playbook_tools.py
+++ b/robosystems/middleware/mcp/tools/playbook_tools.py
@@ -85,10 +85,13 @@ _RECURRING_SEQUENCE: list[str] = [
   "the new closed_through. Nothing needs re-basing — the walk re-anchors on "
   "the newest closed month by itself and reports the move in its "
   "diagnostics, so the scenario's levers stay as authored. Read "
-  "anchor_period on the response to confirm it advanced; a scenario the "
-  "close has overtaken entirely fails loudly and needs its horizon "
-  "extended. Scenarios deliberately built on a counterfactual base carry "
-  "base_anchor='fixed' and stay put.",
+  "anchor_period on the response to confirm it advanced. A scenario whose "
+  "horizon the close has passed entirely computes no months and says so in "
+  "its diagnostics — its stale months are cleared and its levers kept, so "
+  "extending horizon_months and re-running is all it needs. Scenarios "
+  "deliberately built on a counterfactual base carry base_anchor='fixed' "
+  "and stay put. Note the anchor only ever advances onto a month the close "
+  "stamped: a report published for an open month does not move it.",
 ]
 
 _INITIATE_SEQUENCE: list[str] = [
--- a/robosystems/middleware/mcp/tools/playbook_tools.py
+++ b/robosystems/middleware/mcp/tools/playbook_tools.py
@@ -82,10 +82,13 @@ _RECURRING_SEQUENCE: list[str] = [
   "reads (statements, fact grids) against the newly closed month.",
   "If the tenant runs forecast scenarios: re-run compute-forecast for each "
   "forecast block after the close so the actuals/forecast seam advances to "
-  "the new closed_through. A scenario whose base_period lags closed_through "
-  "re-bases at the seam — keep scenario bases at the current closed_through "
-  "(advance or rebuild them after each close, per the tenant's "
-  "procedures doc).",
+  "the new closed_through. Nothing needs re-basing — the walk re-anchors on "
+  "the newest closed month by itself and reports the move in its "
+  "diagnostics, so the scenario's levers stay as authored. Read "
+  "anchor_period on the response to confirm it advanced; a scenario the "
+  "close has overtaken entirely fails loudly and needs its horizon "
+  "extended. Scenarios deliberately built on a counterfactual base carry "
+  "base_anchor='fixed' and stay put.",
 ]
 
 _INITIATE_SEQUENCE: list[str] = [
--- a/robosystems/models/api/extensions/forecasts.py
+++ b/robosystems/models/api/extensions/forecasts.py
@@ -244,7 +244,21 @@ class CreateForecastRequest(BaseModel):
       "Seed month (``YYYY-MM``) the walk projects forward from. "
       "Defaults to the fiscal calendar's closed-through period, else "
       "the newest actual report month. Resolved and stored at create "
-      "time."
+      "time, and it never moves afterwards — every lever is keyed to a "
+      "month inside ``base_period + 1 … + horizon_months``, so moving it "
+      "would mean restating all of them. ``base_anchor`` decides whether "
+      "the walk still *seeds* here once months close under it."
+    ),
+  )
+  base_anchor: Literal["seam", "fixed"] = Field(
+    "seam",
+    description=(
+      "Where the walk takes its opening balances as periods close. "
+      "``seam`` (default) re-anchors on the newest closed month inside "
+      "the horizon, so the scenario survives a close untouched and its "
+      "first forward month rolls off real balances. ``fixed`` pins the "
+      "walk to ``base_period`` — the deliberate counterfactual, whose "
+      "balances are meant to diverge from actuals."
     ),
   )
   levers: list[LeverAssertionRequest] = Field(
@@ -295,6 +309,14 @@ class UpdateForecastRequest(BaseModel):
   scenario_kind: Literal["budget", "forecast", "projection"] | None = None
   horizon_months: int | None = Field(None, ge=1, le=36)
   base_period: str | None = Field(None, pattern=_PERIOD_PATTERN)
+  base_anchor: Literal["seam", "fixed"] | None = Field(
+    None,
+    description=(
+      "Switch the walk between seam-anchored (default) and pinned to "
+      "``base_period``. Changes nothing about the authored window, so "
+      "unlike ``base_period`` it needs no levers re-supplied."
+    ),
+  )
   levers: list[LeverAssertionRequest] | None = Field(
     None,
     min_length=1,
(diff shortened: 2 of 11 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Review follow-up pushed in `94719978` — the one item raised (the per-candidate-month re-query in `_resolve_anchor_period`).

Worth taking because the scan's cheapest case was also its most expensive: a scenario whose base is still the seam finds nothing, and finding nothing meant probing every month of the horizon to prove it. `_newest_actual_month` now bounds where the scan starts. It's a bound, not an answer — the newest set may be the annual comparative one, which `_actual_set_at` deliberately refuses. Too high is harmless (the scan walks down); too low can't happen, since every monthly set is in the maximum. The scan is otherwise unchanged, so a gap in the closed history still can't strand the anchor.

8 tests, mutation-checked both directions: a bound one month too low fails 7, ignoring the bound fails the 3 asserting the scan stops probing. Also exercised against real Postgres — correct month with a scenario set past the horizon end, annual set taken as a higher bound, entity-isolated.

Everything else in the review was a confirmation; nothing else changed.

jfrench9:
Both findings were real and are fixed in `13cc580e`. Verified each against the code before changing anything.

**P1 — publication snapshots accepted as closed months.** Confirmed the mechanism end to end: `create_report` has *zero* fiscal-calendar references, so it takes whatever window the caller passes and a report generated for an open or future month leaves a valid snapshot; and reopen retraction (`statement_sets._canonical_set_ids_in_window`) deletes only `report_id IS NULL AND scenario_id IS NULL`, explicitly leaving snapshots. So the anchor could land on a month nobody closed — and it has teeth, because the walk seeds there *and* `_invalidate_superseded_head` deletes every scenario month at or before it.

The distinction you drew is the right one: preferring canonical is correct when seeding from a month already known to be closed; it isn't enough when the question *is* whether the month is closed. Anchor candidates and the scan's upper bound now require `report_id IS NULL`. Close-stamping tracks reopen for free, so no calendar clamp was needed — and I deliberately didn't add one, since a baseline-initialised `closed_through` would *loosen* the predicate on tenants that never closed month-by-month. Publication fallback untouched for the authored base.

**P2 — overtaken scenarios skipped the head sweep.** Correct, and the reason it couldn't be fixed by moving the sweep above the raise is that the exception rolls the session back — sweeping then raising would sweep nothing. It now sweeps and returns zero computed months with a diagnostic that leads with the condition and names the way out. Only derived months go; the lever set is excluded as everywhere else. The limit case now matches the trend instead of inverting it.

**Worth flagging: my first mutation pass didn't hold.** Removing the canonical filter left every test green, because the seam tests stub `_actual_set_at` wholesale — the harness implemented the canonical/publication distinction itself, so it proved the call sites pass the flag, not that the flag does anything. Real-schema tests in `test_forecast_base_binding.py` now cover it. A third mutation (the scan requesting non-canonical sets) also survived: the canonical bound usually hides it, since the first month probed is the bound itself. It stops hiding it when the scan walks *down* — the newest closed month can carry an IS without a BS, and the month below can be reopened. That path has its own test now.

All four mutations fail the suite. Full suite 12,975 passed; `just test-code` clean.

PR #1197, opened 2026-08-17
fix(events): lock every path that decides an event's next status

jfrench9:
## Summary

Every path that decides an event's next status now locks the row it decided on. Three paths make that decision — an inbox approval, the QuickBooks sync's auto-commit pass, and the obligation-promotion sweep — and all three were read-decide-write with nothing held across the decision. Two writers could both read `captured`, both fire the event's handler, and leave one event carrying two sets of GL rows: books that still foot, still render, and no longer say what the event says.

The promotion sweep additionally wrote `classified` over whatever its target row currently held, so an obligation voided or committed since its candidate read was silently reverted.

## Changes

**`operations/event_block/locking.py`** (new) — the wait policy, in one place now that several callers share it. The rule it encodes: **background jobs wait, request handlers do not.** A sync or a Dagster sweep takes its locks unbounded, because a batch should not fail behind a single approval. Request-facing callers wrap their locking work in `bounded_lock_wait`, which sets `SET LOCAL lock_timeout` (3s) and translates SQLSTATE `55P03` into a typed `EventLockedError`. Three seconds rather than `NOWAIT` because most contention is another approval or a short write and resolves in milliseconds — failing those would be a spurious error, and only the long-running writers should produce one. Other `OperationalError`s keep their own identity rather than reaching a caller as "retry in a moment".

**`operations/event_block/commands.py`** — `update_event_block` loads the event `FOR UPDATE`, bounded. Under READ COMMITTED the blocked reader re-reads the committed row once the lock releases, so it evaluates the transition against current state and raises `InvalidEventTransitionError` rather than acting on the snapshot it took before blocking.

**`operations/extensions/loader.py`** — the sync's existing-events lookup in `_capture_transactions_as_events` takes `FOR UPDATE`. Locking at the lookup rather than before each dispatch is deliberate: a row lock is held to transaction end either way, so acquiring it once at the top costs one query instead of one per event and also covers the UPSERT decisions below it. Most of these rows were write-locked incidentally by the UPSERT; the gap was the re-sync whose payload is unchanged, which issues no UPDATE and so took no lock.

**`operations/event_block/promotion.py`** — worth a close look, this is the worse of the two races. `promote_pending_obligations` runs on a Dagster sensor every few minutes on every graph, so it races the on-demand op, an inbox approval, and its own next tick if one overruns. Its candidate load now takes `FOR UPDATE`, and the bulk classify UPDATE states the status it assumes (`status = 'pending'`) instead of filtering on id alone — the same guard the orphan void forty lines above it already carried. Handler idempotency per `(schedule, period)` mitigated the duplicate-dispatch half; it never protected the status.

**`operations/roboledger/commands/schedules.py`** — the on-demand `promote_obligations` wraps the sweep in `bounded_lock_wait`. Note the sweep function is called from both the sensor and this request handler, which is why the bound belongs to the caller rather than to the function.

**`routers/extensions/roboledger/operations.py`** — `EventLockedError` maps to **409** on `update-event-block` and `promote-obligations`. It is the one error on these operations a caller should retry rather than fix.

**Tests** — `tests/operations/event_block/test_update_locking_db.py` is new and DB-backed, running real concurrent sessions against a throwaway tenant schema (the `test_share_controls_db.py` pattern). Mock sessions can prove a keyword was passed; they cannot prove a lock blocks. Three cases: the sync's lock blocks an approval into a 409; an approval that waits out a lock re-reads the committed row and rejects on current state; the sweep's lock blocks the on-demand op.

One detail in that last case is load-bearing and easy to get wrong on a later edit: it seeds a **stranded `classified`** obligation, not a `pending` one. A pending obligation is flipped by the bulk UPDATE, whose own row lock would block a second caller whether or not the candidate read is locked — so the test would pass with the fix reverted. The stranded case makes the sweep write nothing, leaving the read's `FOR UPDATE` as the only thing that can block.

The lost-update guard is unreachable end-to-end once the read is locked, so it is pinned at the statement in `test_promotion.py` rather than dressed up as a race test it cannot fail.

The 40 loader tests that stubbed the existing-events lookup on a bare `MagicMock` now go through a `_stub_event_lookup` helper that stubs the locked and unlocked chains together, so a test cannot silently assert against whichever one production stopped using.

## Breaking Changes

None.

Two operations gain a 409 response (`update-event-block`, `promote-obligations`). Purely additive — no request shape, response body, GraphQL type, or operations-envelope field changes, and nothing is removed. It reaches the SDKs' generated tier only, as an OpenAPI regen opportunity rather than a coordinated change.

Worth knowing for whoever picks up the client side: the 409 has no retry affordance in `roboledger-app` yet, so it currently renders as a generic operation failure. The API contract is complete; the UI follow-up is not in this PR.

## Testing

`just test-all` — **12,983 passed, 38 skipped**; ruff, format, typecheck and cf-lint all clean.

The new DB-backed module is `pytest.mark.integration`, which `just test` deselects locally but CI runs (`-m "not slow"`). Run explicitly this session against the local extensions database: 3 passed. Affected suites together (`operations/event_block`, `operations/extensions`, `operations/roboledger`, `routers/extensions`, integration included): 1,215 passed.

Each of the four guards was verified to be a real regression by removing it and confirming the corresponding test fails:

| Guard removed | Result |
| --- | --- |
| `update_event_block` `FOR UPDATE` | mock assertion fails |
| loader lookup `FOR UPDATE` | `DID NOT RAISE EventLockedError` |
| promotion candidate `FOR UPDATE` | sweep-lock case fails |
| classify UPDATE status predicate | statement assertion fails |

That check earned its place: the first two promotion tests I wrote passed with their guards removed, for the reasons described under Changes, and were rewritten.

## Certification

### DIFF
--- a/robosystems/operations/event_block/__init__.py
+++ b/robosystems/operations/event_block/__init__.py
@@ -6,6 +6,7 @@ path regardless of internal layout.
 
 from .commands import (
   DuplicateEventError,
+  EventLockedError,
   EventNotFoundError,
   InvalidEventTransitionError,
   create_event_block,
@@ -16,6 +17,7 @@ from .commands import (
 
 __all__ = [
   "DuplicateEventError",
+  "EventLockedError",
   "EventNotFoundError",
   "InvalidEventTransitionError",
   "create_event_block",
--- a/robosystems/operations/event_block/__init__.py
+++ b/robosystems/operations/event_block/__init__.py
@@ -6,7 +6,6 @@ path regardless of internal layout.
 
 from .commands import (
   DuplicateEventError,
-  EventLockedError,
   EventNotFoundError,
   InvalidEventTransitionError,
   create_event_block,
@@ -14,6 +13,7 @@ from .commands import (
   preview_event_block,
   update_event_block,
 )
+from .locking import EventLockedError
 
 __all__ = [
   "DuplicateEventError",
--- a/robosystems/operations/event_block/commands.py
+++ b/robosystems/operations/event_block/commands.py
@@ -365,7 +365,15 @@ def update_event_block(
   Handler errors roll back the entire update, including the status
   change — a failed commit leaves the event in its pre-approval state.
   """
-  event = session.get(Event, body.event_id)
+  # Lock the row for the life of the transaction. The transition check below
+  # is read-decide-write, and the decision is only sound if nothing else can
+  # move the event in between. The live race is an inbox approval against the
+  # sync's auto-commit pass (`extensions/loader.py`): both read `captured`,
+  # both fire the handler, and the event ends up with two sets of GL rows —
+  # a ledger that still foots and is still wrong. Under READ COMMITTED the
+  # blocked reader re-reads the committed row once the lock releases, sees
+  # the new status, and raises InvalidEventTransitionError as it should.
+  event = session.get(Event, body.event_id, with_for_update=True)
   if event is None:
     raise EventNotFoundError(f"Event not found: {body.event_id}")
 
--- a/robosystems/operations/event_block/commands.py
+++ b/robosystems/operations/event_block/commands.py
@@ -383,7 +383,14 @@ def update_event_block(
     f"Event {body.event_id} is being written by another process "
     "(most likely a running sync). Retry in a moment.",
   ):
-    event = session.get(Event, body.event_id, with_for_update=True)
+    # `populate_existing` for the same reason as `execute_event_block`: every
+    # production caller opens a fresh session per request, so the identity map
+    # is empty today, but a caller that reused a session would otherwise get
+    # the lock and a stale status — the one combination this guard exists to
+    # prevent.
+    event = session.get(
+      Event, body.event_id, with_for_update=True, populate_existing=True
+    )
   if event is None:
     raise EventNotFoundError(f"Event not found: {body.event_id}")
 
@@ -667,7 +674,27 @@ def execute_event_block(
 
   from .qb_writeback import QBWritebackError, post_event_to_qb
 
-  event = session.query(Event).filter(Event.id == body.event_id).first()
+  # Locked, and for a sharper reason than the other paths: between this read
+  # and the status write below sits a POST to QuickBooks. A concurrent void or
+  # supersede would otherwise be clobbered by the `fulfilled`/`pending` stamp
+  # *after* the external write already happened — QB's request_id dedup covers
+  # a repeated post, not an event that was voided and published anyway.
+  #
+  # `populate_existing` is load-bearing, not belt-and-braces: close_service's
+  # QB pre-publish loop calls this on a **shared** session that already loaded
+  # these events, so without it the lock would be taken while `event.status`
+  # kept the value it held before blocking.
+  with bounded_lock_wait(
+    session,
+    f"Event {body.event_id} is being written by another process. Retry in a moment.",
+  ):
+    event = (
+      session.query(Event)
+      .filter(Event.id == body.event_id)
+      .populate_existing()
+      .with_for_update()
+      .first()
+    )
   if event is None:
     raise EventNotFoundError(f"Event {body.event_id} not found")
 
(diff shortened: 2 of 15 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Fourth path confirmed and fixed in `c4ecce74` — the review was right, and my audit note claiming no unaudited path remained was wrong.

`execute_event_block` reads the event unlocked and writes `pending`/`fulfilled` over it. One thing sits between that read and that write which the other three paths do not have: **a POST to QuickBooks.** So the lost update is not just a clobbered status — an event voided or superseded in that window gets published to QB anyway, then stamped `fulfilled` over the void. QB's `request_id` dedup covers a repeated post; it does not cover that.

One detail worth flagging back, because the suggested fix would have been subtly incomplete: the locked read needs **`populate_existing()`**. `close_service`'s QB pre-publish loop calls this on a *shared* session that has already loaded these events, so a bare `with_for_update()` query takes the lock and then keeps the in-session status from before it blocked — lock acquired, staleness intact. `update_event_block` got the same flag; there every caller opens a fresh session today, so it closes the footgun you noted rather than a live defect.

Both entry points are request-facing, so the wait is bounded and `execute-event-block` maps to 409. **No close-period handler**, deliberately: that pre-publish loop already collects per-event failures, so a lock conflict lands in the batch payload naming the event rather than failing the close — a handler there would be dead code.

Regression: `TestPublishLock` in `test_update_locking_db.py` holds the row from a second session and asserts `post_event_to_qb` is never called and the status is unchanged. Verified to fail with the lock removed, as with the other four guards.

All three minor observations taken: stub helper renamed with a corrected docstring (it does serve the Element and Agent loads), the on-demand promotion message no longer asserts a cause it cannot attribute, and the sweep's classified-id list now records that its accuracy rests on the candidate read's lock.

`just test-all`: 12,984 passed, 38 skipped, quality gate clean.

The earlier CodeQL failure was a GitHub API outage (`Initialize CodeQL` → `HttpError: No server is currently available`), which also hit #1198 on an unrelated branch and my own API calls. A fresh run is queued on this commit.

jfrench9:
Follow-up to the above: chasing your finding with a grep for the **writes** rather than the decisions — `event.status =`, not "transition" — turned up a fifth path. Fixed in `924e86a4`.

`ScheduleService.supersede_pending_obligations` reads a schedule's pending obligations and voids them when the entry template changes, while the promotion sweep reads the same rows to classify and draft their closing entries. Unlocked, both proceed from one snapshot: the sweep drafts a GL entry for an obligation this call is voiding, and the schedule ends up with **a closing entry for a period that no longer has an obligation.** Its read now takes `FOR UPDATE`; `update_schedule` bounds the wait (request-facing, contends with a background tick) and `update-information-block` maps the conflict to 409.

The grep returns eight `event.status =` sites. The three not locked, and why:

- **`journal_entry_recorded.py:424`** — covered by callers, not an independent writer. Runs only via `dispatch`, reached from `create_event_block` (which inserted the row, so it is self-locked) and `fire_handler_on_commit` (reached from `update_event_block`, the loader's auto-commit, and the sweep's autopilot — all of which lock first).
- **`rebuild_schedule` (`commands/schedules.py:757`)** — left unlocked deliberately. It voids the old originator event *unconditionally* as part of superseding it, so it decides nothing from the status it read. The only concurrent writer is another rebuild of the same schedule, and the loser overwriting a back-link is the intended outcome either way.
- The fifth is the one above.

**The method is the part worth recording.** My first audit asked "what else *transitions* an event?" and found the three paths that take a requested status and decide whether to accept it. `execute_event_block` derives its status from the outcome of an external call; `supersede_pending_obligations` from a template change. Neither reads as a "transition," so a search shaped by that word could not find either. Two completeness claims of mine were wrong before the audit got shaped by what a path *writes* — your review caught the fourth, the grep caught the fifth.

`just test-all`: 12,984 passed, 38 skipped, quality gate clean.

PR #1198, opened 2026-08-17
chore(orgs): collapse the graph-limit default to a single runtime source

jfrench9:
## Summary

`OrgLimits.max_graphs` had two default sources that could disagree. Both now resolve through the same runtime accessor. No behavior change on any existing path.

## Changes

**`models/core/org/org_limits.py`**

The column default was resolved once at import (`env.ORG_GRAPHS_DEFAULT_LIMIT`); `create_default_limits` reads the tuning value at runtime. Every current insert goes through the latter — `user_provisioning` and `scim_bootstrap` both call it, and `get_or_create_for_org` covers anything that reaches provisioning without a row — so the two never diverged in practice.

They could, though. An insert path that omitted `max_graphs` would take the import-time value, which falls back to the code constant rather than the deployed setting whenever the import-time lookup comes up empty. That is one insert path away from being the authoritative source, and the divergence would be silent.

Making the column default a callable over `TuningConfig.get_org_graphs_default_limit()` collapses both to one source, so a future insert inherits the deployed value by construction rather than by convention. The now-unused `env` import drops out.

Tests construct `OrgLimits` with an explicit `max_graphs` throughout, so none exercise the default; `just test-code` and the org / provisioning / billing suites are green.

### DIFF
--- a/robosystems/models/core/org/org_limits.py
+++ b/robosystems/models/core/org/org_limits.py
@@ -8,7 +8,6 @@ from sqlalchemy import Column, DateTime, ForeignKey, Integer, String
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.orm import Session, relationship
 
-from robosystems.config import env
 from robosystems.config.tuning import TuningConfig
 from robosystems.database import Model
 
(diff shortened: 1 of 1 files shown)


jfrench9:
Closing — the change isn't warranted. No production path reaches the column default (`create_default_limits` sets `max_graphs` explicitly and every caller routes through it), and the justification in the original description was wrong: `env.ORG_GRAPHS_DEFAULT_LIMIT` resolves through the same SSM path and the same fallback as the runtime accessor, so the two sources agree by construction rather than by luck.

The only genuine difference is that the module-level value is resolved once at import, so it can go stale against SSM until the next restart. That's real but reaches nothing, and it's already recorded as a known residual in the security notes. Not worth churning a model on.

PR #1200, opened 2026-08-17
fix(graph): delete published report artifacts at teardown

jfrench9:
## Summary

Deprovisioning removes the graph database, the extensions schema, documents, the search index, connections and connection credentials — and performs **no S3 deletion at all**. Published report bundles therefore outlive the tenant indefinitely.

This is not theoretical. Three of the five graph prefixes under `report-bundles/` in the production bucket belong to graphs the registry already lists as `deprovisioned`.

## Why the prefix has no lifecycle rule, and why that was still correct

`cloudformation/s3.yaml` deliberately leaves `report-bundles/` uncovered, and the comment there says not to "fix" it. That reasoning is sound: a time-based rule would delete the published artifact of a live report whose row still exists. Withdrawal is what removes them, via `delete_report_artifacts`.

What the earlier pass missed is that `delete_report_artifacts` has three callers and **none of them is teardown**. It checked that a delete path existed; it never asked whether deprovisioning could reach it. Uncovered was the right state for a clock, and the wrong state for teardown.

This is the fifth instance of the same defect class — after `Document` rows, the OpenSearch index, `Connection`, and `ConnectionCredentials`.

## Changes

`robosystems/operations/graph/deprovision_service.py`
- New `_purge_report_bundles` step, running as 3d immediately after the search-index purge — with the other data-store removals, before the registry deallocation releases capacity.
- Deletes the whole `report-bundles/{graph_id}/` prefix via the same `S3Client.list_objects` / `delete_object` pattern `delete_report_artifacts` uses, so every report, generation and flavor goes together. The final backup lives under `graph-backups/` and is untouched.
- `DeprovisionResult.report_bundles_deleted`, carried into the operator log line so the record states what was actually removed.

**One deliberate departure from its siblings.** The other best-effort steps record an error only when they raise. This one also records an error when an object *fails to delete*, degrading the teardown to `partial`. Incomplete disposal is precisely the outcome this step exists to prevent, so it should not pass quietly. A storage failure still never strands the teardown or blocks the capacity release.

## Why this matters beyond the code

Three approved, published documents describe this disposal, and until now each described a step that did not execute:

- Data Management Policy, Appendix B — `S3 - report bundles … deleted on withdrawal and at tenant decommissioning`
- Master Service Agreement §5 — deletion of Customer Data from active systems
- SOC 2 System Description §4 *Data* — teardown leaving "only the final backup and encrypted point-in-time database backups"

## Testing

- New `TestReportBundlePurge`, three cases: the prefix is scoped to the graph and every object under it is deleted with the count reported; an object that will not delete is recorded as an error and degrades status to `partial`; object storage being unreachable records a warning but still leaves the graph deprovisioned and soft-deleted.
- An autouse `stub_bundle_purge` fixture stubs `S3Client` for the whole module, matching the existing `_stub_search_purge` pattern — without it the new step would reach for a real bucket in every deprovision test.
- `uv run pytest tests/operations/graph` — 651 passed.
- `just test-code` — clean.

## Follow-up, not in this PR

The fix is forward-only. Objects belonging to graphs deprovisioned before it are not cleaned retroactively — nine objects across three prefixes today. Those are removed with a separate operator action against production.

### DIFF
--- a/robosystems/operations/graph/deprovision_service.py
+++ b/robosystems/operations/graph/deprovision_service.py
@@ -36,6 +36,7 @@ class DeprovisionResult:
   documents_deleted: int = 0
   connections_deleted: int = 0
   search_purged: bool = False
+  report_bundles_deleted: int = 0
   errors: list[str] = field(default_factory=list)
 
   @property
@@ -138,6 +139,9 @@ class GraphDeprovisionService:
     # --- 3c. Purge documents from the shared search index ---
     self._purge_search_index(graph_id, result)
 
+    # --- 3d. Purge published report artifacts from object storage ---
+    self._purge_report_bundles(graph_id, result)
+
     # --- 4. Deallocate DynamoDB registry ---
     await self._deallocate_registry(graph_id, result)
 
@@ -165,6 +169,7 @@ class GraphDeprovisionService:
         "documents_deleted": result.documents_deleted,
         "connections_deleted": result.connections_deleted,
         "search_purged": result.search_purged,
+        "report_bundles_deleted": result.report_bundles_deleted,
         "errors": result.errors,
       },
     )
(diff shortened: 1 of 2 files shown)


jfrench9:
@claude please review this PR

PR #1202, opened 2026-08-18
fix(events): order the batch locks and flush before the refreshed read

jfrench9:
## Summary

Three follow-ups to #1197, found by asking what that change made *worse* rather than what it fixed. One is a real defect it introduced, one is a latent footgun it introduced on the close path, and one is a test that could go red on a system that is working correctly.

## Changes

**Deadlock ordering — the defect.** The batch-locking reads added in #1197 had no `ORDER BY`, and two of them overlap: a pending `schedule_entry_due` obligation is matched by both the promotion sweep's predicate and `supersede_pending_obligations`'. Two transactions that take overlapping row locks in different orders deadlock, each holding what the other waits for, and the acquisition sequence was whatever each query plan happened to produce — not a property either query controlled, and not something a test would have noticed.

Postgres resolves a deadlock by killing one transaction with SQLSTATE `40P01`, which `bounded_lock_wait` did not translate. So it would have surfaced as an unhandled **500 on a close-adjacent path that had no locking at all before #1197** — strictly worse than the silent corruption trade #1197 was making.

All three batch reads now order by `id`, and `40P01` joins `55P03` as retryable. The shared key is named and justified in `locking.ORDERED_LOCK_KEY` — primary key, so the order is total and immutable under a concurrent status write. `40P01` stays reachable through the supersede pair in `update_event_block`, which locks two rows without a shared order, so translating it is not redundant.

**`populate_existing()` was discarding unflushed writes.** #1197 asserted this was safe on `close_service`'s shared session because autoflush would write the pending change before the re-read. The extensions factory is **`autoflush=False`** (`db/extensions.py:93`), so there is no autoflush, and the discarded value simply stops existing — no error, no log.

Nothing was lost in practice: `drafts_to_publish` is loaded and consumed with no mutation in between, so no caller has pending event writes at that point. But the close loop is the wrong place to depend on a caller's habits. Flushing before the locked read makes the re-read return our own pending write alongside anything another transaction committed — the behavior the original comment claimed.

**The concurrency test's handoff was a race.** It used a fixed `sleep(0.2)` to let the winner take the lock first. On a loaded runner the winner may not have acquired by the time it elapses; the loser locks first, its transition *succeeds*, and the assertion fails on a system that is working. The winner now signals after acquiring, and the loser releases it on a timer.

## Breaking Changes

None. No request shape, response body, GraphQL type, or envelope field changes. `40P01` maps to the same 409 as `55P03` on the operations that already had it, so no new status codes.

## Testing

`just test-all` — **12,995 passed, 38 skipped**; ruff, format, typecheck, cf-lint clean.

New guards verified to fail with the guard removed:

| Guard removed | Result |
| --- | --- |
| `order_by(Event.id)` on the sweep's candidate load | ordering pin fails |
| `session.flush()` before the locked read | `test_the_command_keeps_a_pending_write` fails |

Both halves of the flush behavior are pinned: one test shows `populate_existing` alone dropping the edit (documenting *why* the flush is there, since it looks removable without it), another shows the command keeping it.

The deadlock itself is not directly testable once the ordering is in place — constructing it requires the opposing acquisition order the fix eliminates — so it is pinned at the statement, the same way #1197 pinned the classify UPDATE's status predicate.

## Certification

### DIFF
--- a/robosystems/operations/event_block/commands.py
+++ b/robosystems/operations/event_block/commands.py
@@ -378,6 +378,7 @@ def update_event_block(
   #
   # Bounded, because the conflicting writer is usually a sync or a promotion
   # sweep that runs long — see `locking.bounded_lock_wait`.
+  session.flush()  # pairs with populate_existing below; autoflush is off here
   with bounded_lock_wait(
     session,
     f"Event {body.event_id} is being written by another process "
@@ -387,7 +388,7 @@ def update_event_block(
     # production caller opens a fresh session per request, so the identity map
     # is empty today, but a caller that reused a session would otherwise get
     # the lock and a stale status — the one combination this guard exists to
-    # prevent.
+    # prevent. The flush that pairs with it is there too, one line up.
     event = session.get(
       Event, body.event_id, with_for_update=True, populate_existing=True
     )
@@ -684,6 +685,16 @@ def execute_event_block(
   # QB pre-publish loop calls this on a **shared** session that already loaded
   # these events, so without it the lock would be taken while `event.status`
   # kept the value it held before blocking.
+  #
+  # The flush pairs with it and is not optional. This factory is
+  # `autoflush=False` (`db/extensions.py`), so a re-read would otherwise
+  # overwrite any in-flight change to this row that the caller had not yet
+  # written — silently, since the discarded value simply stops existing. Flush
+  # first and the re-read returns our own pending write along with anything
+  # another transaction committed. No caller has pending event writes here
+  # today; this makes that a property of the function rather than of its
+  # callers.
+  session.flush()
   with bounded_lock_wait(
     session,
     f"Event {body.event_id} is being written by another process. Retry in a moment.",
--- a/robosystems/operations/event_block/commands.py
+++ b/robosystems/operations/event_block/commands.py
@@ -43,7 +43,7 @@ from robosystems.operations.roboledger.reads.event_block import (
 )
 
 from .engine import EngineValidationError, apply_handler
-from .locking import bounded_lock_wait
+from .locking import bounded_lock_wait, ordered_lock_column
 from .python_handlers import get_python_handler
 from .python_handlers.types import HandlerMetadataValidationError
 from .registry import (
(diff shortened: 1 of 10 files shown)


jfrench9:
@claude please review this PR

jfrench9:
All four findings addressed in `200b5c25`. The main one was correct and I verified it independently — `bounded_lock_wait` only sees errors raised inside its block, the successor was fetched unlocked afterwards, and both UPDATEs issued at `session.commit()` outside the wrapper.

**Eliminated rather than translated.** The locked read now covers every row the operation will write, in one ordered statement keyed on `id` — the same discipline the batch locks use. Two opposing supersedes acquire in the same order, so the cycle cannot form. Error precedence is preserved (`superseded_by_id is None` → self-reference → terminal-state → not-found), which `test_supersede_missing_successor_raises` and `test_supersede_from_terminal_state_rejected` both depend on.

**One extension worth flagging back: the gap is structural, not local to the supersede pair.** `session.commit()` is outside the wrapper in `promote_obligations` (`schedules.py:279-291`) and `update_schedule` as well, and `execute_event_block` doesn't own its commit at all — `extensions_session` does. A deadlock materializes when writes *flush*, so translation at lock acquisition structurally cannot cover those paths. `locking.py` now states that limit instead of implying the opposite; covering them properly means translating at the transaction boundary, which is its own change rather than a patch here.

Also taken:

- **40P01 had no test** — the one guard on this branch that didn't meet the standard #1197 set. It has one now, and it fails when the state is dropped from `_RETRYABLE_LOCK_STATES`. You're right that it only proves the acquisition case; that's the honest limit of what a wrapper at that position can be tested for.
- **`ORDERED_LOCK_KEY` was read by nothing** — three call sites hardcoded `Event.id` next to a comment pointing at a constant that couldn't keep them in step. It's now a shared column the call sites actually `order_by`.
- **Ordering pins compare by identity**, not repr, so a SQLAlchemy repr change won't fail them.

Both new guards verified to fail with the guard removed, same as the rest of this branch:

| Guard removed | Result |
| --- | --- |
| successor from the locked batch → unlocked `session.get` | `test_supersede_locks_both_rows_in_id_order` fails |
| `40P01` dropped from `_RETRYABLE_LOCK_STATES` | `test_deadlock_also_becomes_event_locked_error` fails |

`just test-all`: 12,997 passed, 38 skipped, quality gate clean.

PR #1203, opened 2026-08-18
fix(roboledger): lock the remaining state transitions

jfrench9:
## Summary

Sweeping the *writes* rather than the concept — `grep` for assignments to `status`, `closed_at`, `filed_at` and friends across `operations/` — turned up three more instances of the shape #1197 and #1202 have been closing: load a row, branch on a state column, write it back, with nothing held across the decision.

Follows #1202 (merged), whose locking primitives this builds on.

## Changes

**`reverse_journal_entry` double-posted.** `commands/journal_entries.py` read the original entry unlocked, checked `status == 'posted'`, then wrote a full reversing entry with flipped line items. Two concurrent reversals both passed the check and both posted — **the entry got reversed twice.** Each reversing entry is internally balanced, so the trial balance stayed happy while the books sat one entry's worth off in the other direction. Nothing at the database level stopped it: no unique index on `reversal_of`, and the reversing entry carries no `idempotency_key`.

The command now locks the original. **Migration 0032** adds a partial unique index on `reversal_of`, because "an entry is reversed at most once" is a real invariant rather than a concurrency workaround — it should hold against a double-clicked button and against a future caller that forgets to lock, not only against the race.

⚠️ **The migration must be checked against existing data before deploy.** It will fail if any original already has two reversals, and that failure is the correct outcome: it means genuinely double-reversed entries are in the ledger and need an accounting decision, not a schema decision. The query to run per tenant schema is in the migration's docstring.

**`close_period` / `reopen_period`.** Both read the `FiscalPeriod` row unlocked and branched on its status. Two concurrent closes would each run the entire close — promoting drafts, **publishing to QuickBooks**, stamping statements. Parts of that are accidentally idempotent (QB's `request_id` dedup, replace-semantics on stamping); none of it is a guard. A reopen interleaved with a close leaves a period marked closed whose statements were retracted. Both now lock the same row.

**`file_report`.** The mild one — a concurrent double-file overwrites the audit stamp rather than duplicating anything. But `filed_by` and `filed_at` are exactly the fields an auditor reads, so last-writer-wins isn't good enough for them.

**The structural change, which is the point of this PR as much as the three fixes:** the lock policy moves from `operations/event_block/locking.py` to `operations/locking.py`, and `EventLockedError` becomes `RowLockedError`. Filed inside one consumer, the discipline was invisible to every other subsystem with the same shape — which is why instances two through six were each found one at a time by going to look. `lock_by_id` is the one-call version of the correct read (FOR UPDATE, `populate_existing`, a flush first because these sessions are `autoflush=False`, and a bounded wait), so a caller no longer has to remember four separate things to get it right.

## Breaking Changes

None. `RowLockedError` maps to the same 409 as before on the operations that already had it, plus `update-information-block` and `file-report`. Internal rename — no SDK symbol, no request or response shape change.

## Testing

`just test-all` — **13,001 passed, 38 skipped**; quality gate clean. Migration applied locally against the extensions DB.

New DB-backed regressions in `test_state_transition_locks_db.py` (real concurrent sessions, throwaway tenant schema). Each guard verified by removing it:

| Guard removed | Result |
| --- | --- |
| reversal's `lock_by_id` | `test_a_locked_entry_cannot_be_reversed` fails |
| `uq_entries_one_reversal_per_original` from the model | fixture assertion fails (model/migration drift check) |
| reopen's `with_for_update()` | `test_a_locked_period_cannot_be_reopened` fails |

One verification is worth reporting precisely: removing the reopen guard's **wrapper** rather than its `FOR UPDATE` makes the suite hang rather than fail, because dropping `bounded_lock_wait` also drops the `lock_timeout`. That's the bounded wait demonstrating its own purpose, and it's why the isolation test removes the lock while keeping the wrapper.

The reversal tests also pin that the partial index leaves ordinary entries (`reversal_of IS NULL`) unconstrained.

## Certification

### DIFF
--- a/migrations/extensions/versions/0032_one_reversal_per_entry.py
+++ b/migrations/extensions/versions/0032_one_reversal_per_entry.py
@@ -1,0 +1,73 @@
+"""an entry is reversed at most once
+
+``reverse_journal_entry`` read the original entry unlocked, checked that its
+status was ``posted``, and then wrote a full reversing entry plus flipped line
+items. Two concurrent reversals of the same entry both passed that check and
+both posted — the entry got reversed twice. Each reversing entry is internally
+balanced, so nothing failed to foot and no validation complained; the books
+simply sat one entry's worth off in the other direction.
+
+The command now locks the original, which turns the ordinary race into a clean
+409. This index is the half that does not depend on the code path: two
+reversing entries against one original is wrong however it arose — a
+double-clicked button, a retried request, a future caller that forgets to
+lock — and the database is the only place that can say so unconditionally.
+
+Partial, because ``reversal_of`` is NULL on every entry that is not a reversal
+and those must stay unconstrained.
+
+**Before deploying**: this will fail if any original already has two reversals.
+That failure is the correct outcome — it means real double-reversed entries are
+sitting in the ledger and need an accounting decision, not a schema decision.
+Check first:
+
+    SELECT reversal_of, COUNT(*) FROM entries
+    WHERE reversal_of IS NOT NULL GROUP BY reversal_of HAVING COUNT(*) > 1;
+
+per tenant schema, and resolve any rows it returns before running this.
+
+Revision ID: 0032
+Revises: 0031
+Create Date: 2026-08-17
+
+"""
+
+from __future__ import annotations
+
+from alembic import op
+from sqlalchemy import text
+from sqlalchemy.engine import Connection
+
+from migrations.extensions.helpers import for_each_tenant_schema
+
+revision = "0032"
+down_revision = "0031"
+branch_labels = None
+depends_on = None
+
+_INDEX = "uq_entries_one_reversal_per_original"
+
+
+def _create(conn: Connection, schema: str) -> None:
+  conn.execute(
+    text(
+      f'CREATE UNIQUE INDEX IF NOT EXISTS "{_INDEX}" '
+      f'ON "{schema}".entries (reversal_of) WHERE reversal_of IS NOT NULL'
+    )
+  )
+
+
+def _drop(conn: Connection, schema: str) -> None:
+  conn.execute(text(f'DROP INDEX IF EXISTS "{schema}"."{_INDEX}"'))
+
+
+def upgrade() -> None:
+  conn = op.get_bind()
+  _create(conn, "public")
+  for_each_tenant_schema(conn, _create)
+
+
+def downgrade() -> None:
+  conn = op.get_bind()
+  _drop(conn, "public")
+  for_each_tenant_schema(conn, _drop)
--- a/robosystems/db/extensions.py
+++ b/robosystems/db/extensions.py
@@ -87,6 +87,19 @@ def _get_engine():
   return _engine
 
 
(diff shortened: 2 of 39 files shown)


jfrench9:
@claude please review this PR

jfrench9:
All six confirmed and addressed in `ce55ebf3`. Two of them were defects this branch introduced rather than gaps it left, which is the part worth saying plainly.

**P1b — the report lock was in the wrong function.** My single-occurrence replace matched `regenerate_report`. Fixed. `regenerate_report` turns out to want a lock of its own — it gates on `filing_status` being filed/archived and then rewrites the report's facts, so a file landing in between produces a report stamped `filed` whose contents were replaced underneath it. Both are locked now, each with a comment describing what it actually protects.

**P1a — the close lock is removed, not repaired.** You're right that a transaction-scoped lock can't span `_publish_drafts_to_qb`'s commit: after it, everything before is unprotected; before it, the commit releases it. And my comment claimed the lock covered the QB pre-publish, which runs 14 lines earlier. A lock that misdescribes its own coverage is worse than no lock, because the next reader stops looking. The gap is now documented at the code and in the spec, with the primitive named — `pg_advisory_lock`, session-scoped, *not* `pg_advisory_xact_lock`.

Reopen keeps its lock (it does hold across its decision and write) with the comment corrected: it serializes concurrent reopens, not reopen against close.

**P2a — the reachable 500.** Sharper than "unhandled": the original's status is untouched by auto-reverse, so the `posted` check passes on an entry that already has a reversal. Before the index that silently produced a second one; after it, an opaque `IntegrityError`. Now `JournalEntryAlreadyReversedError`, checked under the lock, mapped to 422, with the index as the guarantee behind it rather than the thing that reports the problem.

**P1c — 409 coverage.** Added to `create-event-block` (how reversal is actually invoked, via the `journal_entry_reversed` handler) and to the hand-written reopen and regenerate handlers.

**P2b — alias kept.** `EventLockedError` still resolves. The module promises a stable import path; breaking it on the release that moved the module isn't worth the line it saves.

**P1d — not fixed here, deliberately.** `assert_period_not_closed` is an unlocked SELECT, so a writer can still land a draft inside a closing period. That needs the same session-scoped lock the close does, acquired by every ledger writer — one change, not two. Recorded in `specs/ledger/event-write-isolation.md` §5, whose earlier remedy ("row lock inside the close transaction") is now marked wrong for the reason you gave.

New coverage for what shipped: DB-backed `file_report` lock test, and an already-reversed test that seeds a schedule-style auto-reversal and asserts the domain error fires with exactly one reversal remaining. The close/reopen test class now states in its docstring that it does **not** cover reopen-vs-close, rather than implying it.

`just test-all`: 13,003 passed, 38 skipped, quality gate clean.

On your last point — the suite missing the decisive scenarios is the accurate diagnosis of how all six got through. Every one of these was green on 13,001 tests. The three that remain untestable without new machinery (concurrent close, entry-writes-racing-close) are exactly the ones I've left open rather than half-closed.

jfrench9:
@claude can you review

PR #1204, opened 2026-08-18
fix(close): isolate each QuickBooks publish in a savepoint

jfrench9:
## Summary

Regression from `v1.9.4`, currently deployed to production.

`execute_event_block` gained a bounded lock wait in #1202, so a conflicting writer now surfaces there as an error over an **aborted** PostgreSQL transaction rather than blocking. The close's QuickBooks pre-publish loop catches broadly and continues, which meant every later statement failed on that aborted transaction and the `session.commit()` at the end of the loop failed with them — the commit whose entire purpose is to make the `qb_external_id` markers durable.

That produces exactly the failure the commit exists to prevent: **QuickBooks holds journal entries the ledger has no record of sending, and the retried close publishes them a second time** once QB's RequestId dedup window expires. Duplicate entries in a customer's books.

## Changes

`_publish_drafts_to_qb` opens a `SAVEPOINT` per entry, so a failure is contained to the entry that failed and the markers for entries that already reached QuickBooks still commit. Same pattern the loader's auto-commit pass already uses for the same reason.

Nothing else changes — no new locks, no error-mapping changes, no schema.

## Breaking Changes

None.

## Testing

`tests/operations/roboledger/fiscal_calendar/test_publish_savepoint.py` drives the real `_publish_drafts_to_qb` with a database error on the middle of three entries and asserts both halves of the property: the other two published, and their markers committed *before* `WritebackFailed` was raised. It models `begin_nested()` the way PostgreSQL behaves — the savepoint absorbs the failure and the outer transaction stays usable — because that behaviour is the thing under test rather than an incidental detail.

**Verified to fail without the fix**: removing the savepoint leaves `session.journal == ['commit']` instead of the begin/release/begin/rollback/begin/release/commit sequence.

`tests/operations/roboledger` + `tests/operations/event_block`: 993 passed. `just test-code`: clean.

Full-gate note: three consecutive `just test-all` runs on the parent branch failed on three *different* tests, two of them in areas untouched by any of this work, each passing in isolation and in its own module. That flakiness is pre-existing and worth its own attention — `pytest.ini` sets `-x`, so a flaky early failure masks everything after it, which is how a real break in the parent branch hid for one round.

## Deployment

This should go out ahead of the remaining locking work. The exposure needs a concurrent writer holding an event row during a close — a sync finishing as a close begins is enough, and the close gate's staleness check encourages that ordering.

## Certification

### DIFF
--- a/robosystems/operations/roboledger/fiscal_calendar/close_service.py
+++ b/robosystems/operations/roboledger/fiscal_calendar/close_service.py
@@ -525,14 +525,29 @@ class PeriodCloseService:
     failed_events: list[dict] = []
     for entry, event in drafts_to_publish:
       try:
-        result = execute_event_block(
-          session,
-          ExecuteEventBlockRequest(
-            event_id=str(event.id),
-            connection_id=qb_connection_id,
-          ),
-          created_by=actor_id,
-        )
+        # Each publish runs in its own SAVEPOINT so a database error here
+        # cannot take the whole batch down with it.
+        #
+        # `execute_event_block` bounds its wait for the event's row lock, so a
+        # conflicting writer surfaces as an error over an **aborted**
+        # transaction rather than as a block. Without a savepoint the loop
+        # would carry on collecting failures while every later statement failed
+        # on that aborted transaction, and the `session.commit()` below — the
+        # one whose entire purpose is to make the qb_external_id markers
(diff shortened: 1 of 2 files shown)


jfrench9:
@claude please review this PR

PR #1205, opened 2026-08-18
fix(roboledger): refuse retracted events and paginate teardown purge

jfrench9:
## Summary

Follow-up to the v1.9.4–v1.9.5 lock series. `execute-event-block` still posted a retracted event to QuickBooks and stamped it `fulfilled`, and close still selected those leftover drafts — including posting them locally. Teardown also treated a failed or truncated S3 listing as an empty prefix, so published report artifacts could survive deprovision.

## Changes

- **Publish** (`execute_event_block`): after the locked read, refuse `voided` / `superseded` with `EventNotPublishableError` (409) before any QuickBooks write. An existing `qb_external_id` or `fulfilled` status is an idempotent skip. Execute promotes the event's drafts to `posted`, so it now also takes the shared period fence before the row lock and returns `ClosedPeriodError` (422) when the event's posting date is in a closed period — that check runs ahead of the idempotent skips and the native fast-path, so a re-execute in a closed period is a 422 rather than a no-op. Close already holds the exclusive period fence on a dedicated connection, so it calls execute with `acquire_period_fence=False`.
- **Close writeback**: `select_writeback_eligible_entries` excludes retracted event statuses (shared with the outbox preview). The close bulk draft→posted pass uses the same exclusion so leftover drafts are neither published nor posted locally.
- **Lock order**: inbox approval (`transition_to='committed'`) and autopilot promotion take the shared period fence *before* the event row lock, matching journal update/delete and close. Fixes the inversion that could fail close mid-publish after a 3s `lock_timeout`. Approval fences both the event's current posting date and the one a same-call `effective_at` patch moves it to (as `update_journal_entry` does). Autopilot fences only its **write set** — pending obligations plus stranded classified ones — not every classified obligation: dispatched obligations rest at `classified` for good, so fencing the whole candidate set would fence every period that ever held a schedule and fail the sweep on the first one since closed (a tenant's second month-end). An obligation whose period is closed is left out of the sweep and reported in `errors`, not fatal to it.
- **Teardown / withdrawal**: new `S3Client.iter_object_keys` paginates and raises on list failure. Graph deprovision and `delete_report_artifacts` use it instead of the single-page, empty-on-error `list_objects`.
- **Comment**: the stale `ORDERED_LOCK_KEY` reference now points at `ordered_lock_column()`.

Reviewers: the close-vs-execute fence skip (`acquire_period_fence=False`) is load-bearing — taking the shared fence inside close would wait on close's own exclusive lock.

## Breaking Changes

None. `EventNotPublishableError` is a new 409 on `execute-event-block` for a case that previously succeeded incorrectly, and `execute-event-block` now returns 422 (`ClosedPeriodError`) for an event whose posting date is in a closed period (additive; SDK regen opportunity).

## Testing

Targeted suite against local Postgres (not `just test-all`):

- `tests/operations/event_block/test_execute.py`
- `tests/operations/event_block/test_commands_update.py`
- `tests/operations/event_block/test_promotion.py`
- `tests/operations/event_block/test_promotion_fence_db.py` (new — real Postgres: closed history does not block the open period; stranded/pending obligations in a closed period are reported, not fatal)
- `tests/operations/event_block/test_update_locking_db.py`
- `tests/operations/event_block/test_engine.py`
- `tests/operations/roboledger/fiscal_calendar/test_qb_writeback.py`
- `tests/operations/roboledger/fiscal_calendar/test_close_service.py`
- `tests/operations/roboledger/fiscal_calendar/test_publish_savepoint.py`
- `tests/operations/roboledger/fiscal_calendar/test_close_publish_durability.py`
- `tests/operations/roboledger/commands/test_journal_entries.py`
- `tests/operations/roboledger/commands/test_fiscal_calendar.py`
- `tests/operations/roboledger/commands/test_guards.py`
- `tests/operations/roboledger/commands/test_share_controls_db.py`
- `tests/operations/graph/test_deprovision_service.py`
- `tests/operations/aws/test_s3_client.py`
- `tests/operations/test_locking.py`

309 passed. `ruff check` / `ruff format` and `basedpyright` on the changed source files: clean.

## Certification

### DIFF
--- a/robosystems/operations/aws/s3.py
+++ b/robosystems/operations/aws/s3.py
@@ -6,6 +6,7 @@ import asyncio
 import gzip
 import hashlib
 import json
+from collections.abc import Iterator
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import asdict, dataclass
 from datetime import UTC, datetime
@@ -23,8 +24,9 @@ from robosystems.logger import logger
 class S3Client:
   """Sync S3 client for general object storage (XBRL textblocks, temp files).
 
-  All methods swallow errors and return a falsy value rather than raising,
-  so callers can translate to HTTP without try/except.
+  Most methods swallow errors and return a falsy value rather than raising,
+  so callers can translate to HTTP without try/except. ``iter_object_keys``
+  is the exception: a failed or truncated listing must not look empty.
   """
 
   def __init__(self, region_name: str | None = None, endpoint_url: str | None = None):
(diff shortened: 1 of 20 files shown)


jfrench9:
@claude please review this PR

PR #1209, opened 2026-08-18
fix: harden the extensions write path — staleness, provisioning, error surfacing, curation guards

jfrench9:
## Summary

Follow-on to #1205–#1208: the rest of the OLTP mutation path (REST operations, MCP tools, operators, background jobs) reviewed for writes that could land wrong, be lost, or hide a database failure. This PR is the pressing set; the remainder is queued as follow-up patches (design notes: `specs/extensions/oltp-write-path-hardening.md`, `specs/ledger/ledger-write-guards.md`).

### Extensions platform (`5444084e`)
- **`Graph.mark_fresh` compares before it clears.** A write that lands during a materialization stamps a later `graph_stale_at`; clearing unconditionally dropped it from the graph until an unrelated later write. The compare runs in SQL (immune to a stale identity map); all four materializers pass their snapshot time.
- **Tenant provisioning is a first-time act.** The OLTP loader ran the full CHECK/TRIGGER DDL under AccessExclusive locks on every sync; it now provisions only a missing schema (`ensure_tenant_schema`), and provisioning refuses a deprovisioned graph instead of resurrecting its schema.
- **Database faults surface.** The registrar translated every `ProgrammingError` and unmapped `ValueError` to "not initialized" 404 with only that text audited. Now only a missing schema/relation is 404; other programming errors are logged and surface; an unmapped `ValueError` is the 422 it is. `set-close-target` answers 409 on a held calendar row; `update-event-block` maps what its handler can raise on approve, as `create-event-block` does.
- **MCP registrar tools** return fixed messages for unmapped errors (detail logged server-side) rather than DBAPI strings.
- **Mapping operator** counts a mapping once the write lands; a rejected/failed persist is skipped, not reported as mapped.

### Ledger (`85559ecc`)
- **Savepoints in the ledger's loops.** The promotion sweep dispatches each obligation under `begin_nested()` and re-raises lock waits — a DB-level failure in one handler no longer aborts the transaction (one poison obligation blocked every other one each tick, and the commit failed with "transaction is aborted" instead of `RowLockedError`). The close's rule evaluations get the same treatment.
- **Curation never reaches immutable facts.** `delete-taxonomy-block cascade_facts=true` and `update-taxonomy-block structures_to_remove` refuse before the first delete when a filed report's snapshot or a closed month's canonical statement sets would go; the validator reports the same condition.
- **The ledger's own entity is resolved by predicate** (parent, not `linked`) at the four sites that used `LIMIT 1` — the QuickBooks CompanyInfo overwrite, the CoA auto-link, `link-entity-taxonomy` — so a shared-in counterparty can never receive them.
- **A chart of accounts is deletable again** (the cascade sweeps the entity's adoption row — RESTRICT FK); element removal through update sweeps the rules/classifications that FK the removed atoms.
- **`update-taxonomy-block` locks the taxonomy row** (409 on contention), so concurrent envelope updates cannot duplicate auto rules or race the qname index.

## Behavior notes
- Registrar ops: unmapped `ProgrammingError` → 500 (was 404), unmapped `ValueError` → 422 with message (was 404). MCP registrar tools: `not_initialized` / `invalid_request` / fixed `command_failed` (was raw `str(exc)`).
- `delete-taxonomy-block` / `update-taxonomy-block`: new 422 refusal (`protected_facts`) when the cascade would reach a filed report or closed period.

## Test plan
- New real-Postgres tests: `mark_fresh` compare-and-clear (incl. stale identity map), sweep savepoint (poison obligation + lock propagation), immutability refusals (filed/draft, closed/open, validator + apply), CoA-with-adoption delete, element removal with targeting rules, parent-entity resolver under heap-order mis-pick; MCP registrar error surfacing; provisioning ensure/refuse; mapping operator counts. Load-bearing new tests verified to fail on `main`.
- 3,444 tests across the touched suites pass; `just test-code` clean.

### DIFF
--- a/robosystems/dagster/jobs/extensions.py
+++ b/robosystems/dagster/jobs/extensions.py
@@ -11,6 +11,7 @@ Can be triggered:
 - On a schedule (future)
 """
 
+from datetime import UTC, datetime
 from typing import Any
 
 from dagster import (
@@ -105,6 +106,10 @@ def materialize_extensions_to_graph(
         },
       )
 
+    # Stamped before the source is read: a write that marks the graph stale
+    # after this point is not in the snapshot, and mark_fresh must leave the
+    # flag set for it (compare-and-clear).
+    started_at = datetime.now(UTC)
     materializer = ExtensionsMaterializer()
     result = loop.run_until_complete(
       materializer.materialize(
@@ -140,7 +145,12 @@ def materialize_extensions_to_graph(
   try:
     graph_row = db_session.query(Graph).filter(Graph.graph_id == graph_id).first()
     if graph_row:
-      graph_row.mark_fresh(session=db_session)
+      cleared = graph_row.mark_fresh(session=db_session, started_at=started_at)
+      if not cleared:
+        context.log.info(
+          f"{graph_id} was written during the materialization; leaving it stale "
+          "for the next sweep"
+        )
   finally:
     try:
       next(db_gen)
--- a/robosystems/dagster/jobs/graph.py
+++ b/robosystems/dagster/jobs/graph.py
@@ -817,6 +817,9 @@ def materialize_graph_tables(
   )
 
   start_time = time.time()
+  # Compare-and-clear anchor for mark_fresh: a write stamped after this
+  # point is not in this build and must keep the graph stale.
+  started_at = datetime.now(UTC)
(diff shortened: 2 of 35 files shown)


jfrench9:
@claude can you review

PR #1210, opened 2026-08-18
fix(extensions): OLTP write-path hardening — tiers B and C

jfrench9:
## Summary

Follow-on to #1209: the queued tiers B and C of the same OLTP write-path review, folded into one PR so the whole set rides the next patch release together (design notes: `specs/extensions/oltp-write-path-hardening.md` §4, `specs/ledger/ledger-write-guards.md` §4 — the two rows that need a design decision or a dedupe migration stay open there).

### Runners off the event loop; one error policy (`c06b9f79`)
- **Every REST operation runner, sync `on_fresh_success` hook, and MCP tool body now runs in a worker thread** (`run_off_loop`, limiter sized to the extensions pool). The API runs one uvicorn worker; sync database and network work executed inline on the loop, so a close posting to QuickBooks or a bounded lock wait froze every other request — including the ALB health check.
- **The 22 hand-written roboledger handlers share the registrar's error policy** via `guard_command_runner` (schema-missing → 404, other database faults logged and surfaced, unmapped `ValueError` → 422 with its message) instead of each mapping every `ProgrammingError`/`ValueError` to the not-initialized 404. Hand-written MCP tools answer database faults with a fixed message rather than echoing the driver string.
- **Subgraph ids are refused up front (403)** by the extension dependency, matching the GraphQL endpoint, instead of failing inside a handler as 404 or 500 depending on which one caught it. Extensions engine singleton built under a lock.

### Idempotency in flight; attribution (`c9238c3b`)
- **Idempotency keys are reserved before the run** (`SET NX` pending marker). A duplicate that lands while the first request is still executing gets 409 ("still in progress") instead of executing alongside it; a failed run releases the claim so the retry executes. MCP calls, which carry no key, refuse an identical concurrent call the same way.
- **MCP tool calls emit the same audit line as their REST twin** (`surface=mcp`); MCP write-role denials emit the `AuthorizationDenied` audit; writes made through operators and in-process tools are stamped with the acting user (`DirectToolAccess`/`HttpToolAccess` carry `user_id`) rather than a fixed tag, the graph's first member, or `mcp:{graph_id}`.

### Ledger ownership, share consistency, teardown (`94dc91a5`)
- **Report lifecycle writes share one actor rule**: the author, never a copy shared in from another graph — file and status transitions were unchecked, and regenerate accepted a shared-in copy from its sender. Publish-list membership is the list owner's to change; a recipient only needs an extensions tenant (the rule `share-report` already applied).
- **`share-report` keeps the source report row-locked from snapshot to share rows in one transaction**, so a concurrent delete/regenerate waits (409 on the bounded wait) instead of leaving recipients with copies the sender can no longer revoke; a recipient's block that lands mid-copy still wins.
- **Backfill re-stamps a closed month as one fenced transaction** (reopen + re-close), so a failing re-close leaves the month closed rather than open with its statements retracted.
- **Deprovisioning** stamps `deleted_at` before it drops anything (the schema provisioner refuses such a graph, closing the in-flight-sync window), re-runs the idempotent disposal steps on an already-deprovisioned graph (admin returns 200 with what was still there), and gains an orphan-schema sweep (`GET/POST /admin/v1/graphs/orphan-schemas`, `admin graphs orphan-schemas [--drop]`). The extensions alembic env commits per migration under a `lock_timeout` with retry.

### Tier-C hygiene (`287dea3f`)
- Unique-key races surface as the typed conflict the pre-check raises (calendar init, block, publish-list members, mapping association, entity-taxonomy link, CoA auto-link, scenario dimension, event external id); JSON read-modify-write updates lock the row; `create_element` under a savepoint; portfolio dispose-before-add.
- Verification results swept at every report fact-set deletion; a deleted report's stored artifacts removed after commit; S3 client failure after commit no longer 500s.
- Engine `idle_in_transaction_session_timeout` (5 min; no `statement_timeout`), fence `RESET lock_timeout`, operator adapters enforce `graph_scope`, model CHECK vocabularies single-sourced with the provisioning widener (guard test), `qb_response_at` tz-aware.
- **Migration 0033 (extensions)**: `idx_events_qb_external_id` for tenants provisioned by `create_all` after 0014 (skips tenants that hold the 0014-named index); the index is on the `Event` model now.

## Behavior notes
- Hand-written roboledger ops: unmapped `ProgrammingError` → 500 (was 404), unmapped `ValueError` → 422 with message (was 404); subgraph ids → 403 (was 404/500).
- `Idempotency-Key` reuse while the first request runs → 409 (was a second execution). MCP: `{"error": "in_progress"}` for an identical concurrent call.
- `file-report` / `transition-filing-status` / `regenerate-report`: 403 for a non-author or a shared-in copy. `add-publish-list-members` / `remove-publish-list-member`: 403 for a non-owner; targets may be roboinvestor tenants. `share-report`: 409 on a held report.
- Admin `POST /admin/v1/graphs/{id}/deprovision` on an already-deprovisioned graph: 200 with `status=already_deprovisioned` (was 409).
- Extensions DB: run `just migrate-up extensions` (0033) with the patch.

## Test plan
- New tests: off-loop runner/hook/contextvars; idempotency reservation (concurrent same-key 409, failed run releases, hook failure releases, cache primitives); MCP registrar worker thread, audit line, in-flight refusal; hand-written tool DB-fault messages; subgraph guard; report ownership (regenerate/file/transition, shared-in copies); publish-list ownership; real-Postgres `share_report` (share rows in the same transaction; a delete during the copy step gets `RowLockedError`); restamp-as-one-transaction; deprovision re-run + orphan sweep + provisioner refusing an in-flight teardown; engine idle timeout; race translations; model-CHECK/widener parity; operator graph-scope gate; created_by attribution.
- Full unit suite: 13,146 passed. `just test-code` clean. Local stack: restarted on the branch, provisioned a skeleton roboledger graph, ran `initialize` / `create-agent` (replay + conflict) / subgraph guard end-to-end; 0033 applied and rolled back locally.

### DIFF
--- a/migrations/extensions/env.py
+++ b/migrations/extensions/env.py
@@ -7,10 +7,12 @@ This is separate from the platform alembic/env.py because:
   tenant schemas created at runtime by schema_provisioner)
 """
 
+import sys
+import time
 from logging.config import fileConfig
 
 from alembic import context
-from sqlalchemy import engine_from_config, pool
+from sqlalchemy import engine_from_config, pool, text
 
 from robosystems.db.extensions import ExtensionsBase, get_extensions_database_url
 
@@ -48,6 +50,31 @@ def run_migrations_offline() -> None:
     context.run_migrations()
 
 
+# Extensions migrations fan out over every tenant schema, and they run at
+# daemon boot while the previous release's API tasks are still serving those
+# tenants. Two things follow:
+#
+# - `transaction_per_migration`: each migration commits on its own, so a
+#   fan-out that fails on tenant N leaves the migrations before it applied
+#   (and `alembic_version` advanced) instead of rolling back an hour of DDL,
+#   and the retry starts from the failed migration rather than from scratch.
+# - `lock_timeout`: a per-tenant `ALTER TABLE` takes an ACCESS EXCLUSIVE lock;
+#   behind one long-running reader it would otherwise wait forever, holding
+#   the locks it already took on every earlier tenant and stalling their
+#   requests. With a bound the migration fails fast on that tenant, releases
+#   everything, and is retried after a pause. Statements can raise their own
+#   `SET LOCAL lock_timeout` when they know better.
+MIGRATION_LOCK_TIMEOUT = "30s"
+MIGRATION_LOCK_RETRIES = 5
+MIGRATION_LOCK_RETRY_PAUSE_SECONDS = 10.0
+
+
+def _is_lock_timeout(exc: BaseException) -> bool:
+  """psycopg2 raises `lock_not_available` (55P03) for a lock_timeout."""
+  orig = getattr(exc, "orig", None)
+  return getattr(orig, "pgcode", None) == "55P03"
+
+
 def run_migrations_online() -> None:
   """Run migrations in 'online' mode."""
   connectable = engine_from_config(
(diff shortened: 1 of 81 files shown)


jfrench9:
@claude can you review?

jfrench9:
Review follow-ups landed in `f924a74e`:

1. **Reservation vs. cancellation** — fixed, and one step further than release-on-cancel: `execute_operation` now runs the run-and-record half under `asyncio.shield`. The runner's write commits in its worker thread regardless of the client, so the envelope is cached and the audit line written before the cancellation is honored — the dropped client's retry *replays* rather than re-executing or getting 409. Reservation release moved to a `finally` on every path that did not store the envelope; a failing `on_fresh_success` hook is now audited (item 6). Test: cancel mid-run, assert the write recorded, reservation cleared, retry replays with one execution.
2. **MCP subgraph gate** — `require_graph_extension_mcp` refuses subgraph ids (`subgraph_not_addressable`) before loading metadata, matching the REST dependency and GraphQL. Test added.
3. **Fence `lock_timeout`** — explicit rollback + `RESET` on the acquisition-failure path. Note for the record: the leak wasn't reachable — the `SET` was issued inside the same implicit transaction that the failed `pg_advisory_lock` aborted, and the pool's reset-on-return rollback undid it (a real-Postgres test draining the pool passes on both the old and new code). Kept the explicit reset so the invariant doesn't depend on that accident.
4. **Constraint-specific race translation** — new `db/integrity.py` (`violates(exc, *names)` — `diag.constraint_name` first, quoted-name parse as fallback); every race site names its constraint/index and re-raises anything else. Test: an FK violation at the publish-list insert surfaces as `IntegrityError`, not "already present".
5. **Scenario-dimension fallback** — `scalar_one_or_none`, re-raise when the winner isn't there.

PR #1211, opened 2026-08-19
fix: multi-tenancy hardening — tenant binding, connection scoping, query classification, limiter keying

jfrench9:
## Summary

Tier A of the multi-tenancy resilience review that followed #1206 / #1210 (`local/RoboSystems/specs/security/multitenancy-resilience-review.md`, §1). Four independent hardenings, each with red-then-green tests; intended as the patch right behind v1.9.6.

- **Extensions tenant binding fails closed** — a session bound to a tenant schema that does not exist now refuses on its first statement (`invalid_schema_name`, which the surfaces already translate to "not initialized") instead of resolving unqualified names elsewhere. Every graph carrying an OLTP extension gets its schema at creation independent of `create_entity`; a soft-deleted graph reads as gone to every default lookup; teardown drops members' cached access with their rows; report shares and publish lists admit only live, provisioned recipients.
- **Event connections are joined to the calling graph** — `execute-event-block` (REST + MCP) refuses a connection not registered on the graph; the routing id is validated the same way at capture and through metadata patches.
- **Query classifier** — statement-leading `BEGIN`/`COMMIT`/`ROLLBACK`/`CHECKPOINT` are refused as administrative and `COMMENT ON` as schema DDL, at the kernel and again in the graph API; the engine connection pool rolls back anything left open before reuse.
- **Rate limiter keying** — a graph's budget is charged only by callers authorized on it; an API key is an identity to the limiter only once the auth cache has validated it (unverified callers share the anonymous IP bucket).

## Test plan

- [x] `tests/db/test_extensions_isolation.py` — fail-closed bind (real Postgres), including the schema-dropped-mid-session case
- [x] creation-service, deprovision, graph-model, share/publish-list tests for the admission doors
- [x] `tests/operations/event_block/*` — cross-graph connection refused before any QB client is built; capture + patch validation
- [x] `tests/security/test_cypher_analyzer.py`, `tests/graph_api/core/ladybug/test_pool.py`
- [x] `tests/middleware/rate_limits/*` — members-only graph buckets, unverified-key identity
- [x] Local stack: the classifier gate refuses the new verbs on `/query/cypher`; with `RATE_LIMIT_ENABLED=true`, junk-key requests land in `anon_sub:{ip}` and only the owner charges `graph_sub:{graph}`
- [x] `just test-code` clean; full `just test` (in progress at PR open — will confirm on CI)

No migration. No SDK contract change (no request/response shape changed; `execute-event-block` and `update-event-block` gain a server-side graph check only).

### DIFF
--- a/robosystems/db/extensions.py
+++ b/robosystems/db/extensions.py
@@ -190,6 +190,9 @@ def _bind_statement(search_path: str, tenant_schema: str | None) -> str:
   set_local = f"SET LOCAL search_path TO {search_path}"
   if tenant_schema is None:
     return set_local
+  # Validated here as well as by the caller: this is the interpolation, and
+  # the guarantee should not depend on statement order in `extensions_session`.
+  tenant_schema = _sanitize_schema(tenant_schema)
   return (
     "DO $$ BEGIN "
     f"IF to_regnamespace('{tenant_schema}') IS NULL THEN "
--- a/robosystems/db/extensions.py
+++ b/robosystems/db/extensions.py
@@ -38,6 +38,17 @@ IDLE_IN_TRANSACTION_TIMEOUT_MS = 5 * 60 * 1000
 # all sentinel call sites share one string and renames don't drift.
 LIBRARY_GRAPH_ID = "library"
 
+# The schema extensions that give a graph an OLTP tenant schema. A graph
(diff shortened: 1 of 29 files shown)


jfrench9:
@claude can you review?

PR #1212, opened 2026-08-19
fix(security): multi-tenancy tier B — lifecycle, role, and read-only parity across surfaces

jfrench9:
## Multi-tenancy resilience — tier B (parity + robustness)

Follow-up to #1211 (tier A). Closes the same-tenant parity and platform-robustness rows from the 2026-08-18 resilience review: surfaces that authorized correctly but did not all enforce the same *state* — graph lifecycle/subscription, write role, and read-only classification — across REST, GraphQL, and MCP.

### What changed

- **One lifecycle gate on every surface.** The write half of the lifecycle/subscription check is folded into the single write-role gate the command surfaces already share, so a state that blocks writes on one surface blocks them on all. The read half is applied at each read surface's own entry (MCP access, GraphQL context, extensions dependency, analytical views, operators). `require_graph_access` treats a soft-delete stamp as gone before status flips, and gates a subgraph through its parent.
- **The role resolver refuses a graph that is gone.** `GraphUser.get_effective_role` returns no role on a missing/deprovisioned/soft-deleted graph — including for the org-owner implicit grant, which has no membership row to invalidate. Deprovision drops those implicit holders' cached decisions by name.
- **Connections require a write role.** Create, OAuth init/callback, and sync authenticated on membership only; each now runs the write-role gate. Connection reads and policy writes are graph-scoped rather than creator-scoped, so co-admins reach a colleague's connection.
- **MCP parity.** `create-subgraph` over MCP runs the same feature-flag/admin/lifecycle/tier/quota gates as REST; `read-graph-cypher` stays read-only under the queued execution strategies.

### Tests

- Unit + integration coverage for every gate above, each with a red-then-green check against the pre-fix behavior.
- Isolation harness gains an **owner-side landing axis** (a write lands only in its own tenant schema, verified by reading it back through GraphQL on the writing graph and confirming its absence from another graph the same owner controls) plus **vertical probes** for connections, cypher classifier verbs, and MCP admin tools. Local run: 65 probes, 0 leaks.

No migration. Detailed mechanics and the review that scoped this are in the private vault (`specs/security/multitenancy-resilience-review.md` §2).

### DIFF
--- a/robosystems/graphql/context.py
+++ b/robosystems/graphql/context.py
@@ -46,6 +46,7 @@ from robosystems.middleware.auth.dependencies import (
   get_current_user,
 )
 from robosystems.middleware.auth.utils import validate_api_key_with_graph
+from robosystems.middleware.billing.enforcement import require_graph_access
 from robosystems.middleware.extensions import load_graph_metadata
 from robosystems.middleware.graph.types import GRAPH_OR_SUBGRAPH_ID_PATTERN
 from robosystems.middleware.graph.utils.subgraph import is_subgraph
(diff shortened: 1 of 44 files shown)


jfrench9:
@claude can you review

PR #1213, opened 2026-08-19
fix(platform): multi-tenancy tier B — runtime robustness (limiter, sessions, teardown, DSN)

jfrench9:
## Multi-tenancy tier B — robustness

Third and final follow-up to the 2026-08-18 resilience review (after #1211 tier A, #1212 tier B parity). This closes the platform-robustness rows — concurrency, session lifetime, teardown/creation recovery, and error-surface hardening — that don't change the authorization model but make the multi-tenant plane hold under failure.

### What changed

- **§2.4 — the runner limiter stays honest under timeouts.** anyio's `to_thread.run_sync` shields only anyio cancellation, so `asyncio.wait_for`/`timeout` (the MCP transports and the worker's operator budget) raise straight through and release the `CapacityLimiter` token while the worker thread — and its DB connection — is still running. The token uncoupled from the real thread count means the extensions pool overflows under timeouts and a "timed out" write still commits later. The thread run is now shielded, tying the token's lifetime to the thread's, and a post-cancellation failure is logged through the structured logger.
- **§2.5 — non-dependency callers get an independent session.** `platform_session()` and the MCP/subgraph helpers drove the request-scoped session; opened inside a request (a resolver, an MCP tool, a runner thread that inherited the request contextvars) it resolved to the endpoint's own Session and closing it tore that session down mid-flight. They now open an independent `SessionFactory()` session; the scoped registry stays for FastAPI `Depends` only.
- **§2.6 (source) + §2.7 — teardown and creation-rollback stop stranding a database.** A failed creation before the metadata commit deletes the orphaned `.lbug` before releasing the allocation (a freed slot whose file is still on disk poisons the instance). A teardown whose DB-delete fails no longer frees the registry or flips the status — the graph is left stranded (`deleted_at` set) and the sensor re-selects it to retry (reusing its final backup rather than re-dumping). The deprovision job rolls its shared session back per failure so one graph's DBAPI error can't cascade to the rest of the run.
- **§2.8 — harden the materialization error surface.** Redact database credentials from extensions-materialization error text before it reaches an operation result or the logs (diagnostics preserved). Mechanism detail is in the vault.

### Tests
Every fix has unit coverage with a red-then-green check against the pre-fix behavior. Broad affected sweep green (~2,000+ tests across operations, MCP, dagster, extensions, graph_api, connections).

### Deferred (need a staging drill / soak, not just code)
- **§2.6c** — the instance-side parent-level reconciliation sweep. It is a *destructive* auto-delete; it needs a dry-run/observability soak before it deletes real databases. The source fixes above prevent the orphan at creation and teardown, so this is a backstop.
- **§2.10** — the tier-change drain (the `/admin/drain` route is absent); needs a staging drill before/with the fix.

No migration. Full mechanics and the review that scoped this are in the private vault (`specs/security/multitenancy-resilience-review.md` §2). **Sequence the deploy promptly after merge.**

### DIFF
--- a/robosystems/dagster/jobs/graph_lifecycle.py
+++ b/robosystems/dagster/jobs/graph_lifecycle.py
@@ -108,6 +108,16 @@ def deprovision_suspended_graphs(
         error_msg = f"Failed to deprovision {graph_id}: {e}"
         errors.append(error_msg)
         context.log.error(error_msg)
+        # A DBAPI error leaves this shared session in a failed transaction
+        # (e.g. a delete in _clean_pg_records). Without a rollback the NEXT
+        # graph's first statement raises PendingRollbackError and the whole
+        # run cascades — one poison graph took down every graph after it.
+        # Reset so each graph is independent; the failed one is left stranded
+        # (deleted_at set, status not deprovisioned) for the sensor to retry.
+        try:
+          session.rollback()
+        except Exception:
+          context.log.warning(f"Session rollback after {graph_id} failure failed")
 
   context.log.info(
     f"Deprovisioned {deprovisioned_count}/{len(config.graph_ids)} graphs"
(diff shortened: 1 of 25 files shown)


jfrench9:
@claude can you review

PR #1215, opened 2026-08-19
fix(security): pre-onboarding hardening — OLTP statement ceiling, GraphQL off-loop, read-path audit, async op lifecycle

jfrench9:
## Summary

Pre-onboarding hardening pass over the extensions data plane — the availability, detectability and async-correctness follow-up to the multi-tenancy resilience review (#1211–#1213). No authorization semantics change; every fix makes the shared plane hold under load, leave a trail on denial, and keep background operations retryable. Design record and residuals: `local/RoboSystems/specs/security/pre-onboarding-hardening.md`.

- **Extensions OLTP: per-statement ceiling on interactive sessions.** `extensions_session` binds `SET LOCAL statement_timeout` with the search_path (SSM `database/EXTENSIONS_STATEMENT_TIMEOUT_MS`, default 30s); loader / Dagster / migration paths opt out. A cancelled statement is a 504 + `Retry-After` on REST, `statement_timeout` on MCP, `STATEMENT_TIMEOUT` on GraphQL.
- **GraphQL: resolvers off the event loop; unexpected errors masked.** Sync resolvers now run under the shared runner limiter (`OffloadSyncResolvers`); only deliberate `GraphQLError`s reach `errors[]` verbatim (`MaskUnexpectedErrors`, with `requestId`).
- **Ingest: row counting off the loop, per-file row cap, signed Content-Length.** Parquet is footer-only (ranged GET), CSV is streamed record by record, JSON arrays are walked element by element over a text held once (bounded by the size gate — not streamed); the S3 body is closed on every exit. The upload presign signs `ContentLength` when declared and the S3 client is pinned to SigV4 (verified locally that `content-length` joins `X-Amz-SignedHeaders`; the HEAD-measured size in ingest stays authoritative).
- **Read-path audit + attribution.** GraphQL and MCP read denials on user graphs emit `AuthorizationDenied` (same alarm as the write gates). Every auth success publishes a request-scoped principal; the access log now records the caller (it was `user_id=None` on every line); audit lines and security events carry `request_id`, `auth_method`, `api_key_prefix`.
- **Async idempotency**: reserve on miss, release on failed dispatch, evict the pending envelope on the transition into FAILED/CANCELLED (incl. the reaper's DLQ path, now a compare-and-set) — a failed backup/subgraph/materialize is retryable again under the same key. The first terminal status is final: a late/duplicate error after a completion neither flips the record nor evicts the key.
- **Materialization lock**: fail closed (503/409), one lock over both build paths, compare-token extend per table, compare-and-delete release, `materialize_db` tag on tenant sensor runs.

## Deploy notes

- New SSM tunable `database/EXTENSIONS_STATEMENT_TIMEOUT_MS` (defaults in code; `0` disables).
- S3 presigned URLs are now SigV4 — same-region buckets only (true today).
- Materialization now refuses to run when Valkey is unreachable rather than proceeding unlocked; the staleness sensor retries.

## Test plan

- [x] `just test-code` clean
- [x] `just test` — 13,362 passed (incl. real-DB tests for the session ceiling)
- [ ] Staging: a GraphQL read against a tenant graph while a `pg_sleep(60)` runs on another tenant's session; confirm the API stays responsive and the sleeping session gets 504/`STATEMENT_TIMEOUT`
- [ ] Staging: probe a non-member graph via GraphQL/MCP with a valid key; confirm `AuthorizationDenied` in the security stream with `api_key_prefix` + `request_id`

### DIFF
--- a/examples/_scenario/reset.py
+++ b/examples/_scenario/reset.py
@@ -51,7 +51,7 @@ def reset_demo_state(graph_id: str) -> None:
 
   assert_local_extensions_db()
 
-  with extensions_session(graph_id) as session:
+  with extensions_session(graph_id, statement_timeout_ms=None) as session:
     # 1-3. All entries + line items + transactions (tenant-generated;
     # the library itself never authors these). Dimension tables
     # FK back to their parent rows so wipe them first.
--- a/examples/roboinvestor_demo/_reset.py
+++ b/examples/roboinvestor_demo/_reset.py
@@ -40,7 +40,7 @@ def reset_investor_state(graph_id: str) -> None:
 
   assert_local_extensions_db()
 
-  with extensions_session(graph_id) as session:
+  with extensions_session(graph_id, statement_timeout_ms=None) as session:
     # Positions FK both portfolios and securities, so they go first.
     session.execute(text("DELETE FROM positions"))
     session.execute(text("DELETE FROM portfolios"))
@@ -105,7 +105,7 @@ def reset_issuer_share_state(graph_id: str) -> None:
 
   assert_local_extensions_db()
 
-  with extensions_session(graph_id) as session:
+  with extensions_session(graph_id, statement_timeout_ms=None) as session:
     session.execute(text("DELETE FROM report_shares"))
     session.execute(text("DELETE FROM publish_list_members"))
     session.execute(text("DELETE FROM publish_lists"))
--- a/examples/roboledger_demo/_reset.py
+++ b/examples/roboledger_demo/_reset.py
@@ -54,7 +54,7 @@ def reset_demo_state(graph_id: str) -> None:
 
   assert_local_extensions_db()
 
-  with extensions_session(graph_id) as session:
+  with extensions_session(graph_id, statement_timeout_ms=None) as session:
     # 1-3. All entries + line items + transactions (tenant-generated;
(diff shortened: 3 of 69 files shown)


jfrench9:
@claude can you review?

jfrench9:
Review follow-up in 2c26c203:

- **#1 reaper race** — `_fail_operation_sync` is now a WATCH/MULTI compare-and-set: it only moves `pending`/`running` to `failed`, leaves an already-terminal operation alone, and evicts the idempotency envelope only when its own write landed. Tests cover completed-before-read and completed-between-read-and-EXEC.
- **#2 terminal-tier guard** — the SSE store now refuses any move between distinct terminal statuses (same-status re-writes still merge result data) and evicts only on the transition *into* failure, after the CAS succeeds — a lost CAS evicts nothing. Tests: late error after completion, repeated error, lost CAS, both async and sync branches.
- CSV/JSON: the S3 body is closed on every exit (incl. the row-cap early stop). PR body corrected: JSON is walked element-by-element over a size-bounded text, not streamed; the SigV4 binding is verified as a signed header, with the HEAD size staying authoritative.
- Not changed, recorded in the vault queue: `swap.py` no-token fallback (Q5), `reset_principal` symmetry (per-request task context makes it moot today), the `-1` sentinel / `int()` hygiene.

PR #1217, opened 2026-08-19
fix(graph): validate identifiers probed from uploaded files before building staging SQL

jfrench9:
Column names probed from an uploaded file are interpolated into the DuckDB staging DDL as SQL identifiers. This hardens that path:

- Refuse any probed column name outside \`^[A-Za-z_][A-Za-z0-9_]*\$\` (max 128 chars) with a 400 — at the ingest-file edge (from the parquet footer's top-level Arrow names) and again in \`create_table\` / \`insert_into_table\` before any SQL is built.
- Route every identifier interpolation in the staging and materialize paths through a shared \`quote_identifier\` helper (doubles embedded quotes) as defense in depth.

Producer-safety confirmed: every compiled schema's property names (SEC, roboledger, roboinvestor) are already in-class, so no legitimate ingest is affected.

Regression tests drive real DuckDB with a crafted parquet file rather than asserting on generated strings. Full graph_api + operations/graph + dagster suites green.

### DIFF
--- a/robosystems/graph_api/core/duckdb/README.md
+++ b/robosystems/graph_api/core/duckdb/README.md
@@ -152,6 +152,16 @@ SQL, values are bound as parameters, and paths are resolved through
 `get_duckdb_staging_path` with a defense-in-depth check that the result stays
 under the base directory.
 
+Column names are tenant-controlled too — they are probed from the uploaded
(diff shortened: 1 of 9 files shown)


jfrench9:
@claude can you review

PR #1218, opened 2026-08-19
fix(auth): require re-authentication and an interactive session to change email

jfrench9:
Hardens the account-email change path on `PUT /v1/user`:

- **Interactive session only.** The route now uses the JWT-only dependency, so a programmatic API key cannot reach it — matching the rule the passkey surfaces already enforce (a programmatic credential must not be able to mint or pivot to an interactive one).
- **Re-authentication on email change.** Changing the email requires a fresh proof — password re-entry or a `mgmt`-flow passkey assertion — exactly as passkey enrollment and removal do. Name-only updates need no proof.
- **Notice to the previous address.** On an email change, the old address is emailed a security notice naming the (masked) new address. New `email_changed` SES template + notification-job wiring.
- **Lowercase the stored email.** A mixed-case address would never match at login or password reset (both lowercase their input first). Prod checked: 0 of 22 users affected, so this is forward-looking only.
- **Password change hashes at the policy work factor** instead of bcrypt's default cost, so a password change no longer silently downgrades the hash. The current-password check goes through the same `PasswordSecurity` path (consistent >72-byte handling).

Tests: takeover-from-API-key is refused, email change without reauth is refused, name-only needs no reauth, the notice is queued to the previous address, the address is stored lowercased, and a changed password is hashed at cost 14. Full user/auth/ses/dagster suites green.

Note for SDK consumers: `PUT /v1/user` called with only an `X-API-Key` (no bearer) now returns 401. The frontends use JWT and are unaffected; no integration/template path updates the profile. Rides a client minor.

### DIFF
--- a/robosystems/dagster/jobs/notifications.py
+++ b/robosystems/dagster/jobs/notifications.py
@@ -28,6 +28,7 @@ class SendEmailConfig(Config):
   org_name: str | None = None  # For org_invitation emails
   inviter_name: str | None = None  # For org_invitation emails
   passkey_name: str | None = None  # For passkey_enrolled emails
+  new_email: str | None = None  # For email_changed notice (masked new address)
 
 
 class EmailResult:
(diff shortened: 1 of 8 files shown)


jfrench9:
@claude can you review?

jfrench9:
Review addressed in `d30d3c20`:
- **Note 1 (self-email case difference)** — fixed. The email-change entry check is now case-insensitive (`main.py:162-166`), so submitting your own address in any case is a no-op: no reauth prompt, no false "already in use" 409. Test: `test_own_email_in_different_case_is_a_noop`.
- **Note 3 (duplicate `detect_app_source`)** — fixed. Computed once as `app_source` when the email changed and reused by both the change-notice and verification-email blocks.
- **Note 2 (`password.py` uses `get_current_user`)** — intentional, confirmed. A password change requires the current password, so a bare leaked API key can't drive it; the takeover vector this PR closes was email, not password. The JWT-only boundary is specific to the email-change route.

PR #1219, opened 2026-08-19
fix(security): make dormant public-API controls actually enforce

jfrench9:
Several controls existed in the codebase but weren't actually enforcing on the public API. This wires them:

**Request-body size limit (public app).** A new ASGI `RequestSizeLimitMiddleware` bounds body size on the internet-facing surface — by `Content-Length` and by the streamed bytes of a chunked request with no `Content-Length` (closing the gap the graph_api size middleware documents). FastAPI reads the body before dependencies run, so an unbounded body was a pre-auth allocation nothing downstream could cap. The Stripe webhook — which reads its body before verifying the signature — carries a tighter per-path limit and a per-source-IP rate limit.

**Parameter-store reachability.** A deployed boot that couldn't read SSM would serve its whole life on permissive code defaults (flags resolve once at import). `EnvValidator` now refuses to start in a deployed environment when the store is unreachable, no flags preloaded, or rate limiting is off.

**`ORG_GRAPHS_DEFAULT` code default 10 → 1.** The value is snapshotted into `org_limits.max_graphs` at org bootstrap, so a permissive default during an SSM blip durably over-provisions. The safe value becomes the floor; SSM can only raise it. (The four feature-flag defaults are deliberately left as-is — billing-off is intentional for dev/dedicated/forks; the reachability assertion is their safety net.)

**Auth threat store.** The IP-threat/progressive-delay layer stored raw dicts, which redis rejects, so it silently no-op'd — and read back `str` it never parsed. Values are now JSON-encoded/decoded, the progressive-delay table gets a zero-failure key (a fresh IP was being charged the 8th-failure delay), failures are logged instead of swallowed, and the test fake is made as strict as redis so this class of bug can't pass CI again.

⚠️ **Behavior change:** the auth-threat fix activates IP-based login escalation that was inert for months — logins now escalate (a 15-minute IP block after repeated failures). This shares the axis of the existing per-IP auth limiter; the thresholds are tunable, and NAT-shared-IP lockout is worth watching after deploy.

Tests: new size-limit middleware suite (Content-Length + chunked + per-path override), reachability-assertion suite across environments, auth round-trip through a redis-strict fake incl. the zero-delay case. Full middleware/config/security/auth suites green (3031 passed).

### DIFF
--- a/main.py
+++ b/main.py
@@ -38,6 +38,7 @@ from robosystems.middleware.otel.metrics import (
   record_request_metrics,
 )
 from robosystems.middleware.rate_limits import RateLimitHeaderMiddleware
+from robosystems.middleware.request_size import RequestSizeLimitMiddleware
 from robosystems.routers import (
   auth_router_v1,
   billing_router_v1,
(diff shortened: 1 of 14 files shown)


jfrench9:
@claude can you review?

jfrench9:
Review addressed in `6fcfdee2`:
- **Keep-alive desync on the Content-Length 413** — hardened. The reject response now sends `Connection: close`, since the middleware answers without consuming the over-limit body and an unread body on a reused HTTP/1.1 connection could desync the next request. Test asserts the header.
- **Non-standard `ENVIRONMENT` (e.g. a `ci`/`sandbox` deploy target) would fail to boot without real SSM** — confirmed intentional. `deployed = ENVIRONMENT not in ('dev','local','test')`, and the assertion is deliberately fail-safe: any deployed target must reach SSM or it would otherwise serve permissive code defaults for its whole life. A new deploy target that isn't `prod`/`staging` should get SSM access (or be added to the non-deployed set) by design.

The NAT-shared-IP lockout from the now-active login escalation is flagged in the PR body for post-deploy watch; thresholds are tunable via the auth-protection config.

PR #1220, opened 2026-08-19
fix(graph): restore the backup-export grace period and let teardown converge

jfrench9:
Three offboarding fixes, two of them regressions a prior hardening pass introduced.

**Backup export after teardown.** A departing org's OWNER/ADMIN could no longer list or download a torn-down graph's backups: the effective-role resolver denies any gone graph before the handler runs, for everyone. This adds a scoped `allow_deprovisioned` relaxation to `get_effective_role`, threaded from exactly the backup **list** and **download** routes via a `require_graph_access` dependency factory (a named singleton variant) plus `verify_admin_access`. Every other authorizer keeps denying, and the flag bypasses — and never writes back — the JWT and API-key graph-access caches, so it can't leak to another route or persist.

**Teardown convergence.** A teardown whose database delete returned 404 (the file is already gone — e.g. a daemon restart between the delete and the status flip) re-selected the graph and failed the same way every cycle, forever. A 404 now counts as a completed delete so teardown converges; a non-404 error still strands for retry. The invariant that gates the rest of teardown is *the file is not on the instance*, not *we removed it*.

**Staged-upload purge.** `user-staging/{user_id}/{graph_id}/` objects are now deleted **before** the PG rows that index them (parent and subgraph paths). The convergence fix is what makes `_clean_pg_records` actually run in the 404 case, so this ordering had to ride along — once the rows are gone the platform can't enumerate what it still holds.

Tests: the real-auth grace-period route tests that do **not** override the auth dependency (admin passes on a deprovisioned graph, viewer still 403s), model-level `allow_deprovisioned` cases (org owner regains implicit admin; stranger and missing-graph still denied), a convergence test driving a real 404 vs a real 500, and a staged-purge ordering test. Full graphs/auth/models/deprovision suites green (1818 passed).

Remaining backlog (not in this PR): the subgraph search-index leak and the `user_orgs[0]` org resolution.

### DIFF
--- a/robosystems/middleware/auth/dependencies.py
+++ b/robosystems/middleware/auth/dependencies.py
@@ -529,11 +529,13 @@ async def _resolve_user_with_graph(
   )
 
 
-def require_graph_access(allow_deprovisioned: bool = False):
+def graph_access_dependency(allow_deprovisioned: bool = False):
   """Build the graph-membership dependency.
 
-  The default instance (``get_current_user_with_graph``) denies a torn-down
-  graph. ``allow_deprovisioned=True`` is used by exactly the backup-list and
+  Named to avoid colliding with ``billing.enforcement.require_graph_access``
+  (a different check with a different signature). The default instance
+  (``get_current_user_with_graph``) denies a torn-down graph.
+  ``allow_deprovisioned=True`` is used by exactly the backup-list and
   backup-download routes so a departing org's OWNER/ADMIN can export during the
   grace period; it must not be applied to any other route (see
   ``GraphUser.get_effective_role``).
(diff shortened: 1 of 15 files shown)


jfrench9:
@claude can you review?

jfrench9:
Review addressed in `7f615a45`:
- **Bug: subgraph staged-upload purge was a silent no-op** — fixed. `_purge_staged_uploads` now resolves the parent id (`parse_graph_id`) for the `GraphUser` membership lookup while keying the S3 prefix on the graph_id it was handed, so a subgraph's staged objects (under `user-staging/{user_id}/{subgraph_id}/`) are enumerated via the parent's members and deleted. New test `test_subgraph_purge_resolves_members_from_the_parent` (with a realistic kg-hex parent id, since the shared `test_graph` fixture's `kg_<uid>` form has a stray underscore that defeats subgraph parsing).
- **Nit: `require_graph_access` name collision** — fixed. Renamed the new factory to `graph_access_dependency` so it no longer shadows `billing.enforcement.require_graph_access`.
- **Nit: stale step numbering** — fixed (5, 6, 7, 8).

PR #1222, opened 2026-08-19
fix(platform): bound event-loop occupancy and close the post-hoc credit band

jfrench9:
Follow-up increment to the v1.9.8 hardening set — the availability, robustness, and defense-in-depth items surfaced by the round-3 onboarding-readiness review. None is a cross-tenant isolation issue; the isolation surfaces came back clean.

### Event-loop occupancy (noisy-neighbor)
Single-worker API tasks share one event loop, so synchronous work on it stalls every tenant on the task. Moved off the loop / bounded:

- Bedrock `invoke_model` → threadpool; `AIClient` hoisted to a process singleton (no per-request boto3 build + STS)
- Dagster submit/poll → threadpool, with an explicit short client timeout (was the library default)
- shared-repo query handler → plain `def` so Starlette threads it
- SSE subscriber poll timeout moved onto `get_message` (stops a busy-spin)
- platform DB engine gets a `statement_timeout`, matching the extensions engine

### Post-hoc credit band
An over-budget AI call is billed post-hoc; a short pool now drains to zero and records the shortfall instead of deducting nothing, so the band cannot be re-entered unbilled and the next pre-flight denies. Same shape fixed on the shared-repo pool.

### Robustness / defense-in-depth
- worker consumer survives a failure-recording that itself raises (was a process-exit path)
- boot asserts each safety-critical feature flag is *present* in SSM, not merely that some flag loaded
- `delete-subgraph` purges the subgraph's documents from the shared search index
- extensions DSN redacted in the pre-execution write log

Full context and the review that produced this list are in the vault.

### Tests
New/updated coverage for each item; broad regression sweep across billing, operators, SSE, worker, config, and graph_api green (2402 passed). One pre-existing cross-file ordering flake in `test_graph_table.py` (fails on clean main in the same order; passes in isolation) is unrelated.

### DIFF
--- a/robosystems/config/constants.py
+++ b/robosystems/config/constants.py
@@ -33,6 +33,13 @@ MAX_ERROR_MESSAGE_LENGTH = 1000  # characters
 PUBLIC_MAX_REQUEST_SIZE = 10 * 1024 * 1024  # 10 MB
 WEBHOOK_MAX_REQUEST_SIZE = 512 * 1024  # 512 KB
 
+# Per-request timeout for the Dagster GraphQL client (submit / status poll).
+# The library default is 300 s; a status poll that takes longer than a few
+# seconds is a hung webserver, and the call runs from the API's own event
+# loop (background tasks, SSE monitors), so the default would hold every
+# tenant on the task for five minutes per hang.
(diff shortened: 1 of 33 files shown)


jfrench9:
@claude can you review

PR #1223, opened 2026-08-19
fix(security): bound the credential-redaction regex quantifiers

jfrench9:
## Summary

Bound the quantifiers in the two `redact_connection_secrets` regexes so pathological error text cannot make the redaction scan expensive (CodeQL `py/polynomial-redos`, alert #741). The credential these patterns redact is always our own RDS secret (128 chars max), so a 256-char bound changes no real redaction outcome.

## Changes

- **robosystems/security/error_handling.py** — `_CONNSTR_PASSWORD_RE` and `_URL_CRED_RE` now use bounded quantifiers (`{1,256}`) instead of unbounded `+`, with a comment recording the constraint that makes the bound safe.
- **tests/security/test_redact_connection_secrets.py** — two new regression tests: a 128-char (RDS-max) credential is still redacted in both the libpq and URL forms, and adversarial repeated credential-shaped prefixes pass through quickly (`@pytest.mark.timeout(10)`).

## Breaking Changes

None. Internal helper only; no API surface or SDK tier touched.

## Testing

- `uv run pytest tests/security/test_redact_connection_secrets.py` — 6 passed.

### DIFF
--- a/robosystems/security/error_handling.py
+++ b/robosystems/security/error_handling.py
@@ -251,8 +251,11 @@ def is_safe_to_expose(detail_message: str) -> bool:
   return not any(pattern in detail_lower for pattern in sensitive_patterns)
 
 
-_CONNSTR_PASSWORD_RE = re.compile(r"password=\S+")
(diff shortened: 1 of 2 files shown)


jfrench9:
@claude please review this PR

PR #1224, opened 2026-08-20
Harden error-response sanitization, credit-ledger consistency, and auth hashing

jfrench9:
Round-6 pre-onboarding hardening — the two `building` items from the round-6 review, plus the auth-hashing offload folded in. No isolation/blast-radius findings; this is correctness, robustness, and defense-in-depth. Full write-ups live in the private vault (`specs/security/error-response-sanitization.md`, `credit-ledger-integrity.md`).

## What changed

**Error-response sanitization.** A shared whitelist (`safe_error_message`) decides which exception text may reach a caller: caller-facing query/validation errors pass through (connection secrets scrubbed); backend/driver exceptions collapse to a generic message with the full text logged server-side. Wired into the SQL endpoint, the three streaming generators, the Cypher interactive/transient paths, the async operation failure channel (worker consumer, provisioning, query queue), and the MCP search/document tools. Extends the redaction pattern already shipped for the DSN log path.

**Credit-ledger consistency.**
- Replace post-atomic ORM balance write-backs with `expire` / server-side expressions on both credit pools, so a concurrent movement can't be reverted by a stale absolute write.
- Move the shared-repository monthly refill to a calendar-month gate aligned with the 1st-of-month allocation cron (the day-count gate could drift off the cron and skip a month).
- Retry graph credit-pool creation on transient failure, and surface pool-less active graphs in the admin credit-health check.

**Auth hashing.** Run bcrypt hashing/verification off the event loop (login, register, reset, password change, re-auth), and equalize timing on the credential-check miss path.

## Tests
New coverage for the sanitization whitelist, the lost-update regression (verified it fails on the pre-fix code and passes after), the refill-gate behavior, the admin missing-pools surface, the async bcrypt offload + timing equalizer, and the query/MCP error sinks. Full affected-area regression: 1232 passing; lint / format / typecheck clean.

### DIFF
--- a/robosystems/middleware/graph/query_queue.py
+++ b/robosystems/middleware/graph/query_queue.py
@@ -20,6 +20,7 @@ from robosystems.middleware.graph.admission_control import (
   get_admission_controller,
 )
 from robosystems.middleware.otel.metrics import record_query_queue_metrics
+from robosystems.security.error_handling import safe_error_message
 
 
 class QueryStatus(str, Enum):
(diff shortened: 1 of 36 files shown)


jfrench9:
@claude can you review

PR #1225, opened 2026-08-20
Schedule early termination + prepaid disposal NBV fix

jfrench9:
## What

Closes the schedule-lifecycle gaps surfaced by the July 2026 dogfooding close (the pure-MCP acceptance run):

1. **`terminate-schedule` operation** (REST + MCP via the operation registrar) — ends a schedule early at a month-end cutoff **without booking an entry**, in one transaction:
   - `truncate_schedule` (previously fully implemented with zero callers) deletes facts past the cutoff, with its existing guards: month-end cutoffs only, refuses when posted entries exist past the cutoff, deletes stale drafts past it.
   - The remaining obligation chain past the cutoff is voided — `pending` and `classified` rows both (a `classified` row it reaches can only be an undrafted stray: drafts past the cutoff were deleted in step 1 and posted ones block it).
   - The schedule's native SumEquals rule is rewritten to prove the truncated curve instead of the now-unsatisfiable original total.
   - History at or before the cutoff is untouched, so open months the schedule still covers close normally.

2. **Prepaid-schedule NBV fix in `compute_disposal_plan`** — the disposal plan read the credited element's latest instant fact as *accumulated depreciation*. Correct for depreciation schedules (contra account, rising cumulative), inverted for prepaid schedules, where the credited element **is** the asset and the instant fact is the *declining remaining balance*. Self-carried schedules (asset element == credit element) now read NBV directly and derecognize in a single credit line. This makes `asset_disposed` usable on the entire prepaid-amortization class, as its docs always claimed.

3. **Docstring reconciliation** — three docstrings (operations router, MCP schedule-tools module, `DeleteScheduleRequest`) described `asset_disposed` as truncating facts; the handler deliberately voids obligations and keeps facts as history. All three now describe the actual behavior and route no-entry termination to the new op.

4. **Close playbook additions** (version-locked content): a schedule-termination step (terminate BEFORE `promote-obligations`), JE publish semantics (`source='system'` posts locally only — alignment entries mirroring edits already made upstream must not round-trip), and never-retry guidance when `close-period` outlives a client timeout (the close is atomic server-side).

## Why

At the July close, 7 prepaid schedules had to be terminated mid-life. The disposal handler's NBV inversion forced a manual JE + per-event void path, which could only practically reach the current month's obligations — leaving ~150 pre-materialized future obligations that would mature into `pending_obligations` close blockers every month. With this PR, terminating a schedule is one call and retires its whole remaining chain.

## Notes

- `void_pending_obligations` gains `period_start_after` / `include_classified`; existing callers unchanged (the two SQL-shape test assertions updated from `status = 'pending'` to `status IN ('pending')`).
- New surface is additive (generated SDK tier — rides a client minor).
- Deliberately skipped: the `get-period-close-status` display nit where a voided-obligation schedule shows as `pending` — after this op deletes the facts, that state is unreachable.
- Unit tests: terminate command (ordering, scoped void, guard propagation), SumEquals rewrite, and direct `compute_disposal_plan` coverage for both schedule shapes including the inversion regression.

### DIFF
--- a/robosystems/middleware/mcp/tools/playbook_tools.py
+++ b/robosystems/middleware/mcp/tools/playbook_tools.py
@@ -49,7 +49,19 @@ _RECURRING_SEQUENCE: list[str] = [
   "(reopen → re-close) when no external reporting binds them — the usual "
   "default inside the current fiscal year — or book a CATCH-UP entry in "
   "the open period when the reporting cadence has locked prior months. "
-  "Don't close over unreviewed reconciling items.",
+  "A catch-up/alignment JE that mirrors an edit ALREADY MADE in the source "
+  "system must post locally only — create it with source='system' "
+  "(source='manual' publishes to QB at close and would double-apply the "
+  "edit upstream). Don't close over unreviewed reconciling items.",
+  "Schedules ending this period? (an asset sold or transferred, a prepaid "
+  "cancelled or refunded): terminate BEFORE promoting obligations, or the "
+  "sweep drafts entries the termination says should not exist. "
+  "terminate-schedule ends a schedule at a month-end cutoff with no entry "
+  "(truncates forward facts + voids the remaining obligation chain) — use "
+  "it when the GL effect is already booked or none is wanted. "
+  "create-event-block(event_type='asset_disposed') additionally posts the "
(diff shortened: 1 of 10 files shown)


jfrench9:
@claude can you review

PR #1226, opened 2026-08-20
feat(mcp): guard operation exposure; remove unreachable surface

jfrench9:
## Why

Registrar-mounted operations mount a REST route **and** generate an MCP tool from one `OperationSpec`, so the two surfaces cannot drift. Hand-written tools and routes have no such guarantee — a tool needs three independent wiring steps with nothing checking that any happened — and **nothing declared which operations were meant to reach MCP at all**. Across the whole codebase there was exactly one documented exposure decision.

So an unwired capability and a deliberate hold looked identical. That is how a complete, guarded schedule terminator sat unreachable while three docstrings claimed otherwise (#1225).

This makes the intent explicit and enforces it.

## The guard

`tests/middleware/mcp/test_operation_exposure.py`, static (AST + source scan) because the failure mode is *absence* — a runtime check can only exercise tools that were wired, which is the population that is never broken. Four invariants:

1. Every `*Tool` class is instantiated in `manager.py`
2. Every instantiated tool has `get_tool_definition()` called on it — instantiated but unlisted is invisible
3. Every declared tool name is dispatchable in `call_tool`, and a hand-written class colliding with a registrar op fails (registrar dispatch wins, so the class is dead)
4. Every extensions REST operation has an MCP tool **or** an entry in `MCP_EXPOSURE_HOLDS` with a reason

Plus anti-rot: holds naming operations that no longer exist fail, holds that are also exposed fail, and a placeholder reason fails. Every scan has a minimum-population floor so a layout change can't make the tests pass vacuously.

The policy lives in **production code** (`middleware/mcp/exposure_policy.py`), not the test — the reason strings are the artifact; the test only enforces they exist.

## Newly exposed

`update-entity`, `create-report`, `regenerate-report`, converted to registrar specs so both surfaces come from one declaration.

- **OpenAPI-identical**: same paths, same `operation_id`s (`openapi_operation_id` derives them identically), same response models and descriptions, `business_event_type` set explicitly so metrics labels don't shift. **No SDK regeneration.**
- **Auth-identical**: `_require_roboledger_write` and the registrar both enforce membership + provisioning + write role; the registrar additionally applies the extension gate, so this is strictly no-weaker.
- `update-entity`'s empty-body 400 moved to `pre_validate` (which the registrar runs on both surfaces) so it stays distinct from an unmapped `ValueError`'s 422.

The other 15 REST-only operations each got a recorded hold — configuration, security controls, external disclosure, destructive verbs, the filing lifecycle, and the bulk auto-mapper.

## Removed

Surface the reachability sweep found unreachable, each verified by reading its call sites:

- `ExpandToRsGaapCandidatesTool` + its read — complete, instantiated nowhere; deleted rather than wired because its input is a FAC element id the shipped single-pass mapping flow never produces
- Seven taxonomy row-writers, zero callers and zero tests, six with docstrings claiming the block surface invoked them (it constructs `Taxonomy` directly). The module now states the blocks-not-atoms policy where they were
- `CreateMappingAssociationTool`, shadowed by the registrar op of the same name
- A dead second JWT-revocation store. **No security gap** — the live path (`revoked_jwt:{jti}`, checked in `verify_jwt_token`) is untouched — but `maintenance.py` published a counter sourced from the dead store, so it read 0 forever. Now counts the live store through a shared key-prefix constant and reports `jwt_revoked`

None of the removed models reached `openapi.json`, so no client sees a change.

## Also

`update_parent_entity` had the `commit()` → `refresh()` → read defect #1128 fixed at seven other sites; fixed here and `commands/entity.py` added to the `GUARDED` list, since exposing the operation raises its call volume.

## Verification

- `ruff` + format + `basedpyright`: clean
- Full unit suite: 2603 passed. One pre-existing failure in `tests/graph_api/core/duckdb/test_staging_identifiers.py` — order-dependent, passes in isolation (26/26), and **reproduces identically on a clean `main` tree with this branch's work stashed**. Untouched by this PR.


jfrench9:
@claude can you review?

PR #1227, opened 2026-08-20
fix(security): emit the GraphiQL CSP only where the playground is served

jfrench9:
## Summary

The Content-Security-Policy middleware has three variants: strict for API routes, a self-hosted policy for the docs pages, and a relaxed (CDN + inline + eval) policy for the GraphiQL playground. The playground is mounted in development only, but the variant matcher chose its policy by URL shape alone — so outside development the graph-scoped GraphQL path, which only ever answers with JSON there, still carried the playground's header. This gates the variant on the same condition as the mount, so production emits the strict policy on every response except the two docs pages and their static assets.

## Changes

- **`main.py`** — `csp_variant_for_path` takes a keyword-only `graphiql_enabled` flag, **defaulting to `False`** so a caller that omits it can never relax production. The security-headers middleware passes `env.is_development()`, which is the condition the `GraphQLRouter` mount already uses for `graphql_ide`. Docstring and comment updated to say why.
- **`tests/test_main_csp.py`** — the pure-matcher pins for the relaxed variant now pass `graphiql_enabled=True`; a new test asserts all three GraphQL URL shapes return `"api"` when the flag is off *and* when it is omitted. The operations-route and path-traversal tests now run with the flag on so they still prove something non-trivial.
- **`tests/test_main_security_baseline.py`** — two end-to-end pins through the real middleware on the graph-scoped GraphQL path: with `is_development` false, the CSP contains no `'unsafe-inline'`, no `'unsafe-eval'`, and no `https://` origin; with it true, the playground keeps its CDN policy. (Patched `env.is_development` itself rather than `env.ENVIRONMENT`, since the latter doesn't reach a classmethod.)

No route, response body, or schema changes — header value only, on one path, outside development.

## Breaking Changes

None. No change to the GraphQL schema, the operations envelope, or any REST shape; nothing for the client SDKs to regenerate.

## Testing

- `uv run pytest tests/test_main_csp.py tests/test_main_security_baseline.py` — 33 passed.
- `just test-code` — ruff, format check, basedpyright, cf-lint all clean.
- Full `just test-all` not run in-session; the change is confined to the header middleware and its tests.

## Certification

### DIFF
--- a/main.py
+++ b/main.py
@@ -101,18 +101,22 @@ _SENSITIVE_PATH_PREFIXES = (
 )
 
 
-def csp_variant_for_path(path: str) -> str:
+def csp_variant_for_path(path: str, *, graphiql_enabled: bool = False) -> str:
   """Which CSP variant a path gets.
 
   - "docs": Swagger UI / ReDoc pages and their assets, self-hosted from
     /static — no third-party script origins and no 'unsafe-inline' script.
   - "graphiql": the GraphiQL playground, which loads React/GraphiQL from
-    CDNs and needs the historical relaxed policy.
+    CDNs and needs the historical relaxed policy. Returned only while the
+    playground is actually served (``graphiql_enabled`` — development
+    only); elsewhere the graph-scoped GraphQL path answers with JSON and
+    gets the strict policy like every other API route. The default is
+    closed so a caller that omits the flag can never relax production.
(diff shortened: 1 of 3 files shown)


jfrench9:
@claude please review this PR

PR #1233, opened 2026-08-21
Persist the close receipt; fix a second-libarrow collision that disabled PageRank weighting

jfrench9:
## Summary

Two unrelated threads that landed together, both about a result that existed but was never recorded anywhere durable.

**The close receipt.** A period close's result — entries posted, the QuickBooks/local split, statement stamping, rule summaries — lived only in the HTTP response. A close that *succeeded* but whose transport failed left no record on the books, so the operator reconstructed what happened from four separate state reads and had to know not to retry. On the MCP surface that is routine rather than rare: a real close runs past the 25s client timeout. `fiscal_periods.close_receipt` is now written in the same transaction as the `status='closed'` flip, so a committed close always carries its receipt and a rolled-back one carries none.

**A second libarrow.** `icebug` bundles its own `libarrow` beside pyarrow's, and the two collide on Arrow's process-global filesystem registry. Once anything in a process has imported networkit, `pyarrow.fs.LocalFileSystem()` raises `ArrowKeyError` permanently — import order does not help. Chasing the staging-identifier test that has been failing in CI for days led to a live instance of the same thing: `_load_pagerank_scores` has never once succeeded in production.

## Changes

**Close receipt** (`c30fd3b`)
- `fiscal_periods.close_receipt` JSONB + extensions migration `0034` (public schema and every tenant schema via the `TenantOps` fan-out).
- Written by `PeriodCloseService.close()` alongside the status flip. It is a projection, not a dump — `calendar` is a live ORM object, and a receipt that changes after the fact is not a receipt.
- Typed `CloseReceiptResponse` rather than a free dict, surfaced on `get-period-close-status`, plus a `has_close_receipt` flag on the calendar's period rows so listings stay compact.
- No backfill. Periods closed before this read as closed with no receipt — a real state, not a failed close. A stored receipt that fails validation degrades to none rather than raising, since that read is exactly what an operator calls when a response was lost.

**PageRank weighting** (`7035e6c`) — the one to look at closely
- `_load_pagerank_scores` raised on every call, the surrounding `except` swallowed it at debug level, and it returned `{}`. `weighted_score = freq * (1 + pagerank)` collapsed to `freq * 1`, so disclosure profiles have been silently frequency-only.
- Both tests around it were asserting the broken output: `test_weighted_score_uses_pagerank` asserted `ws >= freq - 0.001`, which the degraded path (`ws == freq`) also satisfies, so it and `test_weighted_score_degrades_without_pagerank` were checking the same result. It now asserts the exact product and a strict increase, guarded against a vacuous pass.
- The `except` logs at warning now; the debug level is why this survived.

**Parquet path resolution** (`1fea2ca`, `20d07ee`)
- Staging-identifier test fixtures write through a handle. This is what has been red in CI — deterministic, not flaky, and `-x` in the addopts meant it aborted the run and masked everything after it.
- `entity_sync/transform.py` likewise. Not a live defect (separate job, separate step subprocess), but it was the last path-string write and an exception to a convention is what the next writer copies.

**Housekeeping** (`5958360`) — ignore the HuggingFace cache that `ArelleClient`'s process-global `XDG_CACHE_HOME` causes to be written into the source tree.

## Breaking Changes

None. The GraphQL and REST additions are purely additive — a new `CloseReceipt` type, an optional `closeReceipt` field, and an optional `has_close_receipt` flag.

Worth flagging for coordination rather than as a break: `robosystems-python-client` carries a checked-in `schema.graphql` with a regen-must-be-clean gate, so these additions will surface there on its next `just refresh-schema`.

## Testing

- Full suite green: 13,506 passed, ruff/format/basedpyright clean.
- The close-receipt tests run against **real Postgres**, deliberately. Every extensions test mocks `extensions_session`, so nothing in-suite executes the raw SQL this touches — the gap that let #1228 ship two defects onto live books. The writer/reader contract test was mutation-checked: renaming a key on the writer fails it.
- The strengthened PageRank test was mutation-checked the same way — reverting the fix fails it.
- Close receipt verified end to end on a local tenant: `2026-07` closed, the receipt landed on the row matching the response field-for-field, and read back through GraphQL. A period closed earlier correctly reports no receipt.

## Certification

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -144,3 +144,10 @@ GEMINI.md
 robosystems/adapters/sec/arelle/EDGAR/
 # Arelle schema cache - downloaded during build
 robosystems/adapters/sec/arelle/cache/
+# Collateral of the line above: ArelleClient._setup_cache_directory sets the
+# process-global XDG_CACHE_HOME to robosystems/adapters/sec so Arelle resolves
+# its own cache, and every other XDG-respecting library in that process then
+# caches there too. huggingface_hub is the one that shows up locally
+# (.agent_harnesses.json + xet/logs). Ignore the whole tree, not one file —
+# the next such library will land here as well.
+robosystems/adapters/sec/huggingface/
--- a/migrations/extensions/versions/0034_close_receipt.py
+++ b/migrations/extensions/versions/0034_close_receipt.py
@@ -1,0 +1,55 @@
+"""close receipt — fiscal_periods carries the outcome of its own close
+
+The close result (entries posted, the QB/local split, statement stamping,
+rule summaries) existed only in the HTTP response. A close that SUCCEEDED
+but whose transport failed — the MCP surface's 25s client timeout is the
+live case — left no record on the books, so the operator reconstructed
+what happened from four separate state reads and had to know not to retry.
+
+``close_receipt`` is written by ``PeriodCloseService.close()`` in the same
+transaction as the ``status='closed'`` flip, so a committed close always
+carries its receipt and a rolled-back one carries none.
+
+Nullable with no backfill: periods closed before this shipped have no
+receipt to reconstruct, and inventing one would be worse than its absence
+(readers distinguish "closed, no receipt" from "closed, here is what
+happened"). Public schema AND every existing tenant schema get the column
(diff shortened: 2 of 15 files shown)


jfrench9:
@claude can you review

PR #1238, opened 2026-08-21
fix(graph-api): derive capacity accounting from the graph id

jfrench9:
## Summary

Capacity accounting on the graph node read two different signals for the same decision: the cap counted top-level databases by name shape, but took its exemption from a request field. Both sides now derive from the graph id, so the exemption cannot be claimed for a database the count would still charge.

## Changes

- `graph_api/core/ladybug/manager.py` — the cap exemption derives from the graph id, matching the count on the line above. The same derivation is already used in `core/ladybug/config.py`.
- `graph_api/models/database.py` and `graph_api/client/client.py` — `is_subgraph` is retained for what it still controls, skipping schema application (a subgraph inherits its parent's schema through the fork). Both descriptions said it governed the capacity check; corrected.
- Tests — two existing capacity tests used graph ids containing an underscore, which the count has always treated as not-counted, so they were asserting against an id shape that cannot occur for a top-level graph. Updated to realistic ids, and added a regression test on each side asserting the request field cannot buy an exemption.

Reviewer note: the schema-application path deliberately still keys off the flag. Shared repositories such as `sec_historical` contain an underscore, so deriving there would skip schema application for them and break shared-repo creation.

## Breaking Changes

None. `is_subgraph` remains on the request model with its existing default; only what it governs has narrowed. Subgraph creation is unaffected — a subgraph id always carries the separator the derivation reads.

SDK note: description text only on an internal service model; no operation, field, or type changes.

## Testing

- `uv run pytest tests/graph_api` — **1367 passed**, including the two updated tests and the two new ones.
- Full unit suite not run locally; leaving it to CI.

## Certification

### DIFF
--- a/robosystems/graph_api/client/client.py
+++ b/robosystems/graph_api/client/client.py
@@ -680,9 +680,10 @@ class GraphClient(BaseGraphClient):
     """Create a database.
 
     ``schema_type`` is one of ``entity``/``shared``/``custom``;
-    ``repository_name`` applies to shared databases. ``is_subgraph`` bypasses
-    the node's ``max_databases`` check, since a subgraph is accounted against
-    its parent's tier limit rather than the instance's.
+    ``repository_name`` applies to shared databases. ``is_subgraph`` tells the
+    node to skip schema application, since a subgraph inherits its parent's
+    schema through the fork. Capacity accounting is derived from ``graph_id``
+    on the node, not from this flag.
     """
     payload = {
       "graph_id": graph_id,
--- a/robosystems/graph_api/core/ladybug/manager.py
+++ b/robosystems/graph_api/core/ladybug/manager.py
@@ -106,18 +106,21 @@ class LadybugDatabaseManager:
         status_code=status.HTTP_400_BAD_REQUEST, detail="Graph ID is required"
       )
 
-    # Subgraphs are exempt from the cap: they share the parent's slot, and are
-    # identified by the "_" in their name.
+    # Subgraphs are exempt from the cap: they share the parent's slot. Both
+    # sides of that accounting derive from the same thing — the "_" in the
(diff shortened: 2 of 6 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Confirmed — this was a real regression and the review caught it correctly. Verified each step:

- `materialize.py:1579` builds `{graph_id}-wip` with a hyphen, and the comment directly above the call at :1592 documents that it depends on the exemption.
- `list_databases()` already excluded `-wip`/`-prev`, documented as "so they never count toward the capacity limit" — so the codebase already had the notion my derivation was missing.
- `max_databases` is 1 on every tier (`graph_creation_service.py:658`), so the live primary fills the node and the WIP would have taken a 507.

Fixed in `9f5eb88c`. Both halves now run through one `counts_toward_capacity()` predicate covering subgraphs, shared repositories, and the blue-green suffixes; `list_databases()` calls the same function, so there is a single definition rather than two that agree by coincidence.

Added `test_blue_green_wip_is_exempt_on_a_full_node`, and verified it fails against the previous logic rather than passing either way.

Also re-ran the search that missed this. There are exactly four `is_subgraph=True` call sites: two in `subgraph_service.py` (ids built as `{parent}_{name}` by `types.py:438`, so underscore-exempt), one in `shared_repository_service.py` (`sec_historical`-style, same), and the WIP. All four are covered. Two of those turned out to be Graph-model creates rather than `create_database` calls, so the cap never applied to them at all.

Updated the stale docstring at `materialize.py:1658`, which still claimed the flag drives the exemption.

On the minor point about the log line: agreed, and it is reworded — it no longer claims a "parent's slot", which was wrong for both the shared-repo and WIP cases.

PR #1239, opened 2026-08-21
docs: state security invariants positively in comments

jfrench9:
## Summary

A consistent habit had built up across the codebase: comments justified a control by describing the failure it prevents. That reads well in review and badly in a public repository, because the description is often more useful to a reader than the code is. This restates those comments as the rule that holds — what the code guarantees, and what a future change must not break.

Comments, docstrings and one test class name. No behavioral change.

## Changes

Roughly 40 sites, all following the same rewrite: state the invariant, drop the counterfactual.

- **Auth and session handling** — `models/core/user/user.py`, `middleware/auth/{utils.py,dependencies.py,README.md}`, `routers/auth/{login.py,register.py,utils.py}`, `security/{password.py,device_fingerprinting.py}`. Where a bound is deliberate, the remediation direction is kept, since that is the part a maintainer acts on.
- **Query and tool surfaces** — `security/cypher_analyzer.py` and its test. The tokenizer comment now states the requirement directly: it must close a backtick identifier exactly where the engine does, because divergence changes which keywords classification can see. Fixtures and assertions are untouched; the test class is renamed after the property it pins.
- **Operations** — `operations/{locking.py,oidc.py,event_block/commands.py,graph/deprovision_service.py,taxonomy_block/*,information_block/rules/engine.py,operators/tool_access.py,extensions/materialize.py}`.
- **Routers and middleware** — `routers/graphs/{health.py,backups/*,subscriptions.py,operator/execute.py}`, `routers/user/main.py`, `middleware/{request_size.py,rate_limits/*,mcp/tools/*}`, `graphql/resolvers/{ledger.py,investor.py}`.
- **Config and infrastructure** — `config/{rate_limits.py,validation.py,constants.py}`, `cloudformation/{s3.yaml,waf.yaml,postgres.yaml,api.yaml,dagster.yaml}`, two workflow comments, `.claude/commands/deploy.md`.

Two things deliberately **not** changed:

- Forward-looking justifications of an existing guard ("without this, the lookup falls through to …") are left alone. They explain why the code is shaped as it is, which is exactly what a comment is for. Only narration of a specific past failure was rewritten.
- `middleware/logging.py`'s suspicious-request indicator list is code, not a comment, and is log-only. Changing it is a behavior change and out of scope here.

## Breaking Changes

None.

## Testing

- `just test` — **3797 passed, 18 skipped** on this branch.

## Certification

### DIFF
--- a/.claude/commands/deploy.md
+++ b/.claude/commands/deploy.md
@@ -13,7 +13,7 @@ Most deploys (~95%) go green untouched and need no attention. The real use case 
 
 - **`gh` reads are free; the deploy trigger is not.** Reading runs, jobs, and logs (`gh run list/view/watch`) needs no confirmation. **Triggering or re-triggering a deploy** (`gh workflow run`, `just deploy`) is an outward-facing action — confirm the target (env + ref) with the user first, and default to watching a run they already started.
 - **AWS is read-only here.** `describe-*` / `list-*` only. CloudFormation changes and stack deletions are the user's to run — never `create-stack`/`update-stack`/`delete-stack` directly.
-- **Never deploy the default branch to prod.** Production should ride a version tag / release branch produced by the release workflow. **Nothing in the pipeline enforces this** — the prod workflow has no ref guard, and the deploy helper will accept a branch without complaint. So check the ref *before* triggering, and when reviewing history (`gh run list --json headBranch,displayTitle`), flag any past prod run that rode a branch rather than a tag.
+- **Never deploy the default branch to prod.** Production should ride a version tag / release branch produced by the release workflow. Verify the ref *before* triggering rather than assuming the pipeline will reject a bad one, and when reviewing history (`gh run list --json headBranch,displayTitle`), flag any past prod run that rode a branch rather than a tag.
 - **`just deploy` defaults are dangerous if you're not explicit.** The recipe defaults to the **prod** environment and the **current branch**. Always pass both arguments.
 - **Output can be sensitive.** Failure logs name internal hostnames, stack names, and resource IDs. Don't paste raw infra detail into anything public; summarize.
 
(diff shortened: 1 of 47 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Good critique — it's aimed at judgment rather than correctness, and it's right on most of the list. Acted on four of the five in `fac066b4`:

- **`information_block/README.md`** — agreed, a clear over-trim. "Dozens of blocks per graph" is a scaling trigger, not a disclosure, and vaguer prose bought nothing. Restored.
- **`password.py` / `login.py`** — agreed. The ~0.7 s figure is a property of bcrypt cost 14 that anyone can measure, and a maintainer needs it to judge whether a cost change breaks the equalization. Restored, with an explicit note to re-measure if the cost factor moves.
- **`routers/user/main.py`** — partially. The attack chain stays out, but you're right that the load-bearing half was *why an API key specifically* matters here: keys are long-lived and widely copied, so they are not proof of presence. That reasoning is back without the chain.
- **`cloudformation/postgres.yaml`** — partially, by inverting it. Now states what REQUIRED would *add* rather than what DISABLED *accepts*. Same threat-model information, framed as the addition rather than the gap.

**Left as trimmed: `config/rate_limits.py`.** The "~1% CPU observed" figure is live production state, which is both the thing to keep out of a public comment and the thing that rots silently — a stale utilization number would mislead the next person sizing these. The decision it supported (don't halve limits on existing customers) survives on its own. The measurement belongs in the internal record.

Two things I verified separately while going through this:

- **No route-handler docstrings changed.** The one `routers/` docstring in the diff is `_enforce_shared_repository_agent_limits`, an undecorated private helper, so the published OpenAPI spec is untouched by this PR. That matters because FastAPI uses a handler's docstring as its spec description when no explicit `description=` is given.
- **AST comparison across the whole PR**: 36 of 37 changed Python files parse to an identical AST with docstrings stripped; the only difference is the renamed test class. All 7 changed YAML files parse identically, checked with a loader that handles CloudFormation intrinsic tags — plain `safe_load` silently reports CF templates as equal because both sides fail to parse the same way.

PR #1240, opened 2026-08-21
docs: replace historical narration with present-tense statements

jfrench9:
## Summary

Comments across source, CloudFormation and workflows recorded what the codebase used to do, what an earlier approach was, or when a specific incident happened. Git history holds that; a comment should describe what is true now. Each rewrite keeps the constraint and drops the account of how it was learned.

Comments, docstrings and markdown only. No behavioural change.

## Changes

**Historical narration → present tense** (~45 sites)

- **CloudFormation** — `worker.yaml`, `graph-volumes.yaml`, `graph-infra.yaml`, `graph-ladybug-replicas.yaml`, `security.yaml`, `audit.yaml`. The load-bearing constraints all survive: why the second metric namespace is required, why the disk alarm needs a Metrics Insights query rather than a partial dimension set, why SSM skips must normalise to exit 0, why the failure rule filters on the document name. What goes is the dates, the incident retellings and the abandoned-approach postmortems around them.
- **Source** — `middleware/operations.py`, `db/platform.py`, `dagster/jobs/{backup_cleanup,graph}.py`, `models/core/graph/graph_backup.py`, `models/extensions/{element,roboledger/fiscal_period}.py`, `operations/{graph/commands/metadata,graph/graph_creation_service,connection_service,event_block/__init__,operators/operator_context,information_block/*}.py`, `adapters/sec/{ixbrl_parser,pipeline/sensors}.py`.
- **Config and security** — `config/{env,deprovisioning,valkey_registry,rate_limits}.py`, `security/{auth_protection,password}.py`.
- **Routers** — `routers/__init__.py` (three removed-surface tombstones, one of which named a removed router as "too dangerous for public API"), `routers/graphs/__init__.py`, `subgraphs/__init__.py`, `auth/sso.py`, `offering.py`, `subscriptions.py`.
- **Workflows, scripts and docs** — `service-refresh.yml`, `deploy-vpc.yml`, `refresh-graph-asg/action.yml`, `bin/setup/gha.sh`, `bin/lambda/graph_volume_{manager,monitor}.py`, `CLAUDE.md`, `frameworks/**`.

Also drops a roadmap aside naming unbuilt frameworks in `frameworks/fac/packages/README.md`. The framework names themselves stay throughout — `rs-ifrs`, `rs-call-report`, `rs-statutory` are public accounting standards, and they explain why the framework layout has a sibling slot per regulatory regime.

**Duplication** — `examples/custom_graph_demo/README.md` explained schema customization twice; the reference now points at the worked version under Advanced Usage.

## Two distillation targets checked and deliberately not changed

Both were flagged as redundant. Neither holds up:

- **The admission-control buffer-pool argument** was reported as "written out in full in five files". It is in three, each at a different altitude — the constants' home (`config/defaults.py`), the enforcing code (`graph_api/core/admission_control.py`), and the README whose point is *contrasting* the two controllers and which already cross-references the first. The other two are one-line summaries in bullet lists. Deduplicating would make each site worse.
- **`middleware/robustness/README.md`**'s "Putting it together" section was reported as restating the three preceding snippets. It shows the four helpers *composed* in one handler, which is the thing the individual sections cannot show.

## Breaking Changes

None.

## Testing

- `just test-code` — **passed** (ruff, format, basedpyright, cf-lint-all, lint-actions), re-run green by the pre-commit hook on both commits.
- **AST comparison** vs `main`: 30 of 31 changed Python files parse to an identical AST with docstrings stripped. The one difference is a comment inside a usage-example string in `config/valkey_registry.py`.
- **YAML comparison** vs `main` (CloudFormation-tag-aware loader): all six CF templates and `service-refresh.yml` parse identically. `deploy-vpc.yml` and `refresh-graph-asg/action.yml` differ only because the edited text sits inside a shell `run:` block and an input `description:` respectively — a `#` comment and a UI label.
- Full unit suite not re-run since the last green run on this base; leaving it to CI.

## Certification

### DIFF
--- a/.github/actions/refresh-graph-asg/action.yml
+++ b/.github/actions/refresh-graph-asg/action.yml
@@ -53,8 +53,8 @@ inputs:
   completion-wait-minutes:
     description: >-
       How long to monitor for completion when wait-for-completion is true.
-      Ignored otherwise. The default of 20 preserves the previously hardcoded
-      1200s; raise it for a large fleet — Dagster's replica equivalent budgets
+      Ignored otherwise. Raise the default of 20 for a large fleet — Dagster's
+      replica equivalent budgets
       2 hours, sized for ~100 instances. Note the job timeout is derived from
       this, so raising it raises the cap rather than fighting it.
     required: false
--- a/.github/workflows/deploy-vpc.yml
+++ b/.github/workflows/deploy-vpc.yml
@@ -415,9 +415,8 @@ jobs:
             SECURITY_ENV="${{ inputs.environment }}"
           fi
 
-          # The API ALB alarms this job used to parameterize now live in api.yaml
-          # and reference the ALB and target group as CloudFormation resources,
-          # so there is nothing to resolve here any more.
+          # The API ALB alarms live in api.yaml and reference the ALB and target
+          # group as CloudFormation resources, so nothing is resolved here.
           SECURITY_PARAMS="ParameterKey=Environment,ParameterValue=$SECURITY_ENV \
                       ParameterKey=EnableConfig,ParameterValue=${{ inputs.security_config_enabled }}"
 
--- a/.github/workflows/service-refresh.yml
+++ b/.github/workflows/service-refresh.yml
@@ -160,11 +160,10 @@ jobs:
   # Graph Container Refresh (fleet-wide, via SSM)
   # ============================================
   # One job, one Lambda invocation, one tag-targeted SSM command per node-type
-  # group. This replaced a collector that enumerated every instance into a
-  # GitHub Actions matrix and gave each its own runner job. That design had a
-  # hard ceiling — a matrix caps at 256 jobs, and the fleet runs one database
-  # per instance — and it paid runner-minutes to sleep through a 30-minute
-  # busy-wait. Runner time here is now bounded and independent of fleet size.
+  # group. Do not fan the fleet into a job matrix: a matrix caps at 256 jobs,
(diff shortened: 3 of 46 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Both correct, fixed in `cc295f28`.

**The first one is the substantive catch and I had it backwards.** The measured 78-93% busy at baseline and the ~2.4x improvement at 12000 IOPS / 500 MB/s are the *constraint* — they are why the shared tier is provisioned at that specific value rather than any other figure above baseline. Only the date was tombstone. Cutting them violated this PR's own stated rule, and the botched sentence made it obvious. Restored without the date:

```
# rewrites) into this volume. At the 3000-IOPS baseline the volume sits 78-93%
# busy moving ~20 MB/s; at 12000 IOPS / 500 MB/s the same tables run ~2.4x
# faster, which is why the shared tier is provisioned there. The other tiers
# stay at baseline.
```

The second was a rewrap artifact from the same edit pass; reflowed.

Because a garbled replacement is a *class* of error rather than a one-off, I swept the whole diff for siblings — repeated n-grams within and across added lines, and orphaned short comment lines that would indicate a bad rewrap. Nothing else turned up.

Re-verified after the fix: `just test-code` green, 30 of 31 changed Python files still AST-identical with docstrings stripped (the one exception remains the prose inside `valkey_registry.py`'s `print()` example), and every CloudFormation template still parses identically under the tag-aware loader.

PR #1241, opened 2026-08-21
feat(mcp): give tool descriptions a shared, enforced skeleton

jfrench9:
## Summary

A tool `description` is not a comment — it is the prompt the model reads to decide whether to call the tool and with what arguments, at selection time, with no chance to ask a follow-up. Across 51 of them the vocabulary had drifted into **29 different section headers**, and **14 tools had no sections at all**, so a model comparing tools was comparing different shapes.

This adds a test that makes the shape a contract, then brings every description to it.

**This PR changes functional text.** Unlike the preceding documentation PRs, "no behavioural change" is not a claim I can make here — changing a prompt changes tool selection. That is the point of the change, and the reason it is separated from the inert prose work in #1240.

## Changes

**The contract** — `tests/middleware/mcp/test_tool_description_contract.py` parses every description out of `tools/*.py` (AST, so no live graph client is needed) and enforces:

1. Section headers come from a fixed vocabulary: `WHEN TO USE`, `PARAMETERS`, `RETURNS`, `NOTES`, `RELATED TOOLS`, `WORKFLOW`.
2. No section repeats — two `**NOTES:**` blocks read as two unrelated lists.
3. Every tool says **when to use it** and **what it returns**. Those are the two selection questions; a description that only says what a tool *is* leaves the model inferring applicability from the name.

Length is deliberately **not** asserted. `close-period` is 3.7k characters because a close has real preconditions and a caller who improvises one corrupts a ledger.

**Bringing all 51 descriptions to the contract** — mostly mechanical: `INPUTS`→`PARAMETERS`, `TIP`/`TIPS`/`USAGE TIPS`→`NOTES`, `ALSO READ`→`RELATED TOOLS`, and so on, with content-bearing sections (`STATEMENT TYPES`, `DOCUMENT TYPES SEARCHED`) folded into `PARAMETERS` where they describe an argument's allowed values. Fourteen tools gained the sections they lacked, including every `semantic_memory` and `subgraph_write` tool.

**A hole in the contract, found in review** — the parser keyed "is this a tool description" off the string being multi-line, which is a property of how a description happens to be written rather than of what it is. Four descriptions are a single line, so the contract silently skipped them — and they were four of the ones with no sections at all, the exact case it exists to catch. It now keys off `inputSchema`, which is what actually distinguishes a tool definition from a parameter definition. Coverage went from 47 tools to 51.

**Client-visible text corrected** — two tool names that exist nowhere in the repo (`list-schedule-structures`, `get-schedule-facts`), a roadmap aside advertising an unimplemented `hybrid` write policy, two references to when a feature was introduced, and a schema-type/data-type glossary in `get-graph-schema` that taught the model what `STRING` means.

**`.claude/agents/ladybug-architect.md`** — four state claims corrected: a capacity value that disagreed with `.env.example`/`pytest.ini`, two code constants (`LBUG_MAX_CONNECTIONS_PER_DB`, `LBUG_CONNECTION_TTL_MINUTES`) documented as environment variables with a debugging step telling the agent to change one, a connection limit that disagreed with `pool.py`, and a router inventory missing `metrics.py`. Values now point at the config rather than restating it.

## One thing left alone, deliberately

`constants.py`, `example_queries_tool.py` and `manager.py` use emoji (`✅ GOOD` / `❌ BAD`, `⚠️`) inside prompt text. `CLAUDE.md` bans emoji in "production code or logs" — but these are contrast markers in text a model reads, which is a category that rule predates, and stripping them from a prompt has a plausible downside. Flagging rather than deciding it inside a PR about prompt quality.

## Breaking Changes

None to any API, model or SDK surface. No operation names, arguments or response shapes change.

## Testing

- `tests/middleware/mcp` — **855 passed**, including the 154 new contract assertions.
- The contract test was written before the fixes and initially failed on 33 of them, so it is known to catch what it claims to. After the coverage fix it failed on a further 4, which is how those were found.

## Certification

### DIFF
--- a/.claude/agents/ladybug-architect.md
+++ b/.claude/agents/ladybug-architect.md
@@ -90,6 +90,7 @@ Graph API (FastAPI on EC2:8001)
 - `databases/restore.py` - Restore operations
 - `databases/tables/` - DuckDB staging tables
 - `databases/tables/materialize.py` - Stage to graph ingestion
+- `metrics.py` - Prometheus metrics endpoint
 - `databases/vector_search.py` - LanceDB vector search
 - `databases/semantic_memory.py` - Semantic memory operations
 - `databases/schema.py` - Schema introspection
@@ -171,7 +172,7 @@ Confirm the set with `ls .github/workflows/ | grep graph` before acting on it �
 
 `.github/configs/graph.yml` is the source of truth — **read it rather than quoting sizes from memory**, since instance types and RAM have been resized more than once. Its `instance:` block per tier carries `instance_ram_gb` and `databases_per_instance`; ASG min/max counts come from GitHub variables (`just gha-list LBUG`).
 
-Shape as of this writing (verify before relying on it):
+Shape (read `.github/configs/graph.yml` for current values):
 
 ```
 ladybug-standard: dedicated instance, 3 subgraphs max
@@ -201,13 +202,9 @@ CLUSTER_TIER=ladybug-standard|ladybug-large|ladybug-xlarge|ladybug-shared
 LBUG_DATABASE_PATH=/data/lbug-dbs
 
 # Capacity Settings
-LBUG_MAX_DATABASES_PER_NODE=10
-LBUG_MAX_MEMORY_MB=2048
-LBUG_MAX_MEMORY_PER_DB_MB=0  # 0 = auto-calculate
-
-# Connection Management
-LBUG_MAX_CONNECTIONS_PER_DB=10
-LBUG_CONNECTION_TTL_MINUTES=30.0
+LBUG_MAX_DATABASES_PER_NODE
+LBUG_MAX_MEMORY_MB
+LBUG_MAX_MEMORY_PER_DB_MB  # 0 = auto-calculate
 
 # Admission Control
 LBUG_ADMISSION_MEMORY_THRESHOLD=0.85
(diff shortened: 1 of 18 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Both findings were right, and the first was bigger than a nit — fixed in `429416b9`.

**The exemption list was inert because the contract had a hole.** `_tool_descriptions()` keyed "is this a tool description" off the string being multi-line, which is a property of how a description happens to be *written* rather than of what it *is*. Four descriptions are single-line, so they never entered `DESCRIPTIONS` — and they were four of the ones with no sections at all, which is the exact case the contract exists to catch. A test that looks registry-wide but skips the worst offenders is worse than no test.

It now keys off `inputSchema`, which is what actually separates a tool definition from a parameter definition. Coverage went 47 → 51. The four that surfaced (`remember`, `recall`, `update-memory`, `forget`) failed immediately and now carry the skeleton. The exemption list is gone rather than extended: it named two of the four, never fired, and would have exempted precisely the tools that needed the rule.

**The PR description was wrong and I've corrected it.** `semantic_memory_tools.py` genuinely had no diff when you read it — the claim was written from the audit's finding rather than from the diff, which is the confabulation the PR-authoring guidance exists to prevent. It is true *now*, as a consequence of the coverage fix, but it was not true then. The counts are also corrected: 51 tools not 52, 14 without sections not 10, 855 tests not 843.

On the sandbox limitation — worth noting the direction it cut. Not being able to run the suite is what pushed you to verify by direct code inspection, and that is how both findings turned up; the green checkmark I was relying on would not have surfaced either, because the hole made the test pass.

Re-verified: `just test-code` green, `tests/middleware/mcp` 855 passed.

PR #1242, opened 2026-08-21
fix(graph-api): enforce the body cap on streamed requests

jfrench9:
## Summary

Harden the Graph API's request-body size limiting so it enforces at the same two
points the public API's limiter already does.

Rather than add the missing check in place, this lifts the body handling into a
shared `BodySizeLimitMiddleware` that both services subclass. A subclass supplies
only *the limit*; it inherits *the enforcement*. That is the part worth reviewing:
the two limiters are separate implementations of one control, and a shared base is
what keeps them from drifting.

## Changes

**`robosystems/middleware/request_size.py`**

- New `BodySizeLimitMiddleware`: header check, streamed byte count, buffer-and-replay,
  and the 413 (with `Connection: close`, since the over-limit body is never consumed).
- `_limit_for(path) -> (limit, label)` is the single extension point. The label names
  the family in the 413 and the log line.
- `RequestSizeLimitMiddleware` is now a subclass holding just the public app's
  longest-prefix path overrides. Its behaviour and response body are unchanged.

**`robosystems/graph_api/middleware/request_limits.py`**

- Now a subclass supplying the per-endpoint-family limits (query / schema / body).
  The limits themselves and the constants they derive from are unchanged.
- Drops `BaseHTTPMiddleware` for plain ASGI. The Graph API streams query results,
  so keeping it out of the response path is a small bonus rather than the point.
- 413 body shape changes from `Request query too large. Size: N, Max allowed: M` to
  the shared `Request query too large. Max allowed: M bytes`. Nothing parses it —
  the request size is in the warning log, where it is useful.

**`tests/graph_api/test_request_limits.py`** — new; this middleware had no coverage.
Covers both enforcement points, each family limit, a family limit under a chunked
body, the defaults resolving from the shared constants, a streamed response passing
through, and a client disconnect replaying to the app in order.

**`tests/middleware/test_request_size.py`** — module docstring only: it described the
two limiters as differing, and points at the new file for the shared behaviour.

## Breaking Changes

None. No published-SDK surface is touched — neither the stable tier (facades, the
integration template's emit path) nor the generated tier. The Graph API is
VPC-internal and has no OpenAPI-derived client. The only response-shape change is
the internal 413 detail string described above.

## Testing

- `tests/graph_api/test_request_limits.py` + `tests/middleware/test_request_size.py` — 20 passed.
- `tests/graph_api` — 1378 passed.
- `just lint` / `just format` / `just typecheck` — clean (0 errors, 0 warnings).
- The new tests were run against the previous implementation to confirm they fail on
  it: 4 of 10 fail. A test that passes either way would not have established anything.
- Verified against a real `create_app()` Graph API instance, not just a synthetic
  app: an under-limit body reaches the route, an over-limit body is rejected on both
  enforcement paths, and the limiter constructs with the real defaults.
- The disconnect assertions were confirmed to fail when deliberately broken, since
  `asyncio_mode = auto` makes a silently-unawaited async test easy to miss.

## Certification

### DIFF
--- a/robosystems/graph_api/middleware/request_limits.py
+++ b/robosystems/graph_api/middleware/request_limits.py
@@ -1,18 +1,20 @@
-"""
-Request size limiting middleware for Graph API.
+"""Request size limiting for the Graph API.
 
-Prevents DoS attacks by limiting request body sizes.
+The body handling — header check, streamed byte count, buffer-and-replay — is
+:class:`~robosystems.middleware.request_size.BodySizeLimitMiddleware`. This
+module supplies only the limit, which varies by endpoint family.
 """
 
-from fastapi import Request, status
-from fastapi.responses import JSONResponse
-from starlette.middleware.base import BaseHTTPMiddleware
+from __future__ import annotations
+
+from starlette.types import ASGIApp
 
 from robosystems.config.constants import GRAPH_MAX_REQUEST_SIZE, MAX_QUERY_LENGTH
 from robosystems.logger import logger
+from robosystems.middleware.request_size import BodySizeLimitMiddleware
 
 
-class RequestSizeLimitMiddleware(BaseHTTPMiddleware):
+class RequestSizeLimitMiddleware(BodySizeLimitMiddleware):
   """Reject oversized request bodies before they are read into memory.
 
   The limit is chosen per endpoint family — queries, schema DDL, everything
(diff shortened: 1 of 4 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Reviewed locally with full session context. Both of your points confirmed and no action needed on them — the family-matching semantics are byte-for-byte the old ones, and I re-checked that nothing parses the old `Size: N` detail string. Agreed on the `/query` substring nit being pre-existing and not worth changing here.

Two things I found that the static read couldn't reach, both now addressed:

1. Buffering the body makes the limiter responsible for replaying a client disconnect faithfully, and nothing covered that. Exercised the ASGI messages directly — disconnect-before-body, partial-body-then-disconnect, and the fall-through to live `receive` after the buffer drains all behave correctly, with no hang. Encoded as a test in `30ed4228`. Since `asyncio_mode = auto`, I also confirmed the new async assertions fail when deliberately broken rather than passing un-awaited.
2. The PR description named the enforcement gap plainly while the change is still undeployed. Rewritten to describe the area hardened instead, and it now accounts for the fourth changed file, which the original body omitted.

Also swapped one test's reach into Starlette's `user_middleware` internals for constructing the middleware directly.

PR #1243, opened 2026-08-21
fix(config): make declared configuration surface match what resolves

### LINKED ISSUE #1236: Rate-limit env vars read via a shadowed getattr resolve to defaults
### Summary

`robosystems/middleware/rate_limits/rate_limiting.py:25` defines a local `get_int_env` that shadows the one in `robosystems/config/env.py`. It resolves values via `getattr(env, key, default)`, so a name that `env.py` never declares silently falls back to its default.

None of the `RATE_LIMIT_*` names that module reads are declared in `config/env.py`. The result is a configuration surface that reads as tunable but is inert — setting any of them in the environment or SSM has no effect.

This is the same defect class the OTEL README records as fixed for `OTEL_TRACES_ENABLED`; that fix was applied to the one site rather than swept.

### Impact

Medium

### Environment

Production

### Steps to Reproduce

1. Set any `RATE_LIMIT_*` value read by `rate_limits/rate_limiting.py` in the environment or SSM.
2. Exercise the corresponding limiter.

### Expected Behavior

Either the value takes effect, or the constant is not presented as configurable.

### Actual Behavior

The value is ignored; the hardcoded default applies.

### Additional Context

Two candidate fixes:

1. Declare the names in `config/env.py` so they resolve, and add a test asserting every key passed to `get_int_env` exists on `env`.
2. Drop the local `get_int_env` and inline the constants, so they stop reading as configurable.

Worth a repo-wide check for other `getattr(env, ...)` call sites with the same shape, since the pattern is what makes the failure silent.

### LINKED ISSUE #1237: models.extensions omits Agent, Event, EventHandler from re-exports
### Summary

`robosystems/models/extensions/__init__.py` re-exports the RoboLedger and RoboInvestor models, but omits `Agent`, `Event`, and `EventHandler`. All three are exported from `robosystems.models.extensions.roboledger`, so the inconsistency is only at the parent package.

```
>>> import robosystems.models.extensions as m
>>> [hasattr(m, n) for n in ("Agent", "Event", "EventHandler")]
[False, False, False]
```

### Impact

Low

### Environment

Development

### Steps to Reproduce

1. `from robosystems.models.extensions import Agent`

### Expected Behavior

Resolves, as its sibling models do from the same package.

### Actual Behavior

`ImportError`. The working import is `from robosystems.models.extensions.roboledger import Agent`.

### Additional Context

Cosmetic — no documented path depends on the parent-package import (`models/extensions/README.md` documents these under the roboledger section, where they do resolve). Filing so the export list is consistent rather than because anything is broken.

Note the `Agent` name is the REA counterparty model, not the AI Operator layer.

jfrench9:
## Summary

Two places where a module's declared surface did not match what actually
resolves, each fixed with a guard that compares the two as sets so they cannot
drift apart again. Both were found by the documentation audit rather than by
anything failing.

Closes #1236. Closes #1237.

## Changes

**Burst rate limits (#1236)**

`rate_limiting.py` defined a local `get_int_env` that shadowed the one in
`config/env.py` and resolved values through `getattr(env, key, default)`. None of
the **26** `RATE_LIMIT_*` names it read are declared on `env`, so every one
answered with its default — a configuration surface that read as tunable and
governed nothing.

They are not meant to be tunable, so the fix is to stop them advertising
otherwise rather than to make them resolve:

- `config/rate_limits.py` — the documented home for rate limiting — gains a
  `BURST_LIMITS` mapping, keyed by what each limit *is* (`api_key`, `scim_ip`,
  `billing`) rather than by an environment variable name that never existed. The
  rationale comments move with the values, since that is where the numbers now live.
- `rate_limiting.py` drops the shadowing helper and reads the mapping. A wrong key
  is now a `KeyError` on the path, not a silent default.
- New `tests/config/test_burst_limits.py` asserts the keys read and the keys
  defined are the same set **in both directions**.

**Extensions model re-exports (#1237)**

`robosystems.models.extensions` omitted `Agent`, `Event`, `EventHandler` and
`event_dimensions` from its imports and `__all__`. The issue named three; the
fourth turned up on inspection, and settles it as oversight rather than curation —
`event_dimensions` sits alongside three other `*_dimensions` tables that are all
re-exported. New `tests/models/extensions/test_reexports.py` compares the parent's
surface to each extension package's `__all__` as a set, in both *reachable* and
*advertised in `__all__`* form.

## Breaking Changes

None, in either direction.

Every rate limit keeps its exact current value. `get_int_env` was module-private
to `rate_limiting.py` and is not published SDK surface; the four model re-exports
are additive. No OpenAPI or GraphQL shape changes, so neither SDK tier moves.

## Testing

- Every limit verified unchanged: captured the effective value at all 26 call
  sites before the change and compared after — identical, with no unused entries
  in the new mapping. This was the main risk, since the edit touched 26 sites in a
  security-relevant module.
- Both new guards were run against `main`'s code to confirm they fail on the
  defect they describe, then against the fix.
- `tests/middleware tests/config tests/models tests/routers/auth` — **4228 passed**.

## Notes for the reviewer

- The rate-limit test suite patched `get_int_env` with fixed return values at 28
  sites, so it asserted against a mock rather than the real limit. Those patches
  are gone and the assertions now read the constants — the tests got stronger, but
  it is the largest part of the diff and worth a look.
- Not addressed here, to keep the diff scoped: `config/constants.py` still holds
  five burst limits of the same nature (`AUTH_RATE_LIMIT_LOGIN_DEFAULT`,
  `JWT_REFRESH_RATE_LIMIT_DEFAULT`, the two `RATE_LIMIT_SSE_CONNECTIONS*`). One of
  them is the *count* whose *window* is now in `BURST_LIMITS`, so the pair is split
  across two modules. Consolidating is a separate change.

## Certification

### DIFF
--- a/robosystems/config/rate_limits.py
+++ b/robosystems/config/rate_limits.py
@@ -447,3 +447,42 @@ class RateLimitConfig:
         return EndpointCategory.GRAPH_READ
 
     return None
+
+
+# Per-endpoint burst limits, in requests per window. These bound how fast a
+# single caller can hit one endpoint family; the per-tier limits above bound
+# volume. They are constants on purpose — a limit that can be raised at runtime
+# is a limit an attacker benefits from raising, and the safe direction (down)
+# is already reachable by deploying. Nothing reads these from the environment.
+BURST_LIMITS: dict[str, int] = {
+  # Identity-based limits for the general API bucket, per minute.
+  "api_key": 1000,  # 60k/hour possible
+  "jwt": 500,  # 30k/hour possible
+  "anonymous": 10,  # 600/hour possible
+  # Authentication. Windows are seconds, not counts.
+  "auth_attempts": 10,
+  "auth_window": 300,
+  "login_window": 300,
+  "register_window": 3600,
+  "sensitive_auth": 60,
+  "auth_status": 600,  # 10/second — polled by the login screen
+  "logout": 300,
+  "sso": 100,
+  "oidc": 120,  # 2 per flow
+  "mfa": 120,  # 2 per handshake
+  "passkey_management": 60,
+  # SCIM provisioning, which bursts during a directory's initial sync.
+  "scim": 120,  # per token
+  "scim_ip": 300,  # per IP
+  # Ordinary API surfaces, per minute.
+  "general_api": 200,
+  "public_api": 600,  # divided by 10 for anonymous callers
+  "user_management": 600,  # 10/second
+  "tasks": 200,
+  "analytics": 100,
+  "connection_mgmt": 30,
+  "sync_ops": 50,
+  "backup_ops": 10,  # expensive operations
+  "billing": 60,  # checkout polls at ~20/min
+  "webhook": 1200,  # anonymous → 120/min per IP
+}
--- a/robosystems/middleware/rate_limits/rate_limiting.py
+++ b/robosystems/middleware/rate_limits/rate_limiting.py
@@ -13,6 +13,7 @@ from ...config.constants import (
   RATE_LIMIT_SSE_CONNECTIONS,
   RATE_LIMIT_SSE_CONNECTIONS_WINDOW,
 )
+from ...config.rate_limits import BURST_LIMITS
 from ...security import SecurityAuditLogger, SecurityEventType
 from .cache import rate_limit_cache
 from .subscription_rate_limits import (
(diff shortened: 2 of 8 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Both observations are right.

Removed `MockEnvConfig.RATE_LIMIT_API_KEY` in `67f5085d` — `EnvValidator` never reads it and `env.py` never declared it, so it was a mock of surface that did not exist, which is the same illusion this branch removes. Worth doing here rather than in the `constants.py` follow-up.

Agreed on leaving the `# 60k/hour possible` comments; they came over verbatim and I would rather not silently reword rationale while moving it.

On the tooling note: this is the sixth review in a row that could not execute the suite, so it is the configuration rather than this PR. Worth saying that it has not cost much — reading the code found real defects in four of those six, including the substantive one on #1244. I would not enable `Bash(uv run *)` casually: it would let a pull request modify a test and have the reviewer execute it with the workflow's credentials.

PR #1244, opened 2026-08-21
fix(sec): stop submitting a retired job from the SEC connection provider

### LINKED ISSUE #1235: sec_entity_sync job is submitted but never registered in Definitions
### Summary

The `sec_entity_sync` Dagster job is defined and submitted, but never collected into `Definitions`, so submitting it fails.

`robosystems/adapters/sec/pipeline/entity_sync/` defines `sec_entity_sync_job` and exposes its own `get_dagster_components()` (`entity_sync/__init__.py:53`). Neither `robosystems/dagster/definitions.py` nor `robosystems/adapters/sec/pipeline/__init__.py` references it, so it is absent from the repository's `Definitions`.

Meanwhile `robosystems/operations/providers/sec_provider.py:224` submits `job_name="sec_entity_sync"`, and `operations/providers/registry.py:72` wires `sync_sec_connection` in as the `sync` operation for SEC connections — a live customer path.

### Impact

High

### Environment

Production

### Steps to Reproduce

1. Create a SEC connection on a graph.
2. Trigger a sync for that connection.
3. The provider submits `sec_entity_sync` to Dagster.

### Expected Behavior

The job is registered in `Definitions` and runs.

### Actual Behavior

Dagster has no job by that name, so the submit fails.

### Additional Context

Found during a documentation audit — the same pipeline is also undocumented in all three adapter READMEs, which is likely why the missing wiring went unnoticed. Worth fixing both together.

Suggested fix: collect `entity_sync.get_dagster_components()` in `adapters/sec/pipeline/__init__.py` alongside the other SEC components, and add a regression test asserting every `job_name` the provider registry can submit resolves in `Definitions`.

jfrench9:
## Summary

`sync_sec_connection` submitted a Dagster job named `sec_entity_sync`, which
`Definitions` does not carry — so the SEC connection sync operation could only
fail at the scheduler. The per-connection entity-sync pipeline was retired in
favour of consolidating SEC filings into the shared repository nightly, and its
package was left behind along with the call that submitted it.

Closes #1235.

## Changes

- **`operations/providers/sec_provider.py`** — `sync_sec_connection` now reports
  that there is nothing to pull, mirroring `sync_external_connection`. The provider
  protocol already documents its return as "a run id **or a human-readable
  status**", so this case was anticipated; SEC just never used it.
- **Removed `adapters/sec/pipeline/entity_sync/`** (7 modules) and its test file.
  Nothing referenced the package outside itself: no entry in the SEC pipeline's
  component collector, none in `definitions.py`, and no documentation.
- **`tests/operations/providers/test_sec_provider.py`** — three tests asserted the
  submission and its run config, so they passed against a job that could not run.
  Replaced with tests asserting the absence of a submission.
- **New `tests/operations/providers/test_submitted_jobs_are_registered.py`** —
  covers the class rather than this instance.

## The guard is the point

A provider submits a job *by name*. Nothing links that string to the job it refers
to, so a job that is renamed, retired, or never registered leaves the call site
compiling, importing and unit-testing green, and fails only at the scheduler — on a
path a feature flag can keep dark for months. The new test walks the provider
modules for literal `job_name=` arguments and asserts each resolves in
`Definitions`. It currently covers `qb_sync`, and was verified to fail against the
previous provider with the message it was written to produce.

## Breaking Changes

None to any published surface. No OpenAPI or GraphQL shape changes, so neither SDK
tier moves — the return type of the sync operation is unchanged (`str`), and the
value was previously a run id for a run that could not start.

Behaviourally, calling sync on a SEC connection now returns an explanation instead
of raising at the scheduler. `CONNECTION_SEC_ENABLED` defaults to `false`, so the
provider is not registered unless explicitly enabled.

## Testing

- `tests/adapters/sec tests/operations/providers tests/dagster` — **1559 passed, 9 skipped**.
- The new guard was run against the previous provider to confirm it fails there,
  and passes now.

## Certification


jfrench9:
@claude please review this PR

jfrench9:
The sync-lock finding is correct, and worse than described — I verified it and it is **live today**, not merely newly reachable.

`CONNECTION_EXTERNAL_ENABLED` defaults to **true**, and `sync_external_connection` returns a status string, so the leak is already happening on that provider: lock acquired, nothing releases it, every later sync on that connection 409s for the full 30-minute TTL. `qb_load` is the only consumer of `sync_lock_id` anywhere in the repo, so QuickBooks was the only provider the release path ever covered. You were also right that raising is what masked it for SEC — the dispatch-failure handler was doing the release by accident.

Fixed in `0863f4b7`. The registry now declares whether a provider's sync starts a run that outlives the call (`starts_async_run`), and the dispatcher releases the lock on a clean return when it does not. QuickBooks behaviour is unchanged, and the failure-path release became a shared helper rather than a second copy of it. Tests cover both directions — releasing for an async provider would let a second sync race the run just started, so that is asserted too — and the release test was verified to fail without the fix.

On the empty-parametrize nit: correct, and it is the same hole I closed in #1241's contract test a few PRs ago, so applying the lesson in one file and not the other was mine to catch. Added `test_the_scan_finds_something_to_guard`.

On the vestigial patch — I kept it, and I think it earns its place. The old code imported `submit_dagster_job_sync` inside the function body, so patching the source module is exactly what would intercept a re-introduced call; `assert_not_called` is the regression guard against the submission coming back. Added a comment saying so, since the reason is not obvious from the line.

PR #1245, opened 2026-08-21
refactor(rate-limits): give the burst limits one home

jfrench9:
## Summary

Three commits. The first is the follow-up #1243 called out; the second and third
came from looking for the last of the documentation audit's "distillation" work and
finding something better.

**1 — Give the burst limits one home.** `config/constants.py` still held seven rate
limits alongside the ones that moved to `BURST_LIMITS`, and the split had become
visible at the call site — the login attempt *count* from one module, its *window*
from the other:

```python
limit = AUTH_RATE_LIMIT_LOGIN_DEFAULT      # config/constants.py
window = BURST_LIMITS["login_window"]      # config/rate_limits.py
```

Five move into `BURST_LIMITS` next to the windows they pair with. Two —
`AUTH_RATE_LIMIT_WINDOW_LOGIN` and `AUTH_RATE_LIMIT_WINDOW_REGISTER` — are deleted:
they had **no consumers** and duplicated `login_window` (300) and `register_window`
(3600) exactly. That duplication is most likely *how* the pair came to be split.

**2 — Remove a database-limit knob that enforces nothing.**
`MultiTenantUtils.check_database_limits()` is a no-op whose docstring states limits
are "enforced at the instance level based on `LBUG_MAX_DATABASES_PER_NODE`". Nothing
reads that name — it is not declared in `config/env.py` — and the real enforcement is
the tier's `max_databases` from `.github/configs/graph.yml`, applied by the Graph
API's database manager. A no-op naming a nonexistent variable as the mechanism
answers the question wrongly, which is worse than not answering it.

Dead surface, so no deprecation cycle: the method was called only by a test asserting
it does nothing, and the env var was *set* in the test workflow and *mocked* in
`MockEnvConfig` while being read by nothing. Removed from all four; the agent doc now
points at where the limit actually lives.

**3 — Two prose fixes** found in the same pass: the 0002 migration called prior
columns "concocted" three times (the accurate statement is that they were not
XBRL-intrinsic, which is also *why* they were replaced), and `create-pr.md` restated
the `test-all` stage list that the justfile defines and `test.md` already enumerates.
That list has gone stale before; the copy that is not the recipe is the one that goes
stale.

## Breaking Changes

None. Every rate-limit value is unchanged, and all seven names were internal to
`rate_limiting.py`. The removed method and env var were dead in the sense the
contract allows to skip deprecation: never functional, no consumer, and removal
changes symbol resolution only. No OpenAPI or GraphQL surface, so neither SDK tier
moves.

## Testing

- Every value checked against `main` by reading the old `constants.py` out of git:
  `20→20, 5→5, 3→3, 10→10, 60→60`.
- The drift guard from #1243 covered the move **for free** — it asserts keys read and
  keys defined are the same set in both directions, so the five new entries could not
  have been added unwired or left orphaned. Passes at 31 entries.
- The new `constants.py` guard was run against `main` to confirm it fails there,
  naming all seven.
- Confirmed `LBUG_MAX_DATABASES_PER_NODE` and `check_database_limits` have zero
  remaining references anywhere in the repo.
- `tests/middleware/graph tests/config tests/middleware/rate_limits` — **1371 passed**.
- `just lint` / `just format` / `just typecheck` / `just lint-actions` — clean.

## Note on what is *not* here

This started as the audit's remaining "verbosity / distillation" work. Most of it did
not survive inspection, and the cuts were declined rather than made:

- `cloudformation/security.yaml`'s 46-line alarm comment — every paragraph prevents a
  specific mistake, including that the alarm must not be rewritten as `SEARCH`
  (rejected at deploy time) and that it cannot report its own brokenness.
- `operations/locking.py`'s module docstring — carries the background-jobs-wait /
  request-handlers-do-not policy and why the unlocked bug is silent.
- `.claude/agents/ladybug-architect.md`'s file inventory — all 21 paths resolve, so it
  is accurate config, not stale padding.

The audit flagged these by doc-to-code ratio and line count, which measures density
rather than redundancy. A locking policy module with a high prose ratio is correct.

## Certification


jfrench9:
@claude please review this PR

jfrench9:
Nit fixed in `993a30a4` — and thanks for catching it, because it is the same edit-artifact class the reviewer caught on #1240. I swept the rest of `constants.py` for sibling empty banners programmatically rather than eyeballing it; that was the only one.

Heads up that your review predates two commits. The PR has grown since, and the description is rewritten to match:

- `8751f21f` removes `MultiTenantUtils.check_database_limits()` and `LBUG_MAX_DATABASES_PER_NODE`. Same defect class as #1236 — the method is a no-op whose docstring names that env var as the enforcement mechanism, nothing reads the name, and the real enforcement is the tier's `max_databases` in `.github/configs/graph.yml`. It was *set* in the test workflow and *mocked* in `MockEnvConfig` while being read by nothing. Worth a second look since it spans four files.
- Also two prose fixes: the 0002 migration described prior columns as "concocted" three times, and `create-pr.md` restated a stage list the justfile owns.

On the sandbox note — that is six or seven reviews running now, and it has not cost much: static reading has found a real defect in most of them, including the substantive one on #1244. Not something I would change, since enabling `Bash(uv run *)` would let a pull request modify a test and have the reviewer execute it with the workflow's credentials.

PR #1246, opened 2026-08-22
fix: correct defects the documentation sweep introduced

jfrench9:
## Summary

A ten-agent read-only review of the eight-PR hygiene sweep. This fixes what it
confirmed. The headline: **the sweep introduced defects of its own, and the worst
of them were in the MCP tool descriptions** — which are prompt text a model acts
on, not documentation.

### The sweep's own defects — MCP tool descriptions

Renaming section headers without re-reading what sat under them, plus RETURNS
lines written from intent rather than from the handler:

- **`delete-subgraph` claimed a backup that never happens.** It promised "the
  backup id when `backup_first` is set", and described the flag as creating a
  backup, **default true**. `SubgraphService.delete_subgraph_database` accepts
  `create_backup` and does nothing with it — its own docstring says "not yet
  implemented". That told a model a hard delete with `force` available was
  reversible. The most dangerous line the sweep produced.
- **`write-graph-cypher` invented "rows affected".** The handler returns the rows
  the query itself RETURNs — empty for a bare CREATE.
- **`get-information-block` turned "not implemented" into "the block has none".**
  `dimensions` is never populated by any envelope builder, so an empty list said
  nothing about the block. The caveat that was removed was load-bearing.
- **Three PARAMETERS sections named no parameters** — the header was applied to
  lists of enum values and document types, leaving the sole required parameter
  unnamed in each. `search-documents` also advertised hybrid search while
  `semantic` defaults to false.

### The contract test was still vacuous, a second way

It resolved a description only when `ast.literal_eval` accepted it. A tool that
assembled its text into a variable, or concatenated a shared guidance block,
escaped **silently** — and both that did were violating the contract.
`read-graph-cypher`, the most-called tool on the surface, broke all three rules at
once. The parser now resolves names and concatenation against a package-wide
symbol table; coverage 51 → 53, and both are brought to the contract.

The discoverability floor was `>= 40` against 51 real tools — eleven could have
dropped out unnoticed, the same failure one level up. Pinned to the exact count.

### Four claims the sweep got wrong

- `auth_protection`'s delay table said the lookup "floors to the highest tier at
  or below the count". It is an exact-key get whose fallback is the *highest*
  delay, so a gap charges the 8th-failure delay, not the next tier down. A correct
  explanation was replaced with a wrong one, on a caller-facing throttle.
- The SQL endpoint's public description lost **"30s timeout, 10K row limit"**.
  Both are real and enforced (`MAX_QUERY_ROWS`, `QUERY_TIMEOUT_SECONDS`), and the
  route still declares a 408 with nothing left to explain it. That is OpenAPI text
  an integrator sizes pagination against, not narration.
- The connection-sync description still described the SEC pipeline removed in
  #1244, including a `last_sync` timestamp that no longer advances.
- `counts_toward_capacity`'s docstring called a shared repository exempt. `sec` is
  a primary and **does** hold a slot; only its subgraphs are exempt. Read the
  other way it would license a real over-admission bug.

### Disclosure — one HIGH the sweep missed

`bin/entrypoint.sh` carried an internal appsec finding ID, a step-by-step account
of how the per-IP limiter degrades and what value defeats it, and a claim the
mitigation was not deployed — **also false**, since `cloudformation/api.yaml` sets
`FORWARDED_ALLOW_IPS` to the VPC CIDR. Rewritten to state the rule, wildcard
imperative included, without the mechanics or the posture claim.

Five vault citations also survived, because the earlier pass searched prose files:
three in taxonomy JSON-LD `description` values, one in a test docstring, one in a
setup script. Re-swept across every tracked file type; none remain outside
`.claude/commands`, where naming a paired runbook is deliberate policy.

### Two test holes closed

- Nothing asserted which providers hand the sync lock to a run. Flipping
  `quickbooks` to `starts_async_run: False` left the **entire suite green** — one
  token from the concurrent write the lock exists to prevent.
- Nothing asserted `counts_toward_capacity("sec") is True`, the case whose wrong
  answer costs an outage rather than a spurious 507.

Both verified to fail against the defect before being trusted.

## Breaking Changes

None. Two public OpenAPI descriptions change text only — the SQL one *regains*
limits that were always enforced. No schema, endpoint, or field changes, so
neither SDK tier moves; both are a doc regen on a client minor.

## Testing

- `tests/middleware/mcp tests/graph_api tests/operations/providers tests/config` — 1557 passed.
- Both new assertions verified to fail against the pre-fix code, then pass.
- The widened contract immediately surfaced `read-graph-cypher` failing all three
  rules — the tool it had been skipping.
- Full unit suite on merged `main` before this branch: **12,400 passed**.

## Certification


jfrench9:
@claude please review this PR

PR #1247, opened 2026-08-22
fix(config): make the feature flags real switches

jfrench9:
## Summary

Two findings from the post-sweep review, both cases of a declared surface not
matching what the code does.

The feature-flag question started as "clean up the unused SSM flags" and the
answer turned out to be that **there are none** — `bin/setup/aws.sh` seeds 31
flags and every one is read. The gap runs the other way.

### Twelve flags the code reads were never seeded

They do not exist in a fresh environment and `just ssm-list <env> features` does
not show them:

`PASSWORD_AUTH_ENABLED` · `PASSKEYS_ENABLED` · `MFA_ENFORCEMENT_ENABLED` ·
`SSO_OIDC_ENABLED` · `SCIM_ENABLED` · `STORAGE_RECLAIM_ENABLED` ·
`CONNECTION_EXTERNAL_ENABLED` · `SEMANTIC_MEMORY_ENABLED` · `MCP_GRAPHQL_ENABLED` ·
`OPERATOR_POST_ENABLED` · `EXTENSIONS_PROMOTION_AUTO_DISPATCH` ·
`INTUIT_REPORTS_TESTING_MIGRATION`

Five are the identity and access controls; `STORAGE_RECLAIM_ENABLED` gates a daily
sweep that deletes data. This contradicts the script's own stated reason for
seeding `GRAPH_USAGE_ALERTS_ENABLED` at its code default — *"a kill switch you
cannot find is not one."*

**Each is seeded at the default the code already applies**, read from the
`get_parameter_value(...)` call in `config/env.py` rather than from my local
resolved values, so no behaviour changes.

### Three flags that read as live and governed nothing

- **`SSE_ENABLED`** — resolved through SSM, assigned to `self.sse_enabled`, and
  never read. The SSE kill switch did nothing. It now refuses the connection in
  `add_connection`, so every SSE surface sits behind one check and a client that
  cannot stream falls back to polling. This is the one that mattered: an operator
  flipping it mid-incident was getting no effect while believing they had acted.
- **`GRAPH_HEALTH_CHECKS_ENABLED`** — declared in `env.py` and documented in
  `graph_api/client/README.md` under "feature flags read through
  `robosystems.config.env`", with zero readers. Removed; its three siblings in the
  same block are genuinely wired.
- **`SEC_INCREMENTAL_PIPELINE_ENABLED`** — existed only in two comments telling an
  operator to set it. The schedule is declared `STOPPED` and started in the Dagster
  UI, as the module header two hundred lines above already says. The same comment
  also had the cadence wrong ("every 3 hours" against a `0 21 * * 1-5` cron).

### `/v1/graphs/extensions` advertised an extension the create path rejects

It listed `knowledge` as available while `EntityCreate.validate_extensions` 422s
it — offered by one half of the API and refused by the other, and published to the
OpenAPI spec and both SDKs. Its missing entry in the endpoint's own `display_names`
map was the tell.

`knowledge` is a real schema and stays: it is installed on a subgraph via
`subgraph_type="knowledge"` (knowledge-only, no base), and is in the MCP
`create-subgraph` enum with `memory` as an alias. It appeared here only because the
endpoint enumerated every `SchemaType` rather than the ones an entity graph can be
built from.

Both sides now read `ENTITY_GRAPH_EXTENSIONS`, with a test asserting the advertised
set is exactly the accepted set.

## Breaking Changes

None. The seeded values match the code defaults exactly. `SSE_ENABLED` defaults
`true`, so wiring it changes nothing until someone turns it off — which is the
point. The extensions endpoint stops returning a value that already 422'd on use;
no schema or field changes, so neither SDK tier moves.

## Testing

- `tests/routers/graphs tests/config tests/adapters/sec` — 2480 passed, 14 skipped.
- `tests/schemas tests/middleware/sse tests/models/api` — 786 passed.
- New tests: the SSE kill switch actually refuses a connection when off and admits
  one when on; the advertised extension set equals the accepted set.
- `bash -n bin/setup/aws.sh` clean; `just lint` / `format` / `typecheck` clean.
- Defaults cross-checked against the `get_parameter_value` declarations in
  `config/env.py`, not against locally resolved values, since a local `.env` can
  override them.

## Certification


jfrench9:
@claude please review this PR

PR #1248, opened 2026-08-22
refactor(connections): drop the SSE kill switch and the SEC connection

jfrench9:
Two items surfaced by the post-overhaul review, plus one the review attributed to the wrong provider.

## 1. Remove `SSE_ENABLED`

The flag read as an operational kill switch and was not one:

- It was checked inside `SSEConnectionManager.add_connection()`, which the operation stream reaches **after** the generator has already emitted a `connected` event. The resulting 503 was caught and re-emitted as an SSE error inside a `200` response, so no client ever saw the status code it would need to fall back to polling.
- Several SSE surfaces build their `EventSourceResponse` without consulting the flag at all, so those streams stayed up regardless.

SSE is a platform feature now. A switch an operator can flip during an incident while believing they have acted is worse than no switch. Connection limits, timeouts, load shedding and infrastructure controls remain the real levers. Removed from config, SSM setup, `.env.example`, CI, `pytest.ini`, and the manager.

## 2. Remove the SEC connection provider

Removed rather than repaired. It was a filing-driven entity-creation path predating the shared SEC repository, **disabled in every deployed environment**, and its sync had degenerated to returning an explanatory sentence that the dispatcher handed back as a `task_id` inside a `202 pending` envelope — an operation id with no run behind it and a poll instruction that never terminates.

SEC filings still arrive nightly into the shared SEC repository. Only the per-graph connection surface goes. `"sec"` stays in `RESERVED_SOURCE_NAMES` because the SEC pipeline still emits it as a source name.

## 3. The same defect on `external`, which the review missed

The fake-async bug was never SEC-specific. `sync_external_connection` has the identical shape and, unlike SEC, `CONNECTION_EXTERNAL_ENABLED` defaults to **true** — so deleting SEC would have removed the disabled instance and left the enabled one.

Provider sync now returns a `SyncOutcome` stating whether a run was actually dispatched:

- `dispatched` — carries the run id, keeps the `pending` envelope and the polling instruction.
- `unsupported` — completes with a null `task_id` and the provider's own message. Nothing to poll.

The outcome also decides who releases the per-connection sync lock, replacing the registry's separate `starts_async_run` declaration, which could drift from what the provider actually did.

Also fixed: the QuickBooks OAuth callback was placing the outcome object where `OAuthCallbackResponse` wants a run-id string (a 500 after successful authorization); it now extracts `task_id`.

## SDK impact

`SECConnectionConfig` and the `"sec"` provider discriminator leave the OpenAPI surface. This is **generated-tier** surface for a path that never functioned in any shipped configuration, so per the two-tier client contract it rides a client minor and belongs in the release notes — not a blocking break.

- `roboledger-app` still has SEC connection UI (`SecSetupForm.tsx`, `sec` branches in `ConnectionCard.tsx`) that this orphans. Handled separately.
- Two SSM parameters can be deleted once this deploys: `features/SSE_ENABLED` and `features/CONNECTION_SEC_ENABLED` in each environment.

## Validation

- Pre-commit gate green: ruff check, ruff format, basedpyright (0 errors, 0 warnings).
- Test suite migrated to the new contract across 12 files, including new regression coverage for the no-op path at both the REST route and the `sync-connection` MCP tool.


jfrench9:
@claude please review this PR

PR #1249, opened 2026-08-22
feat(security): record admin-surface actions in the audit trail

jfrench9:
## Summary

Every request that passes admin authentication now emits an `ADMIN_ACTION` security audit event, so the application-side trail records *what* was done on the admin surface alongside the AWS-side record of *who* opened the session.

The practical effect is retention. These records go out as `SECURITY_AUDIT:` lines, which the existing `?SECURITY_AUDIT` subscription filter (`cloudformation/audit.yaml:207`) forwards to the audit bucket at 400 days. Admin activity previously lived only in the API log group, which rolls at 30 days — shorter than a SOC 2 Type II observation window, so evidence aged out mid-window.

## Changes

**`robosystems/security/audit_logger.py`**

- New `SecurityEventType.ADMIN_ACTION`, filed with the "evidence, not alerts" cohort (the membership and SCIM lifecycle events).
- New `SecurityAuditLogger.log_admin_action()` recording the acting key id, method, path, status code, success, and redacted query string. `risk_level` is `low` for safe methods and `medium` otherwise.

**`robosystems/middleware/logging.py`**

- New branch in `SecurityLoggingMiddleware.dispatch`, alongside the existing `/v1/auth/` and 403 branches, rather than a new middleware.

Three points worth a close look:

1. **No CloudWatch metric is published.** `ADMIN_ACTION` is deliberately absent from both `_METRIC_FOR_EVENT` and `_ADMIN_FLAGGABLE_EVENTS`. The event carries `details["admin"] = True` as a log dimension, and that flag only reaches the metric path for event types in the flaggable set — so routine admin work cannot trip `FailedAdminAuthAlarm`. Two tests pin this.
2. **The branch is guarded on `request.state.admin_key_id`, not on the path alone.** That attribute is set by `AdminAuthMiddleware` only after the key verifies, so unauthenticated requests are excluded structurally. Authentication failures keep their existing route through `log_admin_auth_failure` (which alarms) and are not double-logged here. This also correctly excludes the Stripe webhook, which shares the `/admin/v1/` prefix but is signature-verified rather than admin-authenticated.
3. **The recorded path is the literal one, identifiers included** — which invoice or subscription was acted on is the substance of the record. Query strings pass through the existing `redact_sensitive_query_params` first.

**`tests/middleware/test_admin_action_audit.py`** — 12 tests across the record shape and the middleware branch.

One test builds a real ASGI app rather than mocking `Request`. The branch reads state that the route sets, *after* `call_next` returns; that works because Starlette backs `request.state` with `scope["state"]`. A mocked `Request` would assert the assumption instead of exercising it, so this is verified end-to-end against the installed Starlette.

## Breaking Changes

None. The change is purely additive (261 insertions, 0 deletions) and touches no externally visible surface — no OpenAPI shape, GraphQL schema, or operations envelope changes, so neither client SDK is affected and no regen is required.

## Testing

- `just test-code` — passed (0 errors, 0 warnings, 0 notes). Ruff reformatted the new test file; the reformatted version is what's committed and tests were re-run after.
- `uv run pytest tests/middleware/ tests/security/ tests/routers/admin/ tests/admin/` — 3243 passed, 29 skipped. Chosen to cover the modified middleware, the modified audit logger, and every consumer of the admin auth path.
- `uv run pytest tests/middleware/test_admin_action_audit.py` — 12 passed.
- Full `just test-all` was **not** run; the suites above are a targeted subset.

## Certification

### DIFF
--- a/robosystems/middleware/logging.py
+++ b/robosystems/middleware/logging.py
@@ -20,6 +20,7 @@ from robosystems.logger import (
   log_auth_event,
   security_logger,
 )
+from robosystems.security.audit_logger import SecurityAuditLogger
 from robosystems.security.request_context import bind_request_id, reset_request_id
 
 logger = api_logger
@@ -301,6 +302,22 @@ class SecurityLoggingMiddleware(BaseHTTPMiddleware):
           "method": request.method,
           "path": request.url.path,
         },
+      )
+
+    # Record every authenticated admin-surface action. `admin_key_id` is set on
+    # request.state by AdminAuthMiddleware once the key verifies, so its absence
+    # means authentication never succeeded — those are already recorded (and
+    # alarmed) by log_admin_auth_failure, and must not be double-logged here.
+    admin_key_id = getattr(request.state, "admin_key_id", None)
+    if admin_key_id and request.url.path.startswith("/admin/v1/"):
+      SecurityAuditLogger.log_admin_action(
+        admin_key_id=str(admin_key_id),
+        method=request.method,
+        endpoint=request.url.path,
+        status_code=response.status_code,
+        ip_address=client_ip,
+        user_agent=user_agent,
+        query=redact_sensitive_query_params(str(request.url.query)) or None,
       )
 
     # Log authorization failures (403 responses) using structured logging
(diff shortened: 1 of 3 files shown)


jfrench9:
@claude please review this PR

PR #1251, opened 2026-08-22
feat(roboledger): give reconciling items a disposition verb

jfrench9:
Closes gaps 1 and 3 of the July-close reach list, plus a close gate that makes them matter. Two of the three items remaining on the 1.9 line; gap 4(b) follows in its own PR.

## The problem

When a QuickBooks transaction is edited after it was synced, the loader does not overwrite approved books — it flags the event and stashes the incoming payload. That flag is how the disagreement becomes visible, and nothing anywhere cleared it. So the reconciliation worklist was append-only: items dispositioned with a customer at one close were still flagged at the next, indistinguishable from unreviewed ones. On the reference tenant, three items re-flag on every single sync.

Separately, an alignment entry that mirrors a change the source system *already has* must not travel back — publishing it applies the change twice. That lane existed, but only as a side effect of `source='system'` being absent from a tuple, discoverable by reading `qb_writeback.py`.

## What this adds

**`preview-reconciling-item`** reads the posted entries against the accepted payload and nets the difference per account. Writes nothing.

**`resolve-reconciling-item`** carries out one of three treatments and clears the flag:

| Disposition | What it does | When |
|---|---|---|
| `restate` | Regenerates the event's entries from the accepted payload, in place | Nothing external binds the affected months |
| `catch_up` | Posts the difference as a local-only alignment entry in an open period | A reporting cadence has locked those months |
| `acknowledge` | Records that an entry authored elsewhere handled it | The work was already done by hand |

The default follows the period: `restate` while every affected period is open, `catch_up` once one is closed.

**`publish_to_source`** on `journal_entry_recorded` metadata makes the local lane explicit, honoured inside the shared write-back predicate and short-circuited in `execute_event_block` ahead of the connection lookup — close hands its write-back connection to every event in the batch, so a check placed after resolving `connection_id` would publish exactly the entries the flag exists to hold back.

**A `reconciling_items` close blocker**, with `allow_reconciling_items` to override, mirroring the stranded-obligation gate. The count rides the audit note either way.

## Why clearing the flag means adopting the payload

The sync decides an event has changed by comparing its stored metadata against the incoming payload. Setting `payload_drift = False` alone would therefore last exactly until the next sync — the same defect wearing a fix's clothes. Every disposition sets the live payload to the accepted one and appends its record to `metadata.reconciliation_history`, a key the comparison excludes. The exclusion list is now one named constant shared by the loader and the resolver, so the two halves of that contract cannot drift apart; it also picked up the `dispatch_*` prefix, which would otherwise have made a previously-retried event drift forever.

## Testing

The load-bearing tests run against real Postgres, because the claim being made is about what the *sync* will do, not about what the resolver wrote — a mocked session can only confirm the latter. After each disposition the resolver's output is fed back through the loader's own comparison function.

Both that test and the write-back predicate's were mutation-checked: reverting the resolver to only clear the flag, or the predicate to ignore the flag, fails them. The JSONB predicate is covered on real SQL for all five cases (source default, flag in both directions, JSON null, already-published) — a predicate that silently matched nothing would pass every mock-based test in the suite and quietly double-apply an upstream edit.

`just test-all` green: 13718 passed, 43 skipped.

## Notes

- No migration — `payload_drift` and the metadata column already exist.
- Both operations reach REST and MCP from one registration; `preview-` is recorded as read-only in the staleness-coverage exemptions.
- SDK impact is generated-tier only (new operations), so this rides a client minor.


jfrench9:
@claude can you review

jfrench9:
Both review findings addressed in d92397f8, plus the comment on the double metadata write.

**The `Event.status` mismatch** — verified reachable and shut. `loader.py:1649` derives the payload's status from the source (`jeh_status = "posted" if _source_auto_commits_on_sync(source) else "draft"`), and `_SOURCE_AUTO_COMMITS` lists only `quickbooks` — so a payload saying `draft` needs a source that does not auto-commit, plus an event already `fulfilled`. Restate now refuses that combination rather than producing the mismatch: the handler only upgrades status to `fulfilled`, and quietly downgrading a terminal state would be the worse answer. The error names `catch_up`, which does work there.

**One more found while checking that** — the re-flag guard was comparing the stored `drift_detected_at` against a round-trip of itself. `datetime.fromisoformat` accepts `Z` and re-emits `+00:00`, so a `Z`-stamped item would have failed that comparison forever behind a misleading *"was re-flagged… Preview it again"* — as would an unparseable stamp, leaving the item impossible to resolve at all. The plan now carries the raw string and the guard compares it directly. Only the loader writes this field today (always `+00:00`), so nothing was broken in practice.

Both guards are mutation-checked — reverting either behaviour fails its test. 17 DB tests now; `just test-all` green at 13721 passed.

**The double metadata write** is correct as you read it: both calls build history from the pre-mutation `live`, so the second replaces the first rather than appending. Added the comment explaining why two writes are structurally necessary — the rebuilt entry ids cannot go on the trail before the rows they name exist.

Also worth flagging, since it falls outside this diff: the app renders unknown blocker codes as the raw string (`PeriodClosePanel.tsx:845`), so this gate would have shown the literal token `reconciling_items` with no next step — and the reference tenant has eleven unresolved items right now, so that is what the first close after deploy would have looked like. Copy added in [roboledger-app#317](https://github.com/RoboFinSystems/roboledger-app/pull/317). The count/sample detail needs the camelCase mapping in the TS client facade, so it follows separately.

PR #1252, opened 2026-08-22
feat(roboledger): run the MCP close on the worker

jfrench9:
Closes gap 4(b) — the last of the three items on the 1.9 line. With #1251 merged, this completes the July-close reach list.

## The problem

A real month's close takes ~30s, most of it one QuickBooks round-trip per draft, against a 25s server-side tool budget. It always *completed*; the operator just stopped hearing about it, got an error, and then had to know not to retry a write that had already landed. Since PR #1233 persisted the receipt, nothing is lost when that happens — which is what makes this a legibility fix rather than a durability one.

## What changes

`close-period` dispatches to the worker and waits inside the budget:

- **Lands in time** → the receipt, exactly as before.
- **Doesn't** → `status: "in_progress"` with the operation id and where to look (`get-period-close-status` for the receipt, `get-fiscal-calendar` for `has_close_receipt`), and the instruction never to call `close-period` again — the close is atomic and already running.
- `worker_started` separates *running now* from *queued behind a busy worker*. Different things to tell someone who is waiting.

**REST is untouched.** The close panel and both SDK facades dereference the synchronous envelope's `result`, and the timeout they'd be insulated from isn't one they have.

## Three things that fell out of the move

**A refused close is a result, not a failure.** The consumer reduces any raised exception to a plain string, so a gate rejection that propagated would arrive as text with its blockers stripped — the one part an operator can act on. Outcome shaping moved to `close_outcomes.py`, shared by the task and the tool so both describe the same close identically; the task returns domain outcomes and lets genuine faults raise. It also picked up the `WritebackFailed` case the tool was previously dropping into a generic `except Exception`, losing `failed_events`.

**`get-period-close-status` never put `close_receipt` on the wire**, though the ops read returned it and the playbook told operators to reconstruct the receipt from there. The instruction pointed at nothing. Now it points at the receipt — which matters more here, since it is the poll target for an in-progress close.

**The tool needs a real user id.** An operation is readable at `/v1/operations/{id}/status` only by the user it was created for, so the `mcp:{graph_id}` sentinel the synchronous path could fall back on would have enqueued work nobody could look up. No user now means no dispatch.

## A defect the tests caught

The poll loop originally accumulated `waited += POLL_INTERVAL_S`. With a zero interval the accumulator never advances and it spins forever — which is exactly what the handback tests did. It now measures against the monotonic clock, which is the honest reading of "how long the caller has waited" regardless.

## Budget

`period_close` gets 600s, matching `operator`, pinned by a test against the constants that bound the run (the QuickBooks retry ladder, the statement ceiling, the fence's lock wait). The sizing is decided by the failure mode: `asyncio.wait_for` cancels the coroutine but cannot cancel the thread doing the publish, so an under-budget close is marked FAILED while its thread runs on and can still commit. Reporting a landed close as a failure is worse than waiting longer for a slow one — and the FAILED message says so.

## Testing

- `test_period_close.py` — the task: receipts, every domain refusal returned rather than raised, faults still raising, `OPERATION_STARTED` emitted first, and the already-closed case handing back the receipt (the stalled-task reaper measures age from enqueue, so it can re-queue a close that already landed).
- `test_close_outcomes.py` — the shapes, including a test that every type in `CLOSE_DOMAIN_ERRORS` actually has shaping, since one listed without it would be swallowed into an empty result.
- `test_fiscal_calendar_tools.py` — rewritten for dispatch and handback.
- Verified in the running worker container that `period_close` registers and resolves to `PeriodCloseTask`.

`just test-all` green: 13747 passed, 43 skipped.

### DIFF
--- a/robosystems/middleware/mcp/tools/fiscal_calendar_tools.py
+++ b/robosystems/middleware/mcp/tools/fiscal_calendar_tools.py
@@ -363,24 +363,39 @@ The receipt:
         "message": "User context required to close a period.",
       }
 
-    dispatch = await enqueue_task(
-      task_type="period_close",
-      graph_id=graph_id,
-      user_id=str(user_id),
-      params={
+    try:
+      dispatch = await enqueue_task(
+        task_type="period_close",
+        graph_id=graph_id,
+        user_id=str(user_id),
+        params={
+          "period": period,
+          "actor_id": str(user_id),
+          "actor_type": "agent",
+          "allow_stale_sync": bool(arguments.get("allow_stale_sync", False)),
+          "allow_stranded_obligations": bool(
+            arguments.get("allow_stranded_obligations", False)
+          ),
+          "allow_reconciling_items": bool(
+            arguments.get("allow_reconciling_items", False)
+          ),
+          "note": arguments.get("note"),
+        },
+      )
+    except Exception as exc:
+      # Nothing was queued, which makes this the one failure here that is
+      # safe to retry — and worth saying so, because every other message
+      # this tool returns says the opposite.
+      logger.warning("close-period could not dispatch: %s", exc, exc_info=True)
+      return {
+        "error": "dispatch_failed",
         "period": period,
-        "actor_id": str(user_id),
-        "actor_type": "agent",
-        "allow_stale_sync": bool(arguments.get("allow_stale_sync", False)),
-        "allow_stranded_obligations": bool(
-          arguments.get("allow_stranded_obligations", False)
+        "message": (
+          "The close could not be handed to the worker, so it has NOT "
+          "started and nothing was written. Retry; if it keeps failing, "
+          "the task queue is unavailable."
         ),
-        "allow_reconciling_items": bool(
-          arguments.get("allow_reconciling_items", False)
-        ),
-        "note": arguments.get("note"),
-      },
-    )
+      }
     operation_id = dispatch["operation_id"]
 
     storage = get_event_storage()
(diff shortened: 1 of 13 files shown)


jfrench9:
@claude can you review?

jfrench9:
Fixed in b0744b0e.

**The error boundary** — and the gap was wider than `_gate_check`. There was no `try`/`except` anywhere in `execute`, and the two calls the async path now depends on both talk to the queue: `enqueue_task` and `get_operation_metadata`. A degraded queue was the one path where this tool's answer degraded to a sanitized string from `manager.py`'s outer handler — on a tool whose entire contract is now *"here is what happened and where to look next"*.

They needed opposite advice, which is why it is two guards rather than one:

- **Dispatch failed** → nothing queued, nothing written. This is the single case here where retrying is right, and worth saying plainly, because every other message this tool returns says not to. Returns `dispatch_failed` with no `operation_id` — there is no operation to look up.
- **Poll failed** → the close is already running. Losing sight of it is a reporting problem, not a reason to do it again, so it falls through to the in-progress handback that sends the operator to the period itself.

Both mutation-checked: making the dispatch failure claim the close is running, or letting the poll error propagate, each fails its test.

**The `assert`** is now a `ValueError` — `-O` strips asserts, and the failure it guards would otherwise surface as a confusing `None` inside a session call.

**Left as-is: the stale-metadata fallthrough.** Distinguishing "no metadata" from PENDING would need a third branch to say "I can't tell", and the handback already sends the operator to `get-period-close-status`, which answers the question authoritatively either way.

**One thing outside this diff, now recorded rather than dropped:** `backfill-plan-history` still runs synchronously on the same 25s budget, and recloses up to 12 months per call (24 max), each a full close with its own QuickBooks publish. The argument that justified moving `close-period` applies there with far more force, and the two are now inconsistent — one dispatches, its neighbour doesn't. It's also the tool the orphaned-arc re-stamp row tells an operator to run chunked, which is exactly the long-running case. Filed as a *"no spec"* row in the ledger domain README with the un-park trigger; the machinery this PR adds makes it about half a day.

`just test-all` green: 13749 passed, 43 skipped.

PR #1254, opened 2026-08-24
docs(identity): document enterprise SSO, SCIM, and passkeys

jfrench9:
## Summary

Enterprise SSO (OIDC), SCIM provisioning, and passkeys were documented in `SECURITY.md` and nowhere else — absent from the README, the public OpenAPI landing description, the architecture tree, and the API/operations model layout tables. This PR is the docs-only half of closing that gap; the companion wiki changes (a new **Enterprise SSO & SCIM** page, plus passkeys/MFA coverage on the Authentication page) are already published.

Documentation only. No source, schema, or behavior changes.

## Changes

**`README.md`**
- Short **Identity & Access** subsection under Architecture: login method is a deployment decision published at `GET /v1/auth/providers`, and enterprise SSO/SCIM ship in the repository off by default with no license gate.
- States the link-only provisioning rule, since that is the design decision integrators most often assume the opposite of. Control-level detail (password policy, passkey enrollment mechanics) is deliberately left to `SECURITY.md` and the wiki rather than restated here.
- Adds the new wiki page to the Operations Layer link list.

**`static/description.md`** — *worth a closer look; this renders on the public API landing page.*
- Replaces `**Authentication**: JWT tokens and API key management`, which predated passkeys and enterprise identity entirely.
- Adds an **Enterprise identity** bullet marked off-by-default and scoped to dedicated/self-hosted deployments, so nobody reads it as available on the managed platform.

**`robosystems/models/api/README.md`**
- Adds the missing `passkeys.py` and `scim.py` rows.
- Corrects the `auth.py` row: it said "SSO flows", which reads as IdP federation but describes the internal cross-app bridge.

**`robosystems/operations/README.md`**
- Adds the missing `passkeys.py` and `locking.py` rows.

**`CLAUDE.md`**
- The architecture tree's `auth/` line named only "SSO"; now names all four mechanisms.

**Accuracy notes for the reviewer**
- Okta is described as validated; Entra deliberately is not. The deactivation path tolerates Entra's non-RFC payload shapes, but it has never been run against one.
- The supported surface is stated as users-only with a single filter — no Groups, bulk, sort, ETag, or password change — matching what `ServiceProviderConfig` declares.
- Nothing here discloses a vulnerability or a control that `SECURITY.md` did not already describe publicly, so there is no deploy-sequencing concern.

One item from the same finding was **not** actioned on purpose: seeding placeholder SSO keys into `bin/setup/aws.sh`. Boot validation refuses to start on `SSO_OIDC_ENABLED=true` with an unconfigured connection by testing those keys for truthiness, so placeholders would satisfy that check with garbage instead of failing loudly. The keys are documented in the Bootstrap Guide with that reasoning instead.

## Breaking Changes

None. No API surface, GraphQL schema, operations envelope, or REST shape changed, so there is no SDK impact in either the stable or generated tier.

## Testing

`just test-all` — **not run.** No Python changed; the diff is five markdown files.

The code-quality gate ran via the pre-commit hook on both commits (ruff, format, basedpyright): `All checks passed`, `1805 files already formatted`, `0 errors, 0 warnings, 0 notes`.

Wiki links and internal anchors on the changed pages were checked programmatically and resolve.

## Certification

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -266,7 +266,7 @@ robosystems/
 │   ├── operators/                # AI Operator operations (Claude/MCP executors — distinct from REA Agent counterparty model)
 │   └── providers/                # Provider registry and implementations
 ├── middleware/                   # Cross-cutting concerns
-│   ├── auth/                     # Authentication (JWT, API keys, SSO)
+│   ├── auth/                     # Authentication (JWT, API keys, cross-app SSO, passkeys/MFA, enterprise OIDC + SCIM)
 │   ├── billing/                  # Credit consumption tracking
 │   ├── graph/                    # Graph routing and multi-tenancy
 │   ├── rate_limits/              # Burst protection
--- a/README.md
+++ b/README.md
@@ -203,13 +203,9 @@ Because tenancy is enforced at the graph rather than in application predicates, 
 
 ### Identity & Access
 
-Credentials are `X-API-Key` for programmatic access and short-lived JWTs for the browser apps, both resolving to the same user and both revocable — deactivating a user bumps their session version *and* revokes their API keys, so there is no credential family that survives an offboarding.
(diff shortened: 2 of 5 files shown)


jfrench9:
@claude please review this PR

PR #1255, opened 2026-08-25
fix(roboledger): keep refused worker closes uncommitted

jfrench9:
## Summary

Two correctness fixes from the GTM readiness pass. A refused worker close no longer persists as a closed period, and session refresh plus the public password-strength check now share the same token-purpose and abuse bounds as the rest of login.

## Changes

**Worker close**
- Domain refusals on `period_close` roll back the extensions session before returning a result, so a failed statement stamp cannot commit a closed period.
- Tests now assert rollback on every refusal, including stamp failure, and that a successful close does not roll back.

**Auth**
- Session-token classification is one predicate, used by bearer verification and the refresh grace path.
- The public password-strength check uses the same rate limiter as login and rejects oversized input.
- Tests cover purpose-scoped tokens on refresh, the limiter mount, and the input bound.

## Breaking Changes

None. The password-strength request model now rejects passwords over 128 characters (the same ceiling registration already enforces). Additive otherwise; no SDK facade change.

## Testing

- `uv run pytest` on the close-task, JWT type, session refresh, password-check, auth-model, and MFA-token modules: 95 passed.
- Pre-commit ruff check/format: passed.

## Certification

### DIFF
--- a/robosystems/middleware/auth/jwt.py
+++ b/robosystems/middleware/auth/jwt.py
@@ -152,6 +152,21 @@ def _get_user_session_version(user_id: str, session: Any = None) -> int | None:
     sess.close()
 
 
+def is_session_access_token(payload: dict[str, Any]) -> bool:
+  """Whether this JWT is a session bearer, not a purpose-scoped token.
+
+  Session tokens carry ``type: "access"``. Tokens with no ``type`` at all are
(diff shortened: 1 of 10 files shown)


jfrench9:
@claude please review this PR

PR #1256, opened 2026-08-26
fix(sec): align the public SEC MCP copy with the subscription gate

jfrench9:
## Summary

The public SEC MCP surface described access it does not grant. The Starter plan's feature list said "Unlimited MCP tool access" while the same manifest rate-limits MCP calls per plan, and the Swagger "Connect via MCP" section showed the `sec` connector URL with an API-key header and no mention that the public repository requires an active repository subscription — so a stranger following the snippet reaches a 403 with no explanation. This fixes the copy at its one writer and guards the manifest so a rate-limited plan cannot describe itself as unlimited again.

## Changes

**SEC manifest** (`robosystems/adapters/sec/manifest.py`)
- Starter plan feature "Unlimited MCP tool access" → "MCP access from Claude, Cursor, or any MCP client". The per-plan MCP rate limits are unchanged; the bullet now describes what the plan is rather than a quantity the manifest contradicts. This string is served verbatim by `GET /v1/offering`.

**API description** (`static/description.md`)
- The "Connect via MCP" section now states that access follows the graph: own graphs need graph membership, and the public `sec` repository needs an active repository subscription (Starter or Advanced), with plans and MCP limits published at `GET /v1/offering`. It says explicitly that an API key alone does not open `sec` and that the server answers 403 until the subscription exists.

**Guard test** (`tests/config/test_shared_repositories.py`)
- `TestPublishedPlanClaimsMatchEnforcedLimits`: for every shared-repository manifest, any plan carrying a per-window rate limit must not advertise itself as "unlimited". Verified to fail against the previous copy and pass against this one; a second test pins that the guard actually sees the `sec` plans.

Not in this PR: `robosystems-app`'s pricing page restates the same Starter bullet in its own markup — that is a separate app-side change.

## Breaking Changes

None. The changed manifest string reaches clients only as the value of an existing `features: string[]` field on `GET /v1/offering`; no schema, endpoint, or SDK facade changes. No SDK regen needed.

## Testing

- `uv run pytest tests/config/test_shared_repositories.py tests/routers/test_offering.py`: 31 passed.
- Mutation check: the new guard fails against the pre-change manifest (stashed and re-run), passes after.

## Certification

### DIFF
--- a/robosystems/adapters/sec/manifest.py
+++ b/robosystems/adapters/sec/manifest.py
@@ -90,7 +90,7 @@ SEC_MANIFEST = SharedRepositoryManifest(
       "access_level": "READ",
       "description": "Full SEC data access for individuals",
       "features": [
-        "Unlimited MCP tool access",
+        "MCP access from Claude, Cursor, or any MCP client",
         "API access",
         "Standard rate limits",
         "1 backup download per month",
--- a/static/description.md
+++ b/static/description.md
@@ -107,7 +107,7 @@ Built on the blocks:
 
 ## Connect via MCP
 
-Every graph is an MCP server, and the graph's URL is the preferred way to connect — Claude, Claude Code, Cursor, or any MCP client that supports HTTP transports, no install required. The URL picks the graph (`sec` for the public SEC repository, your graph id for your own); your API key goes in the `X-API-Key` header:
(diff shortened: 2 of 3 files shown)


jfrench9:
@claude please review this PR

PR #1257, opened 2026-08-26
fix(roboledger): forward allow_reconciling_items on the REST close

jfrench9:
## Summary

The shared `ClosePeriodRequest` model declares `allow_reconciling_items`, the MCP `close-period` tool and the worker task forward it, and the hand-written REST closer did not — so a human closing through `POST /extensions/roboledger/{graph_id}/operations/close-period` could send the documented flag and have it ignored, while an agent on the MCP surface could override the reconciling-item gate. This forwards the field on REST and pins, structurally, that every `allow_*` override the request model declares reaches the kernel. Records §3 of `local/RoboSystems/specs/ledger/async-close-integrity.md`.

## Changes

**REST closer** (`robosystems/routers/extensions/roboledger/operations.py`)
- `close_period_op`'s runner now passes `allow_reconciling_items=body.allow_reconciling_items` to `cmd_close_period`, alongside the `allow_stale_sync` and `allow_stranded_obligations` overrides it already forwarded. One line; the kernel signature already accepted it.

**Test** (`tests/routers/extensions/roboledger/test_operations.py`)
- `TestClosePeriodOp.test_forwards_every_close_override_to_the_kernel`: calls the REST op with every override set, patches `cmd_close_period`, and asserts that the set of `allow_*` fields on `ClosePeriodRequest` is a subset of the kwargs the kernel received (and that each arrives `True`, with `note` and `actor_id` intact). The assertion is derived from the model's fields, so a future override added to the model without being forwarded fails here. Verified to fail against the pre-fix router and pass after.
- Adds `close_period_op` / `ClosePeriodOperation` to the module's imports and a small `ClosePeriodResponse` builder.

## Breaking Changes

None. The request model, OpenAPI shape, and SDK surface are unchanged — the field already existed on the wire; it is now honoured.

## Testing

- `uv run pytest tests/routers/extensions/roboledger/test_operations.py`: 47 passed (46 existing + the new case).
- Mutation check: the new test fails with the one-line router change stashed (`REST closer drops close overrides the model declares: {'allow_reconciling_items'}`), passes with it.

## Certification

### DIFF
--- a/robosystems/routers/extensions/roboledger/operations.py
+++ b/robosystems/routers/extensions/roboledger/operations.py
@@ -2104,6 +2104,7 @@ async def close_period_op(
           service=_fiscal_svc,
           close_service=_close_svc,
           allow_stranded_obligations=body.allow_stranded_obligations,
+          allow_reconciling_items=body.allow_reconciling_items,
         )
     except CloseGateFailed as e:
       if e.no_calendar:
--- a/tests/routers/extensions/roboledger/test_operations.py
+++ b/tests/routers/extensions/roboledger/test_operations.py
@@ -24,6 +24,11 @@ from robosystems.models.api.extensions.entity import (
   LedgerEntityResponse,
   UpdateEntityRequest,
 )
+from robosystems.models.api.extensions.fiscal_calendar import (
+  ClosePeriodRequest,
+  ClosePeriodResponse,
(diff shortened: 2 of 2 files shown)


jfrench9:
@claude please review this PR

PR #1258, opened 2026-08-26
fix(billing): keep money and delivery together on the self-serve doors

jfrench9:
## Summary

Five places on the self-serve billing path where money and delivery could come apart — the class the archived `paid-path-hardening` closed the first four instances of, found again on the doors a stranger with a card actually walks through (checkout retry, graph delete, card-on-file subscribe, renewal). Each is closed at its chokepoint with a test that fails against the previous code. Records `local/RoboSystems/specs/commerce/self-serve-billing-integrity.md` §1–§5.

**Deploy note — read before shipping the release that carries this.** The migration adds a partial unique index over live `billing_subscriptions` rows. If duplicate live `(resource_type, resource_id, user_id)` rows already exist in an environment, the index build fails and, per the fail-closed boot, the Dagster daemon will not start. Run the check in the migration's comment against each environment first and resolve any row it returns by hand.

## Changes

**Checkout retry retires the previous session** (`routers/billing/checkout.py`, `operations/providers/payment_provider.py`)
- Starting a new checkout used to mark the org's earlier `pending_payment` row `canceled` and leave its hosted Stripe session payable for 24 h; paying that URL bound a live subscription to a terminal row that provisioning refuses and the org listing hides. Now the earlier session is expired at the provider first (`expire_checkout_session`, new on the provider interface). If it turns out to have been paid already, the new checkout is refused with 409 and the earlier row is left to provision. Retired rows now carry `canceled_at`/`ends_at` like other terminal rows.
- `_handle_checkout_completed` (`dagster/jobs/billing.py`): a paid session that arrives for a row already terminal *and* still unprovisioned reopens the row to `pending_payment` (audited as `subscription_resumed`, previous status kept in metadata) so the provisioning claim can take it. Money moved, so it becomes a resource. Rows that already own a resource are untouched.

**Delete-graph fails closed on provider failure** (`routers/graphs/operations.py`)
- The Stripe cancel was best-effort: on failure it logged, recorded a low-risk audit event, and cancelled locally anyway — ending access while the charges continued, with nothing to replay the cancel. It now raises 502 with local state unchanged, the same contract as the repository cancel path (`ProviderCancellationError`). The audit event stays, at medium risk, with `local_state: unchanged`.

**Card-on-file subscribe refuses an incomplete first invoice** (`operations/providers/payment_provider.py`)
- `Subscription.create` now passes `payment_behavior="error_if_incomplete"` and, as a backstop, raises `PaymentIncompleteError` (a `ValueError`, so both existing callers already translate it into a `failed` row) if the returned status is not `active`/`trialing`, cancelling the stray provider object first. Previously the provider returned an id regardless of status and both callers activated locally on return.

**Failed renewal moves `active` → `past_due`** (`dagster/jobs/billing.py`)
- `invoice.payment_failed` only ever touched `pending_payment` rows; an active subscription's failed renewal was audit-logged and otherwise left `active`, with the access drop depending entirely on `customer.subscription.updated` arriving. It now moves the row to `past_due` (which the access gate already treats as blocked) and invalidates the access cache. `updated` remains the authority for the way back to `active`.

**One live subscription per (resource, subscriber)** (`models/core/billing/subscription.py`, migration `8c8e469f4a0a`, `routers/graphs/subscriptions.py`)
- New partial unique index `uq_billing_sub_live_user_resource` on `(resource_type, resource_id, user_id) WHERE status NOT IN ('canceled','failed') AND resource_id IS NOT NULL AND user_id IS NOT NULL`. Graph rows have no `user_id` and checkout rows have no `resource_id` until provisioning, so neither is constrained. The repository subscribe handler's check-then-insert raced; it now catches the `IntegrityError` at the INSERT — which happens before any provider call — rolls back, and answers the same 409 the pre-check gives.
- Migration autogenerated (`just migrate-create`), formatted to the repo convention, with the pre-deploy duplicate query in its comment.

## Breaking Changes

None. No request/response model changes; no SDK facade or generated-tier changes. Two endpoints gain a new failure status a client may now see (`POST /v1/billing/checkout` → 409 when the previous checkout was already paid; `delete-graph` → 502 when the provider cancel fails), both on paths that previously proceeded incorrectly.

## Testing

- `uv run pytest` over the affected modules (`tests/operations/providers/test_payment_provider.py`, `tests/routers/billing/test_checkout.py`, `tests/dagster/test_billing_webhook_handlers.py`, `tests/routers/graphs/test_delete_graph_op.py`, `tests/routers/graphs/test_subscriptions_router.py`, `tests/models/core/billing/test_subscription.py`, `tests/models/core/billing/test_provisioning_claim.py`, `tests/operations/graph/test_subscription_service.py`): **251 passed**. Three pre-existing tests asserted the previous behavior (provider mocks with no subscription status; "active stays active on failed renewal") and were rewritten to assert the new one.
- New tests: checkout retry expires-then-cancels and is refused when the previous session was paid; webhook reclaim of a retired row; delete-graph 502 with no local mutation; provider `error_if_incomplete` + incomplete refusal + `expire_checkout_session` (expired / already-complete / already-expired / still-open); router 409 on `IntegrityError`; and a real-Postgres model test proving the partial index refuses a second live row and permits one after cancellation.
- Mutation checks: each code file reverted in turn — the corresponding new tests fail (checkout 2/2, delete-graph 1/1, router 1/1, webhook 2/2).

## Certification


jfrench9:
@claude please review this PR

jfrench9:
@claude both findings addressed in 544bc4be: (1) a `pending_payment` row whose `provider_subscription_id` is no longer a `cs_` id is treated as paid — 409, no expire, row untouched; (2) the retire loop now skips rows whose `resource_config.repository_name` differs from the requested repository, so one repository's checkout cannot expire or be mistaken for another's. Two tests cover the new cases (both fail against the previous loop). Please re-review.

PR #1259, opened 2026-08-26
fix(deprovision): revoke provider grants at teardown; make delete-subgraph's backup real

jfrench9:
## Summary

Two teardown paths said one thing and did another. Graph deprovisioning deleted our copy of a QuickBooks connection's credentials and never ended the authorization at the provider, so a departed customer's grant stayed live at Intuit until it expired on its own — while the user-initiated disconnect has always revoked first. And `delete-subgraph` accepted `backup_first` (default `true`), took no backup, and still reported `backup_created` in its security-audit event and MCP result. Both now do what they claim. Records §7c items 1 and 2 of `local/RoboSystems/specs/security/data-deletion-and-retention.md`; item 3 (the `delete_database` 404 path) is not in this PR.

## Changes

**Teardown revokes provider grants** (`operations/graph/deprovision_service.py`)
- New step 5b `_revoke_provider_grants`, placed before `_clean_pg_records` because the provider cleanup reads the stored token it revokes. It calls the same `provider_registry.cleanup_connection` hook the disconnect endpoint uses, once per connection on the graph. Best-effort per connection: a provider disabled by flag is skipped (same as the endpoint), any other failure is recorded in `result.errors` (teardown finishes as `partial`) and the credential rows are still deleted — a stranded grant is an operator follow-up, not a reason to keep the tenant's data. `DeprovisionResult.connections_revoked` added and included in the teardown log line.

**`delete-subgraph` takes the backup it advertises** (`operations/graph/subgraph_service.py`, `routers/graphs/operations.py`, `middleware/mcp/tools/graph_tools.py`, `models/api/graphs/operations.py`)
- `SubgraphService.delete_subgraph_database(create_backup=True)` now runs a full-dump `BackupJob` for the subgraph through `BackupManager` (the same path the nightly job uses for subgraphs) after the data/`force` check and before `delete_database`. If the backup or its registration fails it raises `GraphAllocationError` and nothing is deleted. The result dict carries `backup_created`, `backup_location`, `backup_id`.
- The backup row is registered on the **parent** graph (`graph_id=parent`, `database_name=<subgraph db>`, initiator `final`, retention = the parent tier's hosting window), because the subgraph's own `Graph` row is hard-deleted right after and a row keyed on it would be unreachable through the listing and download surfaces. `GraphBackup.graph_id` carries no FK, so the row survives the subgraph's deletion.
- Domain `GraphAllocationError`s from the delete path (data without `force`, a failed pre-delete backup) now pass through with their own message instead of being re-wrapped with a generic prefix.
- REST handler: the audit event's `backup_created` and the operation result report what the service did, not what was asked; `backup_id` added to both. Request-model description rewritten to describe the real behavior; the default stays `true`.
- MCP tool: description rewritten (it previously told operators to run `create-backup` on the parent, which does not cover a subgraph's database); the result reports `backup_created` / `backup_location` / `backup_id` from the service.

**Shared row mapping** (`models/core/graph/graph_backup.py`)
- New `GraphBackup.from_completed_export(...)` builds — without persisting — a `COMPLETED` row from a manager export. The teardown's `_register_final_backup` now delegates to it (still under its SAVEPOINT, still uncommitted), and the subgraph path adds/commits it in its own session. One mapping, two callers.

## Breaking Changes

None. `DeleteSubgraphOp` is unchanged in shape and default; its description text changed. The `delete-subgraph` operation result gains `backup_created` and `backup_id` alongside the existing `backup_location` (additive; the envelope's `result` is untyped). The MCP `delete-subgraph` result gains `backup_location` and `backup_id`. No SDK facade changes; an SDK regen would only pick up the description text.

## Testing

- `uv run pytest` over `tests/operations/graph/test_deprovision_service.py`, `tests/operations/graph/test_subgraph_service.py`, `tests/middleware/mcp/tools/test_graph_tools.py`, `tests/routers/graphs/test_ops_shared_repo_gates.py`, `tests/models/core/graph/test_graph_backup.py`, `tests/routers/graphs/test_delete_graph_op.py`: **168 passed**. The one pre-existing test that asserted the echoed flag (`TestDeleteSubgraphExecute.test_happy_path`) was rewritten to assert the service's report.
- New tests: revoke runs while the credential row still exists (asserted from inside the patched hook), revoke failure → `partial` with the credentials still deleted, disabled provider → skipped; backup runs before delete (call order asserted) and is registered on the parent with initiator `final`; backup failure → `GraphAllocationError`, no delete, no row; `create_backup=False` → reported as not taken; MCP result reflects the service; `from_completed_export` field mapping including the cross-graph `database_name` note.
- Mutation checks: reverting `deprovision_service.py` fails the three revoke tests; reverting `subgraph_service.py` fails the three backup tests; reverting `graph_tools.py` fails the two MCP result tests.

## Certification

### DIFF
--- a/robosystems/middleware/mcp/tools/graph_tools.py
+++ b/robosystems/middleware/mcp/tools/graph_tools.py
@@ -321,15 +321,18 @@ class DeleteSubgraphTool:
         "- To permanently remove a subgraph that is no longer needed\n"
         "- Requires admin role on the parent, and cannot target the parent "
         "graph itself\n\n"
-        "**RETURNS:** Confirmation of the deletion, echoing back the "
-        "`backup_first` flag. No backup is taken — see NOTES.\n\n"
+        "**RETURNS:** Confirmation of the deletion with `backup_created`, and "
+        "when a backup was taken its `backup_location` and `backup_id`.\n\n"
         "**PARAMETERS:**\n"
         "- subgraph_id: Full id like `{parent_id}_{name}`\n"
         "- force: Delete even if the subgraph contains data (default false)\n"
-        "- backup_first: Accepted but NOT implemented; it takes no backup\n\n"
-        "**NOTES:** This is irreversible. `backup_first` does not protect the "
-        "data and never has — to keep a copy, run `create-backup` on the "
-        "parent graph first and confirm it succeeded."
+        "- backup_first: Take a full backup of the subgraph first (default "
+        "true). It is registered on the parent graph's backup list, so it "
+        "stays listable and downloadable after the subgraph is gone. If the "
+        "backup fails, nothing is deleted.\n\n"
+        "**NOTES:** The deletion itself is irreversible; only a backup taken "
+        "here (or earlier via `create-backup` on the subgraph id) preserves "
+        "the data."
       ),
       "inputSchema": {
         "type": "object",
@@ -400,7 +403,7 @@ class DeleteSubgraphTool:
 
       service = SubgraphService()
       try:
-        await service.delete_subgraph_database(
+        deletion_result = await service.delete_subgraph_database(
           subgraph_id=subgraph_id,
           force=force,
           create_backup=backup_first,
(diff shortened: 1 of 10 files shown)


jfrench9:
@claude please review this PR

jfrench9:
@claude addressed in 1b3622a6: `_delete_subgraphs` now awaits `_revoke_provider_grants(subgraph.graph_id, …)` before its `_clean_pg_records`, mirroring the parent ordering; new test seeds a QuickBooks connection on both the parent and a subgraph and asserts both are revoked while their credential rows still exist (`connections_revoked == 2`). The `entity_id` nit is taken too. Please re-review.

PR #1260, opened 2026-08-26
fix(worker): let an over-budget close land; fence waits as a background job

jfrench9:
## Summary

The worker's task budget (`asyncio.wait_for`) cancels the handler coroutine but cannot cancel the thread a period close runs on, so an over-budget close was recorded FAILED while its thread ran on and could still commit — and the consumer's `finally` dropped scale-in protection and moved to the next tenant while it did. Separately, the worker close took the period fence with the 3 s *request* wait, so a duplicate dispatch overlapping a running close of the same period was refused (`row_locked`) rather than waiting for the receipt. This closes §2 and §4 of `local/RoboSystems/specs/ledger/async-close-integrity.md`; §1 and §3 shipped in #1255 / #1257.

## Changes

**`robosystems/worker/tasks/base.py`** — new `BaseTask.run_blocking(func, *args)`: runs sync work in a thread, shielded from the budget's cancel. When the budget expires it waits up to one more budget (the *grace*) for the thread; if the thread lands, its outcome becomes the task's outcome (the stdlib `wait_for` contract: a task that absorbs its cancel and returns a value returns that value) and the overrun is logged. Only past the grace is the work abandoned — tracked in `abandoned_work`, with its late outcome logged rather than dropped — and the `TimeoutError` let through. New `budget_seconds` property reads `TASK_TIMEOUTS` by the handler's own `task_type`.

**`robosystems/worker/tasks/__init__.py`** — `register_task` stamps `cls.task_type`, so a handler can size its own waits from its budget.

**`robosystems/worker/consumer.py`** — the timeout branch reports `still_running` (in the stored error text and `error_details`) when the handler has abandoned work, so "timed out" cannot read as "did not happen". `finally` ordering is unchanged; comments record why that is safe (`Engine.dispose()` closes checked-in connections only — verified against SQLAlchemy 2.0.50, which corrects the spec's claim that the fence connection was killed mid-flight).

**`robosystems/operations/locking.py`** — `exclusive_period_fence(..., wait_ms=None)`: the wait is the caller's. Default stays the request wait.

**`robosystems/operations/roboledger/commands/fiscal_calendar.py`** — `close_period(..., fence_wait_ms=None)` forwards to the fence. REST and the synchronous MCP paths pass nothing.

**`robosystems/operations/roboledger/tasks/period_close.py`** — uses `run_blocking`; passes `fence_wait_ms = budget_seconds * 1000` (background jobs wait).

**`robosystems/worker/constants.py`, `task_protection.py`** — comments: the `period_close` budget now does three jobs (budget, grace, fence wait); the protection lease must cover budget + grace.

Reviewers: the interesting file is `base.py`. The abandon-after-grace bound is a deliberate judgment — the spec asked for "do not dispose until the thread joins", but an unbounded join lets one hung QuickBooks call (the Intuit SDK sets no per-request timeout) wedge the single prod worker indefinitely; a thread abandoned 2× over budget loses scale-in protection, which is logged as the accepted residue.

## Breaking Changes

None. `close_period` and `exclusive_period_fence` gain optional keyword parameters with unchanged defaults; no SDK-facing surface changes (neither tier).

## Testing

- `uv run pytest tests/worker tests/operations/test_locking.py tests/operations/roboledger/tasks/test_period_close.py tests/operations/roboledger/commands/test_fiscal_calendar.py` — 107 passed.
- Mutation-checked: with the grace wait removed, all five new `run_blocking` / consumer tests fail (`wait_for` raises `TimeoutError`); with the task's `fence_wait_ms` dropped, `test_the_worker_close_waits_its_whole_budget_for_the_fence` fails; with the fence ignoring `wait_ms`, `test_a_background_closer_waits_as_long_as_it_says` fails.
- `just test-code` — ruff, format, basedpyright (0 errors), cf-lint all pass.

## Certification

### DIFF
--- a/robosystems/operations/locking.py
+++ b/robosystems/operations/locking.py
@@ -191,7 +191,9 @@ def acquire_shared_period_fence(
 
 
 @contextmanager
-def exclusive_period_fence(graph_id: str, period: str, *, detail: str):
+def exclusive_period_fence(
+  graph_id: str, period: str, *, detail: str, wait_ms: int | None = None
+):
   """Hold a session-scoped exclusive fence on ``(graph_id, period)``.
 
   Uses a dedicated connection, not the caller's ``Session``. The close
@@ -199,6 +201,14 @@ def exclusive_period_fence(graph_id: str, period: str, *, detail: str):
   would return the session's connection to the pool and drop any lock
   taken on it. A session-scoped advisory lock on a connection we hold
   ourselves survives that commit.
+
+  ``wait_ms`` bounds the wait for whoever holds the fence. The default is
+  the request wait, per the module rule: a request handler that sat out a
+  whole close would pin a pooled connection for it. A background closer —
+  the worker close — passes its own budget instead. What it is most likely
+  waiting on is another close of the same period, and waiting turns a
+  duplicate dispatch into "already closed, here is the receipt" rather
+  than a refusal the operator is told to retry.
 
   Unlock (or invalidate the connection, so the pool discards it) before
   returning the connection — a pooled connection still holding this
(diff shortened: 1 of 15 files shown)


jfrench9:
@claude please review this PR

PR #1261, opened 2026-08-26
fix(graph-api): dispose side stores on delete when the .lbug is already gone

jfrench9:
## Summary

`LadybugDatabaseManager.delete_database` raised 404 before any cleanup when `{graph_id}.lbug` was missing. The graph file is the first thing a delete removes, so a teardown that died between it and the side stores retried into exactly that 404 — which deprovision counts as "already absent, converge" — and the registry slot was freed with the Lance directory, the DuckDB staging, and any `-wip`/`-prev` temporary still on the volume for the next tenant to inherit. This closes §7c.3 of `local/RoboSystems/specs/security/data-deletion-and-retention.md` (the last code item; §7c.1–2 shipped in #1259).

## Changes

**`robosystems/graph_api/core/ladybug/manager.py`** — a missing `.lbug` now means "continue disposing": connections closed, WAL and file removed if present, Lance indexes and DuckDB staging disposed (unless `preserve_duckdb`), and — for a base name only — the `-wip`/`-prev` temporaries and their WALs removed too (they share the base's staging and indexes rather than owning any; a temporary deleted by its own name, which is the materialize flow under its lock, removes only itself). The result carries `existed` and `removed`. File removal is factored into `_remove_graph_files`.

**`robosystems/graph_api/routers/databases/management.py`** — `DELETE /databases/{graph_id}` returns the manager's result instead of a fixed message, so `existed`/`removed` reach the caller.

**`robosystems/operations/graph/deprovision_service.py`** — logs the already-absent case with the count of paths the node disposed. The 404 branch stays: a graph node on an AMI that predates this change still 404s, and convergence there still matters more than the residue (which storage reclaim can pick up).

Side effect worth knowing: `storage_reclaim` previously skipped orphan vector/staging fragments because the delete 404'd with no `.lbug` to anchor on (its own comment says so). Those now dispose on the same call, which is what the sweep's "an orphan is a whole estate" comment intended.

Reviewers: the ordering question is whether sweeping `-wip` on a base-name delete can race an active blue-green build. Base-name deletes come from deprovision (the graph is being destroyed), creation rollback (no build exists), and the rebuild paths, which already hold the materialization lock; the router's lock protocol for suffixed names is unchanged.

## Breaking Changes

None. The Graph API delete response gains fields; the internal client already returns the JSON as-is. No SDK-facing surface (neither tier) is touched.

## Testing

- `uv run pytest tests/graph_api/test_database_manager.py tests/graph_api/test_cluster_server.py tests/operations/graph/test_deprovision_service.py tests/operations/graph/test_storage_reclaim.py tests/operations/graph/test_subgraph_service.py tests/operations/graph/test_graph_creation_cleanup.py` — 147 passed.
- Mutation-checked: with the 404-before-cleanup restored, `test_delete_database_missing_lbug_still_disposes_side_stores` fails; with the blue-green sweep removed, that test and `test_delete_database_sweeps_the_blue_green_temporaries_of_a_base` fail.
- `just test-code` — ruff, format, basedpyright (0 errors), cf-lint all pass.

## Certification

### DIFF
--- a/robosystems/graph_api/client/client.py
+++ b/robosystems/graph_api/client/client.py
@@ -711,10 +711,13 @@ class GraphClient(BaseGraphClient):
     ``staging_only`` does the inverse, dropping staging and keeping the graph.
     The two are mutually exclusive.
 
-    Deleting a ``-wip``/``-prev`` name is guarded by the base graph's
-    materialization lock. A caller that already holds it (the materialize flow
-    cleaning up its own WIP) passes ``lock_token`` so the endpoint does not
-    re-acquire — without it, the delete 409s against the caller's own lock.
+    Every delete is guarded by the base graph's materialization lock — a
+    ``-wip``/``-prev`` name because it is a build artifact, a base name because
+    the node sweeps that graph's ``-wip``/``-prev`` alongside it. A caller that
+    already holds the lock (the materialize flow cleaning up its own WIP, or
+    deleting its own base on a rebuild) passes ``lock_token`` so the endpoint
+    does not re-acquire — without it, the delete 409s against the caller's own
+    lock; without the lock at all, it 409s while a build is in progress.
     """
     params = {}
     if preserve_duckdb:
--- a/robosystems/graph_api/core/ladybug/manager.py
+++ b/robosystems/graph_api/core/ladybug/manager.py
@@ -264,32 +264,39 @@ class LadybugDatabaseManager:
 
     ``preserve_duckdb`` keeps the staging database *and* the LanceDB indexes
     built from it, so LadybugDB can be rebuilt without re-running staging.
+
+    A missing ``.lbug`` is not a refusal. The graph file is the first thing
+    a delete removes, so a teardown that died between it and the side stores
+    comes back to exactly this state — and refusing it (a 404 before any
+    cleanup) left the Lance directory, the DuckDB staging and any blue-green
+    temporary on a volume the registry was about to hand to the next tenant.
+    Disposal continues from wherever the previous attempt stopped;
+    ``existed`` in the response says whether the graph file was there this
+    time, and ``removed`` lists what this call took off the disk.
+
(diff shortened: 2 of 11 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Addressed in `203fdf6e`:

- **Base-name deletes now take the base `MaterializationLock` in the router** (not only `-wip`/`-prev` names). A rebuild or teardown racing an extensions blue-green build gets a 409 and retries — deprovision's teardown sensor already re-selects a `partial` graph — instead of the sweep removing the WIP under the build. The materialize flow's own rebuild (`_ensure_database(rebuild=True)`, under its lock) now passes its token through, so it does not 409 against itself. `direct_materialization.py` / `dagster/jobs/graph.py` are unchanged: their base-name delete is what the router now guards, and a 409 there surfaces as the rebuild failing rather than a corrupted build. On the question — the overlap is possible today (a stale-triggered ledger materialize and a user-triggered core rebuild on the same tenant graph), which is why the lock, not a comment, is the answer.
- Deprovision's `existed=False` log branch is pinned by a test; `_remove_graph_files` docstring notes the sweep caller.
- Tests: router lock acquire/release, 409 while held, token passthrough; direct rebuild passes its token; mutation-checked (suffix-only condition restored → 2 router tests fail; token dropped → the materialize test fails).

@claude please re-review this PR

jfrench9:
Docstring nit addressed in `3fcd2d77` — `delete_database`'s docstring now says every delete is lock-guarded, and what the two 409s mean.

jfrench9:
CI failure fixed in `8f9093a9`: two test files my focused runs had not covered (`tests/graph_api/core/ladybug/test_manager.py`, `tests/graph_api/routers/databases/test_management.py`) still asserted the old contract — 404 on a missing `.lbug`, a fixed endpoint message, and base-name deletes never touching the lock. Rewritten to the new one; `tests/graph_api` is 1,392 passed locally.

PR #1262, opened 2026-08-26
chore(tests): run the suite in parallel with pytest-xdist

jfrench9:
## Summary

Runs the test suite in parallel with `pytest-xdist` (`-n auto --dist loadfile`) in `just test` / `test-full` / `test-cov` and CI. The CI pytest step was at 12m36s (#1258) against its 15-minute timeout; locally the full non-slow suite (13,792 tests) now runs in ~2 minutes on 14 cores versus 12+ serial. Closes `local/RoboSystems/specs/tooling/test-suite-parallelization.md` §1.

## Changes

**`tests/xdist_workers.py`, `tests/__init__.py`** — each xdist worker gets its own platform test database (`robosystems_test` → `robosystems_test_gw0`, created on demand). The rewrite runs from the test package's `__init__`, which is the earliest point after pytest-env applies the `[pytest] env` block and before anything imports `robosystems` — `robosystems.config.env` reads `DATABASE_URL` at import and `robosystems.db.platform` binds its engine from it at import, so a later hook would leave the app on the shared database while the fixtures truncate a private one. Both `TEST_DATABASE_URL` (fixtures) and `DATABASE_URL` (app engine) move together. The extensions DB stays shared: its tests use uuid-named throwaway schemas (in the platform DB) or per-file module-scoped tenant schemas, and `--dist loadfile` keeps a file's tests on one worker. Valkey and LocalStack stay shared; tests key by their own ids. The module docstring records each of those choices and why.

**`tests/conftest.py`** — the per-test cleanup missed four tables with plain FKs into truncated ones (`org_invitations`, `user_passkeys`, `user_mfa_recovery_codes`, `graph_credit_transactions`). One file's leftover invitation rows made the `users` delete fail, and the silent `except: rollback()` then kept *every* later test's rows alive on that session — surfacing as duplicate-key errors in an unrelated billing file. Serially the billing file happened to sort earlier; `tests/routers/auth/test_auth.py tests/operations/billing/test_repository_subscriptions.py` reproduces it in that order on `main`. The cleanup now warns with the offending constraint instead of hiding it (that warning is how the fourth table was found). The complete gap list was computed by walking `Base.metadata` for non-CASCADE FKs into the truncated set.

**`tests/graph_api/test_cluster_server.py`** — `test_execute_query_endpoint` consulted the real admission controller, which reads host CPU/memory headroom and 503s under a parallel run; it now admits unconditionally like the other endpoint tests.

**`tests/test_xdist_workers.py`** — the URL rewrite, the no-op outside xdist, the extensions DB staying shared, and — only under `-n` — a live assertion that this worker's `TEST_DATABASE_URL` *and* the app's bound engine both point at the worker's copy.

**`justfile`, `.github/workflows/test.yml`, `pyproject.toml`, `uv.lock`** — `pytest-xdist` in the dev extras; `-n auto --dist loadfile` on the three test recipes and the CI step (same flags as `just test`).

Reviewers: the one design call worth a look is the shared extensions DB + `loadfile` rather than a per-worker clone; the alternative (`CREATE DATABASE … TEMPLATE extensions`) needs the template free of connections, which the local dev stack's API never leaves it.

## Breaking Changes

None. Test infrastructure only.

## Testing

- Eight full parallel runs during development (`-n auto --dist loadfile`, 14 cores). The final verification pair, after the last cleanup fix: **13,792 passed, 43 skipped — 2:12 and 2:10**, no cleanup warnings, no failures.
- `uv run pytest -n 2 tests/test_xdist_workers.py` — 6 passed including the live per-worker assertion; serial: 5 passed, 1 skipped.
- The order-dependence reproduction: `uv run pytest tests/routers/auth/test_auth.py tests/operations/billing/test_repository_subscriptions.py` — 4 errors before the conftest fix, 33 passed after; mutation-checked (dropping the `OrgInvitation` delete brings the 4 errors back and the new warning names the FK).
- `just test-code` — ruff, format, basedpyright (0 errors), cf-lint, actionlint all pass.

## Certification

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -238,8 +238,11 @@ jobs:
       - name: Run tests
         timeout-minutes: 15
         run: |
-          # Run tests (exclude slow tests)
-          uv run pytest -v --tb=short -m "not slow"
+          # Run tests (exclude slow tests), one xdist worker per runner core.
+          # Each worker gets its own platform test database, created on demand
+          # (tests/xdist_workers.py); the extensions DB stays shared, which
+          # `--dist loadfile` makes safe. Same flags as `just test`.
+          uv run pytest -n auto --dist loadfile --tb=short -m "not slow"
 
       - name: Run dbt tests
         timeout-minutes: 1
--- a/justfile
+++ b/justfile
@@ -128,19 +128,23 @@ test-all:
     @just cf-lint-all
     @just lint-actions
 
-# Run tests (exclude slow tests)
+# Run tests (exclude slow tests). Parallel across cores: each xdist worker
+# gets its own platform test database (tests/xdist_workers.py); `loadfile`
+# keeps a file's tests on one worker so module-scoped tenant schemas in the
+# shared extensions DB never race themselves.
 test module="":
     uv run pytest \
         {{ if module != "" { "tests/" + module } else { "" } }} \
+        -n auto --dist loadfile \
         -m "not slow"
 
 # Run ALL tests including slow ones
 test-full:
-    uv run pytest
+    uv run pytest -n auto --dist loadfile
 
 # Run tests with coverage
 test-cov:
-    uv run pytest --cov=robosystems tests/
+    uv run pytest -n auto --dist loadfile --cov=robosystems tests/
 
 # Run the tenant-isolation harness against a deployment (default: local stack)
 test-isolation target="http://localhost:8000":
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,25 +110,22 @@ dependencies = [
 dev = [
     # RoboSystems client for demo and testing.
     "robosystems-client>=1.11.0,<2.0",
-
     # Testing framework
     "pytest>=9.0.0,<10.0",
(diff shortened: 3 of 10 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Pushed `8d7f71ba` after opening: the fourth cleanup-list table (`graph_credit_transactions`) that the new warning surfaced in the previous verification pair — the PR body's final numbers (13,792 passed, 2:12 / 2:10, clean) were measured with it. @claude please include it in the review.

jfrench9:
CI failure fixed in `47f23ccb`: `tests/operations/graph/test_table_service.py` assigned `env.USER_DATA_BUCKET = "test-bucket"` directly (its `monkeypatch.setenv` restored the env var, not the attribute), so on the 2-worker CI distribution `test_s3_backup.py` later probed a bucket that does not exist. Now `monkeypatch.setattr`, restored. The CI run itself is the number this spec wanted: **13,288 tests reached at 5:48 on the 2-core runner** before that error, against 12m36s serial.

PR #1264, opened 2026-08-26
feat(mcp): OAuth 2.1 authorization server and graph-agnostic /v1/mcp route

jfrench9:
MCP clients can now connect with OAuth instead of a pasted key. Behind `MCP_OAUTH_ENABLED` (off by default; the surface answers 404):

- Built-in OAuth 2.1 authorization server: RFC 8414 / RFC 9728 discovery under `/.well-known`, authorization-code flow with PKCE, a consent screen on the login home that picks the one graph a grant covers, token endpoint with rotating refresh tokens, RFC 7009 revocation, RFC 9207 `iss`
- Client registration three ways: RFC 7591 dynamic registration, Client ID Metadata Documents, and operator pre-registered clients (`admin oauth create-client`)
- OAuth bearer carriage on `POST /v1/graphs/{g}/mcp` (bound to that URL); new `POST /v1/mcp` accepts OAuth only and dispatches on the grant's graph
- Every MCP tool now carries a `title` and explicit annotations
- `server.json` for the official MCP registry; README and API description cover the OAuth path
- Platform migration `39c73ea18599` (three tables)

**Removed: the REST MCP tool endpoints** `GET /v1/graphs/{g}/mcp/tools` and `POST /v1/graphs/{g}/mcp/call-tool`, their envelope models, and the `MCP` OpenAPI tag. Every MCP client speaks the Streamable HTTP transport; the REST pair had no remaining callers (apps, SDK facades, or production traffic). Generated-tier SDK removal on the next regeneration: `listMcpTools`, `callMcpTool` — call out in the release notes. The `@robosystems/mcp` bridge's legacy mode targeted these endpoints and is retired separately.

Rollout: apply the platform migration, then set `features/MCP_OAUTH_ENABLED` in SSM.
Companion: robosystems-app `feature/mcp-oauth-consent` (the consent page).


jfrench9:
@claude please review this PR. Focus areas: correctness of the OAuth authorization-code and refresh flows, the audience/graph binding of tokens on both MCP routes (per-graph and `/v1/mcp`), the dynamic-registration and metadata-document validation, and whether any of the new surface bypasses the existing MCP authorization gauntlet. The endpoints are schema-excluded by design.

jfrench9:
Review follow-up (c102dbb5): deactivating a client now revokes its grants and tokens immediately rather than at the next cache miss (item 3), and the dead `import re` is gone (item 1). Item 2 left as-is: the token request's `redirect_uri` is checked whenever presented, and the code is already bound to the client and gated by PKCE, which is the OAuth 2.1 posture. The CodeQL note on the token fingerprint is the same full-entropy-digest pattern as `UserAPIKey`/`UserToken` and is dismissed as a false positive with that rationale.

jfrench9:
CI note: the two red runs were one test each — first the `BURST_LIMITS` snapshot (pinned in b1fce13d), then `tests/routers/admin/test_orgs.py::TestListOrgs::test_list_orgs_includes_max_graphs`. The second was caused by this branch: `oauth_grants` has a plain FK into `users`, so the per-test cleanup's `users` delete failed after any test that minted a grant, rolled back, and leaked that test's rows into the rest of the worker until the admin org list overflowed its 100-row page. 692cfd75 adds the three OAuth tables to the cleanup list (the harness README's rule #1); an audit of the model metadata finds no other unlisted FK table.

PR #1265, opened 2026-08-26
fix(graph): keep a parked shared master's volume row out of the stale-volume sweep

jfrench9:
## Summary

The daily stale-volume sweep aged unattached volume-registry rows from the volume's *creation* date, so a data volume that parks nightly (the shared SEC master's) lost its registry row the first time the master slept through a sweep after the volume turned 30 days old. The next wake found no registered SEC volume, minted an empty replacement while the original sat `available` in the same AZ, and the nightly chain failed at staging with nothing alarming. This fixes the sweep, teaches the launch path to recover a tagged-but-unregistered volume before minting, and pages when a shared master mints a volume at all.

## Changes

- **`robosystems/operations/graph/infrastructure.py`** — `InstanceMonitor.cleanup_stale_volumes` now measures "unattached for more than `STALE_VOLUME_DAYS`" from `last_detached` (falling back to `last_attached`, then `created_at` for rows that have never been through a detach), and never drops a row that still lists databases — mirroring the volume manager's orphan sweep, which already refuses to delete a volume carrying data. Docstrings updated here and on the Dagster op (`robosystems/dagster/jobs/infrastructure.py`).
- **`bin/lambda/graph_volume_manager.py`**
  - `handle_instance_launch` (shared-master path): when the registry offers no SEC volume, `recover_unregistered_shared_volumes` looks for `available` EC2 volumes in the AZ tagged `ManagedBy=GraphVolumeManager` / `Tier` / `DatabaseId` for the database, rebuilds their registry rows from the tags (oldest first), alerts, and offers them to the normal claim loop. Only then does the path fall through to minting.
  - `report_shared_master_volume_created`: publishes a `SharedMasterVolumeCreated` metric and sends an SNS alert before a shared master mints a new data volume. Writers minting their first volume are unaffected.
  - `sync_registry_with_ec2` add-missing path: falls back to the `DatabaseId` tag volumes actually carry when the `Databases` tag (a snapshot tag) is absent, so a re-added row is not written with `databases=[]`.
- **`cloudformation/graph-volumes.yaml`** — new `SharedMasterVolumeCreatedAlarm` (`robosystems-graph-volumes-${Environment}-shared-master-volume-created`) on that metric, wired to the existing volume alert topic. Reviewers: this is a once-per-lifetime event for a shared repository, so the threshold is any occurrence in an hour.
- **Tests** — `tests/operations/graph/test_infrastructure.py`: a 45-day-old row detached 2 days ago is kept; a row carrying databases is never dropped regardless of age. `tests/lambda/test_graph_volume_manager.py`: a tagged volume with no registry row is re-registered and claimed (with alert); a registered volume takes precedence over tag recovery; minting a shared-master volume emits the metric + alert while a writer minting its first volume does not.

## Breaking Changes

None. No API, GraphQL, or SDK surface is touched; the volume-registry row gains optional `size` / `recovered_at` attributes on recovered rows only.

## Testing

- `uv run pytest tests/operations/graph/test_infrastructure.py tests/lambda/test_graph_volume_manager.py` — 72 passed (against the local stack).
- The full `just test-all` unit run was not executed in-session.
- Live: the recovery procedure this PR automates was performed by hand in production today (registry row restored, manual wake claimed the original volume, incremental staging proceeded on it).

## Certification

### DIFF
--- a/bin/lambda/graph_volume_manager.py
+++ b/bin/lambda/graph_volume_manager.py
@@ -205,7 +205,14 @@ def handle_instance_launch(event: dict[str, Any]) -> dict[str, Any]:
     # attaching — two concurrent master launches must not both attach the same
     # volume and burn the VolumeInUse retry loop.
     reclaim_stale_claims(az, tier)
-    for existing_volume in find_volumes_with_database("sec", az, tier):
+    candidates = find_volumes_with_database("sec", az, tier)
+    if not candidates:
+      # The registry row is the only thing this path consults, but the EBS
+      # volume's tags carry the same facts durably. When the row is gone and
+      # the volume is not, re-register it — minting an empty replacement while
+      # the real one sits available in the same AZ is the worse failure.
+      candidates = recover_unregistered_shared_volumes("sec", az, tier)
+    for existing_volume in candidates:
       volume_id = existing_volume["volume_id"]
       if not claim_volume(volume_id, instance_id):
         continue
@@ -294,6 +301,8 @@ def handle_instance_launch(event: dict[str, Any]) -> dict[str, Any]:
 
   # No existing volume found, create new one
   logger.info("No existing volume found, creating new volume")
+  if node_type == "shared_master":
+    report_shared_master_volume_created(instance_id, az, tier)
   return create_and_attach_volume(instance_id, tier, az, databases, node_type)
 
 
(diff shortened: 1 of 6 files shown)


jfrench9:
@claude please review this PR

PR #1274, opened 2026-08-27
fix(oauth): harden the MCP OAuth server after its implementation review

jfrench9:
## Summary

Hardening pass on the MCP OAuth 2.1 authorization server (`v1.10.2`) following a line-by-line review of the implementation. Six defects and three low-severity edge cases, each with a regression test. No new surface, no schema change.

## Changes

**Account lifecycle**
- `operations/admin/user_deletion.py` — deleting a user now sweeps their OAuth tokens and grants (deepest first) before the account row, and purges the access-token validation cache entries the same way API-key fingerprints are purged. `oauth_grants.user_id` is a plain FK into `users`, so any account that had ever connected an OAuth client could not be deleted. The deletion plan's `removals` gains an `oauth_grants` count.

**Rate limiting for the graph-agnostic route**
- `config/rate_limits.py`, `middleware/rate_limits/subscription_rate_limits.py` — `POST /v1/mcp` is now categorized as `GRAPH_MCP` and opted into subscription-tier limits. Previously it fell through to the generic API limiter because both functions key on `/v1/graphs/{graph_id}/…`.
- `middleware/rate_limits/rate_limiting.py` — when the path names no graph, the limiter reads the graph the auth dependency published on `request.state.auth_graph_id`, so a directory connector draws from the grant's graph budget rather than the caller's account tier. The existing membership gate still applies.
- The authorization endpoint's burst limit now fails closed like the token and registration endpoints (it cannot complete without Valkey, and it triggers a client-metadata fetch).

**Resilience of the token path**
- `operations/oauth_server/clients.py`, `routers/oauth/server.py` — `resolve_client(..., allow_stale_mirror=True)` on the token and revocation endpoints: when a Client ID Metadata Document cannot be fetched or no longer validates, the mirrored `oauth_clients` row (the last document that did validate) stands in, so an outage at a metadata host does not fail every refresh for that client's users. The authorization endpoint never takes the fallback — redirect URIs must come from a live document. Deactivated mirrors are never served.
- `models/core/user/oauth_client.py` — `upsert_cimd` handles the first-contact race (two concurrent first resolutions of one document) by re-reading the winner's row instead of surfacing the unique-violation as a 500.
- `operations/oauth_server/cimd.py` — a non-numeric `Content-Length` on the metadata response no longer raises; the streamed size cap still applies.

**Refresh-token replay**
- `operations/oauth_server/tokens.py` — the consumed-token check now runs before the expiry check, so a replayed refresh token that has since expired still revokes its family (its successor may be live). A replay against an already-revoked family answers `invalid_grant` without re-logging.

**Retention**
- `dagster/jobs/infrastructure.py` — new `cleanup_stale_oauth_artifacts` op on `hourly_auth_cleanup_job`: deletes tokens expired or revoked past the 30-day retention window (`OAuthToken.cleanup_expired` now covers revoked rows too) and dynamic registrations that expired without ever reaching a consent (`OAuthClient.cleanup_expired_registrations`, new). Rotation mints two rows per refresh, so nothing was reclaiming them.

**MCP tool copy**
- `middleware/mcp/tools/graph_tools.py` — `create-subgraph` and `list-subgraphs` no longer tell every connector to "reuse this connector's API key" on a subgraph URL. An OAuth grant is bound to exactly one route, so for an OAuth connector the answer is to add the subgraph URL and authorize it. The descriptions state both cases; the results carry a `credential` field resolved from the request's auth method.

Reviewers may want to look closest at the `allow_stale_mirror` fallback (deliberately not offered to `/v1/oauth/authorize`) and the `request.state.auth_graph_id` read in the limiter (guarded to `str`, and still subject to `_caller_is_authorized_on_graph`).

## Breaking Changes

None. All OAuth endpoints are excluded from the OpenAPI schema, so neither SDK tier is affected. The `credential` key added to the `list-subgraphs` tool result and the `oauth_grants` key in the deletion plan are additive.

## Testing

`just test-code` — passes (ruff, format, basedpyright 0 errors, cf-lint).

Targeted `pytest` over `tests/operations/oauth_server`, `tests/routers/oauth`, `tests/middleware/auth/test_oauth_dependencies.py`, `tests/routers/graphs/mcp`, `tests/operations/admin/test_user_deletion.py`, `tests/middleware/mcp/tools/test_graph_tools.py`, `tests/middleware/rate_limits`, `tests/dagster/test_infrastructure_jobs.py`, `tests/dagster/test_definitions.py`, `tests/models/core/user`, `tests/routers/admin` — 916 passed. New tests cover: user deletion with a grant and tokens; a parent-graph grant refused at a subgraph URL; `/v1/mcp` categorization and graph-budget keying; the stale-mirror fallback (token path only, never a deactivated mirror); the CIMD first-contact race; the bogus `Content-Length`; replay-after-expiry family revocation; the OAuth sweep op; and the fail-closed posture of the four OAuth buckets.

The full `just test-all` suite was not run in-session.

## Certification

### DIFF
--- a/robosystems/config/rate_limits.py
+++ b/robosystems/config/rate_limits.py
@@ -331,6 +331,12 @@ class RateLimitConfig:
     if path.startswith("/v1/"):
       path = path[4:]
 
+    # The graph-agnostic MCP transport (POST /v1/mcp): the OAuth grant, not
+    # the path, names the graph, so the route is categorized by its suffix
+    # alone — it must land in the MCP bucket like its per-graph sibling.
+    if path == "mcp":
+      return EndpointCategory.GRAPH_MCP
+
     # Non-graph scoped endpoints - check these first
     if path.startswith("auth/"):
       return EndpointCategory.AUTH
--- a/robosystems/dagster/jobs/infrastructure.py
+++ b/robosystems/dagster/jobs/infrastructure.py
@@ -20,7 +20,7 @@ from sqlalchemy import and_, or_
 
 from robosystems.config import env
 from robosystems.dagster.resources import DatabaseResource
-from robosystems.models.core import UserAPIKey
+from robosystems.models.core import OAuthClient, OAuthToken, UserAPIKey
 
 # ============================================================================
 # Environment-based Schedule Status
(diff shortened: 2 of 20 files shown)


jfrench9:
@claude please review this PR

PR #1275, opened 2026-08-27
feat(oauth): connected-apps endpoints to list and revoke a user's grants

jfrench9:
## Summary

Adds the connected-apps surface for MCP OAuth: a user can list the clients they have authorized and revoke one at a time. Until now the only ways to revoke an OAuth grant were a password change (every grant) or an operator deactivating the client (every user of it). This is the API half of a connected-apps page and lands now so it rides the next SDK minor.

## Changes

- `operations/oauth_server/grants.py` (new) — `list_user_grants` (active grants with the client's name/URI/trust flag, the graph and its display name, resource, scope, timestamps; graph names read straight off the table so a grant on a since-deprovisioned graph still shows one) and `revoke_user_grant` (owner-checked; revokes the grant and every token minted from it via the existing `OAuthGrant.revoke`, which also clears the tokens' validation-cache entries; idempotent). `GrantNotFound` is one answer for unknown and someone-else's — no existence oracle.
- `routers/user/oauth_grants.py` (new), registered in `routers/user/__init__.py` — `GET /v1/user/oauth/grants` (`listUserOAuthGrants`) and `DELETE /v1/user/oauth/grants/{grant_id}` (`revokeUserOAuthGrant`), tag `User`, standard `get_current_user` + user-management rate limit, same error and metrics discipline as the API-key endpoints. Active grants only, mirroring the API-key listing rule.
- `models/api/user.py`, `models/api/__init__.py` — `OAuthGrantInfo`, `OAuthGrantsResponse`.
- `middleware/auth/README.md` — documents the endpoints next to the existing revocation paths.

## Breaking Changes

None. Purely additive: two new operations and two new response models on the generated tier — an SDK regen opportunity for the upcoming client minor.

## Testing

`just test-code` — passes (ruff, format, basedpyright 0 errors, cf-lint).

`pytest tests/routers/user tests/operations/oauth_server tests/routers/oauth tests/models/api` — 576 passed. New `tests/routers/user/test_oauth_grants.py` covers: the listing returns only the caller's live grants with client and graph details (another user's and a revoked grant excluded; an unknown graph yields `graph_name: null`); revoking marks the grant and both its tokens revoked and drops it from the list; revoking twice is idempotent; someone else's grant and an unknown id are both 404.

Verified against the running local stack that both operations appear in `openapi.json` under the `User` tag.

The full `just test-all` suite was not run in-session.

## Certification

### DIFF
--- a/robosystems/middleware/auth/README.md
+++ b/robosystems/middleware/auth/README.md
@@ -83,8 +83,13 @@ and revoked tokens answer **401 `invalid_token`** (clients refresh); a valid
 token whose user lost access answers **403 `insufficient_scope`**. Password
 change, deactivation and any other `session_version` bump revoke the user's
 OAuth tokens outright (`User._revoke_oauth_tokens`), since like API keys they
-carry no session version. OAuth tokens are accepted on the two MCP routes only
-— `get_current_user` and the graph-scoped REST dependencies never resolve them.
+carry no session version. A user revokes one connection at a time through the
+connected-apps endpoints (`GET`/`DELETE /v1/user/oauth/grants`,
+`operations/oauth_server/grants.py`) — the grant and every token minted from
+it go together, and their cache entries are cleared, so the client fails at
+its next call rather than its next cache miss. OAuth tokens are accepted on the
+two MCP routes only — `get_current_user` and the graph-scoped REST dependencies
+never resolve them.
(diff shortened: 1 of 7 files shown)


jfrench9:
@claude please review this PR

PR #1277, opened 2026-08-27
fix(infra): tighten deploy-identity boundaries across roles and stacks

jfrench9:
## Summary

Narrows several deploy-time and instance identities to the resources they actually operate on, and binds the production deploy pipeline to the `production` GitHub environment so its protection rules gate the run. Design notes and the residual list live in `local/RoboSystems/specs/security/deploy-identity-boundary.md`.

## Changes

- `cloudformation/bootstrap-oidc.yaml` — frontend deploy role: CloudFormation mutations pinned to the frontend stacks; Secrets Manager access reduced to what the deploy workflows call; statements for services the frontend stacks never create (ECS, ELB, application-autoscaling, EC2 security groups, Lambda, Logs, CloudWatch alarms) removed; `ecs-tasks` dropped from the PassRole condition; service-linked-role creation scoped to App Runner. New `FrontendRoleBoundary` managed policy (ARN exported) with `iam:PutRolePermissionsBoundary` conditioned to it — the enabling step for a permissions boundary on the App Runner roles the app stacks create.
- `cloudformation/bastion.yaml` — bastion secret reads scoped to its own environment.
- `cloudformation/postgres.yaml` — `VpcCidr` parameter replaces the CIDR literal on the database security group; rotation Lambda receives `DB_INSTANCE_IDENTIFIER`; invoke permission carries `SourceAccount`.
- `cloudformation/valkey.yaml` — `ValkeyUrlWithAuth` output/export removed (no importers; `ValkeyUrl` is what the services consume); `SourceAccount` on the rotation permission.
- `cloudformation/api.yaml`, `cloudformation/graph-infra.yaml` — `SourceAccount` on the rotation-Lambda invoke permissions.
- `cloudformation/worker.yaml` — `GraphVolumesStackName` parameter; the volume-manager invoke is pinned to this environment's function.
- Workflows — `deploy-vpc.yml` exports `vpc_cidr`; `deploy-postgres.yml` takes `vpc_cidr`; `deploy-dagster.yml` takes `graph_volumes_stack_name` (required); `prod.yml`/`staging.yml` pass both from existing outputs. `prod.yml`'s `create-deployment` job, which every `deploy-*` job fans in through, now declares `environment: production`.
[...]

PR #1279, opened 2026-08-27
fix(infra): drop the PassedToService condition on the frontend PassRole

jfrench9:
## Summary

Follow-up to #1277. The frontend deploy role's `iam:PassRole` statement carried an `iam:PassedToService` condition that App Runner does not populate, so every pass of the access/instance roles was denied once the role went live. The staging `robosystems-app` deploy after `just bootstrap` failed at the App Runner service update with exactly that denial. The condition had been in the template since July but the OIDC stack had not been re-applied since 2026-07-04, so it was never exercised before today.

## Changes

- `cloudformation/bootstrap-oidc.yaml` — remove the `iam:PassedToService` condition from the frontend role's `IAMPassRole` statement; the resource patterns (`*-apprunner-*` roles) remain the bound, and those roles' trust policies already restrict them to the App Runner principals. Comment updated to say why.

## Breaking Changes

None. Owner-deployed: `just bootstrap` after merge, then `continue-update-rollback` on the staging app stack and re-run its deploy.

## Testing

- `cfn-lint` on the template — clean.
- Evidence for the diagnosis: the failed run's `UPDATE_FAILED` reason names `iam:PassRole` on the instance role; the template history vs. the stack's update history shows the condition was never live before today.

## Certification

### DIFF
--- a/cloudformation/bootstrap-oidc.yaml
+++ b/cloudformation/bootstrap-oidc.yaml
@@ -753,8 +753,7 @@ Resources:
               # access/instance roles the frontend stacks actually create
               # (robosystems-app-*, roboledger-*, roboinvestor-*), so a
               # compromised frontend workflow cannot attach policies to, or pass,
-              # backend/deploy/SSO roles. Reads stay broad; PassRole is
-              # additionally constrained by iam:PassedToService. Role-name
+              # backend/deploy/SSO roles. Reads stay broad. Role-name
               # prefixing alone does not stop creating a NEW privileged role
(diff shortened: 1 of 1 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Cross-checked the reviewer's suggestion 1: in all three app templates the access role's trust policy names only `build.apprunner.amazonaws.com` and the instance role's only `tasks.apprunner.amazonaws.com`, so the resource-ARN scoping plus those trust policies remain the bound as the comment states.

PR #1281, opened 2026-08-27
chore(bootstrap): make re-runs safe — oidc-only mode, change-set preview, gated app config

jfrench9:
## Summary

Makes `bin/setup/bootstrap.sh` safe to re-run against a live account. Today a re-run walks through every first-run prompt — including the two that can rewrite production (Secrets Manager, and `gha.sh`'s re-assertion of all ~185 repository variables) — and applies the OIDC stack blind. This week showed the cost of the latter: a hardening edit to `bootstrap-oidc.yaml` sat unapplied from July until the next bootstrap, with nothing indicating the live stack was behind.

## Changes

- `bin/setup/bootstrap.sh`
  - Flag parsing: `--oidc` (only the OIDC stack + `AWS_ROLE_ARN`/`AWS_ACCOUNT_ID`/`AWS_REGION`), `--with-app-config`, `--help`; profile/region positionals unchanged.
  - OIDC update goes through a **change set**: lists the resource changes (with replacement flags) and asks before executing; a stack that already matches the template is reported as such with no prompt; prints the template's last commit/date/branch (and uncommitted-edit state) next to the stack's last-applied time; refuses busy or failed stack states. First-run `create-stack` path unchanged.
  - On an existing stack the trusted org/repo are read from the stack's parameters rather than prompted, so a re-run cannot re-point the deploy roles by a stray keystroke.
  - Application config (Secrets Manager / GitHub variables) runs only when this run created the OIDC stack or `--with-app-config` is given; otherwise it prints how to run each piece deliberately. The GitHub-variables prompt now requires a typed `yes` and states what it rewrites.
  - ECR lifecycle: when the repository already has a policy, matches → silent success, differs → warning pointing at `just bootstrap-ecr-lifecycle`; the prompt only appears when there is no policy. `ECR_LIFECYCLE_POLICY` override unchanged.
  - Identity variables are written only when the value differs (`set_variable_if_changed`).
  - `--oidc` skips `.envrc` handling and prints a short summary.
- `justfile` — new `bootstrap-oidc [profile] [region]` recipe.
- `bin/setup/README.md` — usage, the two flags, a "Re-running" paragraph, ECR step note.

## Breaking Changes

None. `just bootstrap` on a fresh account behaves as before (first run still offers secrets/variables). On an existing account it no longer reaches the secrets/variables prompts unless asked.

## Testing

`bash -n` and `shellcheck -S warning` (only the four pre-existing SC2155s in the SSO helpers remain); `--help` output; the null-coalescing JMESPath for last-applied time checked against both stack shapes with the `jmespath` library. Not exercised end to end here — the flow needs an interactive SSO session. The first real run of `just bootstrap-oidc` against the current stack should report "already matches the template" (the stack was applied from `main` at 05:20Z today).

## Certification

### DIFF
--- a/bin/setup/README.md
+++ b/bin/setup/README.md
@@ -75,7 +75,8 @@ The complete bootstrap process for a fresh deployment:
 │    - Repository name is always "robosystems" (fleet-uniform, even on        │
 │      a renamed fork — the deploy role's ECR scope assumes it)               │
 │    - Image scanning on push, AES256 encryption                              │
-│    - Prompts for a lifecycle policy: robust (default) / basic / skip        │
+│    - First run: prompts for a lifecycle policy (robust / basic / skip);    │
+│      later runs: silent when the live policy matches the bundled file      │
 └─────────────────────────────────────────────────────────────────────────────┘
                                     │
                                     ▼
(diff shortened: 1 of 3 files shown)


jfrench9:
@claude please review this PR

PR #1283, opened 2026-08-27
chore(ci): build the production image behind the deployment approval gate

jfrench9:
## Summary

Moves the production image build behind the deployment approval gate. Until now `build` ran in parallel with the tests, ahead of `create-deployment`, so the ECR push — and the Docker Hub publish it carries on tag/release refs — happened before anyone approved the run. Now nothing reaches a registry for a run that was never approved.

## Changes

- `.github/workflows/prod.yml` — `create-deployment` no longer needs `build`; `build` now needs `create-deployment`. Tests and `stack-config` still run before the pause, and every deploy job still fans in through `create-deployment` (the four that reach the tests only through it are unaffected). The jobs that consume `build` outputs already list `build` in their `needs`, so nothing else changes.
- The comment on `create-deployment` describes the new order.

Cost: the build (~8 min) no longer overlaps the tests (~11 min), so a prod release takes about that much longer wall-clock, and the approval prompt arrives after the tests rather than after the build. Staging is unaffected — it passes `publish_to_dockerhub: false` and has no gate.

## Breaking Changes

None.

## Testing

- `actionlint -shellcheck= .github/workflows/prod.yml` — clean (the only findings with shellcheck on are pre-existing `SC2086`/`SC2129` notes in `run:` blocks this PR does not touch)
- Job graph checked by hand: acyclic; `create-deployment` consumes no `build` outputs; every `needs.build.outputs.*` consumer already lists `build` in `needs`
- Not exercisable in-session; the next production release is the first real run

## Certification

### DIFF
--- a/.github/workflows/prod.yml
+++ b/.github/workflows/prod.yml
@@ -139,7 +139,7 @@ jobs:
     secrets: inherit
 
   build:
-    needs: [runner]
+    needs: [runner, create-deployment]
     uses: ./.github/workflows/build.yml
     with:
       environment: ${{ vars.ENVIRONMENT_PROD || 'prod' }}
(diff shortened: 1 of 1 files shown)


jfrench9:
@claude please review this PR

PR #1284, opened 2026-08-27
fix(security): last-4 tax IDs in the graph, email tokens by reference

jfrench9:
## Summary

Two small data-hygiene hardenings from the PII and data-classification review. Counterparty tax identifiers now reach the analytical graph as their last four characters only, whatever wrote the OLTP row, and the email notification job no longer embeds verification/reset/invitation tokens in its Dagster run config.

## Changes

- **`operations/extensions/materialize.py`** — the `Agent` staging projection masks `tax_id` to `***` + last-4 (empty → `NULL`). `Entity.tax_id` — the filer's own public EIN — is unchanged, as are its SEC-pipeline and export consumers. No graph schema change: the column name and type are the same, so existing tenants pick the masked value up at their next materialize with no per-graph migration.
- **`dagster/jobs/notifications.py`** — `build_email_job_config` parks the raw token in the AUTH Valkey DB under an opaque reference (`CacheDefaults.EMAIL_TOKEN_REF_TTL`, 15 min) and the run config carries only `token_ref`; `send_email_op` resolves it at send time and discards it after a successful send. A reference that has expired or already been used fails the op without sending, so a re-execution from the Dagster UI cannot resend a stale token. A failed send leaves the reference in place for the op's existing retry policy. The `token` field is removed from `SendEmailConfig` so the raw value cannot be placed in run config by any caller; all seven call sites go through the builder and are untouched.
- **`config/defaults.py`** — `EMAIL_TOKEN_REF_TTL`.
- **Tests** — `tests/dagster/test_notifications.py` (new): run config never contains the raw token, the reference resolves with the configured TTL, the op resolves-then-discards, an expired reference fails without sending, a send failure keeps the reference. `test_materialize.py` pins the mask on `Agent` and its absence on `Entity`. `test_invitations.py` updated for the new key.

Reviewers: the Valkey lookup in `_token_store()` is deliberately per-call rather than a cached client so it can be patched cleanly in tests; it mirrors `SSEEventStorage._get_sync_redis`.

## Breaking Changes

None. No REST, GraphQL, or operations-envelope shape changes; the OLTP `Agent.tax_id` column and its API responses are unchanged — only the graph projection is masked.

## Testing

- Pre-commit gate (ruff check + format + basedpyright) passed on commit.
- `uv run pytest` on `tests/dagster/test_notifications.py`, `tests/operations/extensions/test_materialize.py`, `tests/routers/auth/test_password_reset.py`, `tests/routers/auth/test_email_verification.py`, `tests/routers/orgs/test_invitations.py`, `tests/routers/user/test_user_profile.py` — 155 passed.
- Full `just test-all` not run in-session; CI runs it.

## Certification

### DIFF
--- a/robosystems/config/defaults.py
+++ b/robosystems/config/defaults.py
@@ -73,6 +73,7 @@ class CacheDefaults:
   # Authentication cache TTLs
   JWT_TTL = 1800  # 30 minutes for JWT validation cache
   API_KEY_TTL = SHORT  # 5 minutes for API key validation
+  EMAIL_TOKEN_REF_TTL = 900  # 15 minutes for a queued email's parked token
 
   # Graph/Schema cache TTLs
   SCHEMA_TTL = SHORT  # 5 minutes for schema/config cache
--- a/robosystems/dagster/jobs/notifications.py
+++ b/robosystems/dagster/jobs/notifications.py
@@ -1,5 +1,8 @@
 """Dagster jobs for notification operations (email, SMS, push)."""
 
+import secrets
+from typing import Any
+
 from dagster import (
   Backoff,
   Config,
@@ -9,6 +12,7 @@ from dagster import (
   op,
 )
 
+from robosystems.config.defaults import CacheDefaults
 from robosystems.logger import get_logger
 
 logger = get_logger(__name__)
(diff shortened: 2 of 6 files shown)


jfrench9:
@claude please review this PR

PR #1285, opened 2026-08-27
fix(credits): read repository_plan as a plain string on shared-repository paths

jfrench9:
## Summary

`UserRepository.repository_plan` has been a plain `String` column since 29389680, but four reads in `CreditService` and three in the admin repository-credit-pool endpoints still called `.value` on it — the enum-era idiom. The service reads were dead code until the operator credit pre-flight (c605f92d, v1.6.29) made `check_credit_balance` their first caller; since then every `CypherOperator` run on a shared repository (`sec`) has raised `AttributeError: 'str' object has no attribute 'value'` inside the fail-closed pre-flight and been refused as "Insufficient credits", with the error text surfacing in the console as the answer. This restores the operator on shared repositories and the three admin endpoints, and replaces the test fixture that was hiding it.

## Changes

- **`operations/graph/credit_service.py`** — `check_shared_repository_access` (three return paths: inactive subscription, zero-cost operation, funded) and `get_shared_repository_summary` now read `repository_plan` as the string it is. `access_level.value` is untouched — that column is a real `SQLEnum`.
- **`routers/admin/credits.py`** — same one-token fix on the three `RepositoryCreditPoolResponse` constructions (`list_repository_credit_pools`, `get_repository_credit_pool`, `add_bonus_credits_to_repository`). The response model already typed the field as `str`, so the wire shape does not change; these endpoints simply stop raising on real rows.
- **`tests/routers/admin/test_credits.py`** — the pool fixture built `repository_plan` from a `_FakePlan(str, Enum)` whose docstring read *"Fake enum so .value works like the router expects"*, which is why the admin tests were green against a broken router. The fixture now carries the plain string the column holds; the `Enum` import goes with it.
- **`tests/operations/graph/test_credit_service.py`** — new `TestSharedRepositoryPlanIsAPlainString` covering all four service sites (`check_credit_balance` on `sec` for an `agent_call`, the funded/inactive access-check branches, the zero-cost branch, and the repository summary). The fixtures assign a real `str` rather than leaving a `MagicMock` attribute, since a `MagicMock` answers `.value` happily — verified by stashing the source fix and watching the first test fail with the production error.

Reviewers: the only judgement call is that `access_level.value` stays; everything else is mechanical.

## Breaking Changes

None. No public shape changes; the admin response field was already declared `str`. No SDK impact.

## Testing

- `uv run pytest tests/operations/graph/test_credit_service.py tests/routers/admin/test_credits.py tests/operations/operators/test_credit_preflight.py` — 77 passed.
- Same selection with the two source files stashed — fails with `AttributeError: 'str' object has no attribute 'value'` on the first new test, so the regression coverage is real.
- `just test-code` — all checks passed (ruff, format, basedpyright 0 errors).
- The full `just test-all` unit run was not executed in-session.

## Certification


jfrench9:
@claude please review this PR

PR #1286, opened 2026-08-27
chore(ci): move the prod approval gate ahead of tests and build

jfrench9:
## Summary

Moves `create-deployment` ahead of tests and build so the production approval can be given as soon as a run starts, instead of after the ~10-minute test run. #1283 put the gate after tests, which serialised test + build ahead of every prod deploy.

## Changes

- `create-deployment` needs only `runner`; `build` still sits behind it.
- `deploy-vpc`, `deploy-s3`, `deploy-grafana` now need `test` explicitly — everything else chains from them or from `build`.
- `deployment-successful` / `handle-deployment-failure` list `test` and `build` (and `deploy-opensearch`), so a failure after the deployment record exists marks it failed instead of leaving it `in_progress`.

## Breaking Changes

None.

## Testing

`actionlint -shellcheck= .github/workflows/prod.yml` clean. First real run is the next prod release.

## Certification

### DIFF
--- a/.github/workflows/prod.yml
+++ b/.github/workflows/prod.yml
@@ -154,13 +154,11 @@ jobs:
       DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
 
   create-deployment:
-    needs: [runner, test, stack-config]
+    needs: [runner]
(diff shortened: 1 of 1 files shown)


jfrench9:
@claude please review this PR

PR #1287, opened 2026-08-27
feat(operators): harden the tool loop and schema payload for the console operator

jfrench9:
## Summary

Hardens the AI Operator's tool loop and schema payload so the console operator answers ledger questions instead of exhausting its step budget on orientation. Measured against prod today: a `quick`-mode ledger question ran 3 turns, hit the step limit, spent 114 credits, and returned zero rows plus a Cypher query for the user to run themselves — because the tenant schema (20–30k chars) was truncated at the 12k feedback cap, the one failed Cypher call burned a budgeted turn, and the forced-final nudge could still come back as a `tool_use` block that was silently dropped. First unit of `specs/ai-operators/operator-console-hardening.md` (Phase 1, API side); the console change (`standard` mode + history) follows in `robosystems-core`.

## Changes

- **`operations/operators/tool_loop.py`**
  - Per-tool feedback caps: `get-graph-schema` and `get-example-queries` escape the 12k default (48k), so the model sees the whole schema; query results stay capped.
  - Uncharged error retries: a turn whose tool calls *all* fail is not charged against `max_iterations`, up to `max_error_retries` (default 2). A mixed turn with one success is charged. Total model calls remain bounded at `max_iterations + max_error_retries + 1`.
  - The step-limit nudge sends `tool_choice: {"type": "none"}` — tools stay defined so the tool_use/tool_result transcript is valid, but the final turn is guaranteed to be text. Verified live against Bedrock `InvokeModel` with Sonnet 4.6.
  - `ToolLoopResult.iterations` now consistently counts model calls (including the nudge); new `error_retries` field.
- **`operations/operators/ai_client.py`, `tracked_ai.py`** — thread `tool_choice` through to the Bedrock request body (only alongside `tools`).
- **`operations/operators/implementations/cypher.py`** — the system prompt states the concrete step budget for the mode and that a corrected retry is free, replacing "you have a limited number of steps".
- **`middleware/mcp/client.py`** (the `get-graph-schema` payload)
  - Relationship endpoints resolve from the declared schema (`robosystems.schemas.loader`) before falling back to the name heuristics: 19 of 54 platform relationships were reported as `Unknown → Unknown` (all the `EVENT_*` verbs, `ENTRY_FROM_SCHEDULE`, `FACT_SET_CONTAINS_FACT`, `REPORT_USES_TAXONOMY`, the roboinvestor set) and 4 were mis-guessed (`DIMENSION_HAS_AXIS_ELEMENT → AxisElement`).
  - The ledger spine (`Transaction`) is no longer hand-curated — it is introspected from the catalog like `Entry`/`LineItem`, so `is_live`/`status` appear (the curated list omitted the live-row filter the prompt tells the model to use).
  - Output is nodes-first then relationships, each alphabetical (byte-stable across calls — a prerequisite for the prompt-caching spec), and the duplicated `sample_properties`/`common_properties` pair collapses to `properties`.

Reviewers: the retry accounting in `run_tool_loop` (`turn_succeeded` / `error_retries`) is the piece to read closely.

## Breaking Changes

None to the REST/GraphQL/operations surface. The `get-graph-schema` MCP tool's JSON payload renames `sample_properties`/`common_properties` → `properties` on node entries; nothing programmatic consumes those keys (checked across the three frontend apps, `robosystems-core`, both SDK facades, and this repo) — the payload is read by models.

## Testing

- `just test-code` — passes (ruff, format, basedpyright, cf-lint).
- Targeted: `tests/middleware/mcp`, `tests/operations/operators`, `tests/operations/test_ai_client.py`, `tests/routers/graphs/test_operator*.py` — 1044 + 27 passed. New tests cover the uncharged retry, the bounded retry budget, the charged mixed turn, the per-tool cap, `tool_choice` in the request body (and dropped without tools), declared-schema endpoint resolution with the loader-fault fallback, ledger-spine introspection, and nodes-first ordering.
- Live: Bedrock `InvokeModel` probe confirming `tool_choice: none` returns `end_turn` text on `us.anthropic.claude-sonnet-4-6`.

## Certification

### DIFF
--- a/robosystems/middleware/mcp/client.py
+++ b/robosystems/middleware/mcp/client.py
@@ -358,7 +358,8 @@ class GraphMCPClient:
       tables_query = "CALL SHOW_TABLES() RETURN id, name, type, comment"
       tables_result = await self.execute_query(tables_query)
 
-      schema_info = []
+      nodes: list[dict[str, Any]] = []
+      relationships: list[dict[str, Any]] = []
 
       for table in tables_result:
         table_name = table.get("name", "")
@@ -372,18 +373,17 @@ class GraphMCPClient:
           # Prefer the curated property map for well-known nodes (nice
           # ordering); for everything else introspect the real columns from
           # the catalog so the schema isn't reported as a misleading guess.
-          sample_properties = self._get_common_properties(table_name)
-          if sample_properties is None:
-            sample_properties = await self._introspect_node_properties(table_name)
+          properties = self._get_common_properties(table_name)
+          if properties is None:
+            properties = await self._introspect_node_properties(table_name)
 
-          schema_info.append(
+          nodes.append(
             {
               "label": table_name,
               "type": "node",
               "comment": table_comment,
               "description": self._get_node_description(table_name),
-              "sample_properties": sample_properties,
-              "common_properties": sample_properties,
+              "properties": properties,
             }
           )
 
@@ -391,7 +391,7 @@ class GraphMCPClient:
           # Infer relationship details without querying table structure
           from_node, to_node = self._infer_relationship_nodes(table_name)
 
-          schema_info.append(
+          relationships.append(
             {
               "label": table_name,
               "type": "relationship",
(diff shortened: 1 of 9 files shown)


jfrench9:
@claude please review this PR

jfrench9:
On the `max_tools + 1` note: the tool-turn ceiling is unchanged by this PR. Under the old `for iteration in range(max_iterations)` loop every one of the `max_tools + 1` model calls could be a tool_use turn (the nudge then answered), so the budget was already `max_tools + 1` tool-calling turns and the natural answer only *sometimes* occupied one of them. The new accounting makes that explicit rather than widening it. What was wrong was the comment ("plus a final answer turn"), which the follow-up commit replaces with what the ceiling actually is. Kept as-is deliberately: every mode's `max_tool_calls` was tuned against this ceiling, and dropping the `+1` would take `quick` to two tool turns — schema and examples, with no turn left for a query.

PR #1288, opened 2026-08-27
feat(billing): meter cache tokens and reprice AI credits to Bedrock actual

jfrench9:
## Summary

Prepares the credit meter for operator prompt caching and corrects the AI token rate card. Bedrock reports `cache_read_input_tokens` / `cache_creation_input_tokens` on every response, and once `cache_control` is in play `input_tokens` means the *uncached* remainder only — a meter that reads just `input_tokens` would silently stop billing the majority of input. This PR threads all four token counts through the billing path ahead of enabling caching, and reprices credits to what Bedrock actually bills (the `us.*` regional profiles carry a 10% premium over Anthropic list price, which the old 3/15 rates were indexed on — a structural −10% margin on every token).

## Changes

- **`robosystems/operations/operators/ai_client.py`** — `AIResponse` carries `cache_read_input_tokens` / `cache_creation_input_tokens` (populated with `.get(..., 0)` so pre-caching fixtures keep working). Request shaping is now model-family-aware: Claude 5-family models reject `temperature` with a 400 and run adaptive thinking unless explicitly disabled, so the request builder branches on `model_accepts_sampling_params`.
- **`robosystems/operations/operators/tracked_ai.py`** — accumulates cache reads/writes alongside input/output and forwards them to the credit consumer; the unbilled-spend log line includes them.
- **`robosystems/operations/operators/credit_consumer.py`** — the `CreditConsumer` protocol and all three implementations accept the two cache counts (default 0).
- **`robosystems/operations/graph/credit_service.py`** — `consume_ai_tokens` prices all four token classes and records all four counts (plus per-class costs) in the transaction metadata, so the CUR reconciliation stays reproducible once caching lands. The unknown-model fallback now reuses the Sonnet 4 pricing entry instead of a hardcoded stale copy.
- **`robosystems/config/billing/ai.py`** — rate card repriced to Bedrock `us.*` actual: input 3.3 / output 16.5 credits per 1K (was 3/15); cache reads at 0.33 (0.1× input) and 5-minute cache writes at 4.125 (1.25× input), mirroring Bedrock's own multipliers — the cache discount is passed through to the customer. Adds the `anthropic_claude_5_sonnet` entry (2.2/11) as model-upgrade prep. No silent default entry.
[...]

PR #1289, opened 2026-08-28
feat(operators): enable prompt caching and move orientation into the system prefix

jfrench9:
## Summary

Enables prompt caching on the operator loop and restructures the Cypher operator so its orientation payloads live in the cached prefix. Measured on the real operator payload, the cached prefix is 92% of a first call and cache reads bill at 0.1× — combined with dropping two orientation model calls per question, a `standard` question's input bill falls by roughly two-thirds. Stacked on #1288, which must land first: it teaches the meter to see cache tokens, and enabling caching without it would leave most input tokens unbilled.

## Changes

- **`robosystems/operations/operators/ai_client.py`** — `system` is now sent as a content-block list with a `cache_control: ephemeral` breakpoint at its end, which caches the tool definitions and system prompt together (the request renders tools → system → messages). New opt-in `cache_conversation` flag adds a second breakpoint on the trailing user turn, applied at request build without mutating the caller's transcript — a persisted marker on every past turn would exceed the 4-breakpoint limit.
- **`robosystems/operations/operators/tool_loop.py`** — every loop call (nudge included) passes `cache_conversation=True`: the transcript grows monotonically, so each call reads the cache entry the previous one wrote and extends it.
- **`robosystems/operations/operators/tracked_ai.py`** — passes the flag through.
- **`robosystems/operations/operators/implementations/cypher.py`** — the operator prefetches `get-graph-schema` (and `get-example-queries` where the graph exposes it) once, renders both into the system prompt, and withholds those two tools from the loop; the prompt tells the model to write its first query on turn one. Both payloads are deterministic per graph, so the prefix stays byte-stable per (graph, mode) — the prerequisite for cache hits. On any prefetch failure the operator falls back to the previous tool-driven orientation, for which the two tools remain on the allowlist.

## Breaking Changes

None. No API surface or SDK change. Behavior change per `standard` question: ~2 fewer model calls, a complete (untruncated) schema in front of the model, and most input tokens billed at the cache-read rate from the second call onward.

## Testing

- `just test` — full unit suite passed (14,053 passed, 43 skipped).
- New tests: system rendered as a cached content block; trailing-turn marker placement, string-content wrapping, no caller-transcript mutation, and opt-in default-off; the loop passing the flag on every call; orientation prefetched into the prompt with the reduced tool list; fallback on prefetch failure and on `{"error": ...}` payloads; schema-only graphs; deterministic truncation.

Post-deploy verification is the CUR reconciliation (it matched to 0.1 credit pre-caching and must match after) plus asserting `cache_read_input_tokens > 0` on a live probe — a sub-minimum prefix silently caches nothing, so "the call succeeded" proves nothing.

## Certification

### DIFF
--- a/robosystems/middleware/mcp/tools/manager.py
+++ b/robosystems/middleware/mcp/tools/manager.py
@@ -202,9 +202,10 @@ class GraphMCPTools:
       self.financial_statement_analysis_tool = FinancialStatementAnalysisTool(
         graph_client
       )
-      # OLTP-backed live statement — tenant entity graphs only. Skipped
-      # on shared repos (no OLTP tenant schema) and on read-only graphs.
-      if not self._is_shared_repository() and not read_only:
+      # OLTP-backed live statement — tenant entity graphs only. Skipped on
+      # shared repos (no OLTP tenant schema). A pure read, so it stays
+      # available on read-only surfaces (graph viewers, read-only operators).
+      if not self._is_shared_repository():
         self.live_financial_statement_tool = LiveFinancialStatementTool(graph_client)
 
       # Semantic enrichment tools (roboledger + manifest flag)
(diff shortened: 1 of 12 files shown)


jfrench9:
@claude please review this PR

PR #1290, opened 2026-08-28
feat(operators): widen the read-only tool surface and add a per-question credit ceiling

jfrench9:
## Summary

Completes Phase 2 of the operator console hardening: splits read from write in the MCP tool manager so read-only surfaces get every pure read, widens the Cypher operator's allowlist accordingly with the system prompt routing statement/balance/period questions to the curated tools first, and adds a per-question `max_credits` ceiling to the operator request. Stacked on #1289 (prompt caching) — deliberately, since the ~25KB of added tool definitions ride under the system cache breakpoint that PR introduces.

## Changes

- **`robosystems/middleware/mcp/tools/manager.py`** — four gate blocks split on read vs write: `live-financial-statement`, `get-graph-sync-status`, the period-workflow reads (`get-fiscal-calendar`, `get-period-close-status`, `list-period-drafts`), and the four mapping reads (`list-mapping-structures`, `get-unmapped-elements`, `get-mapping-summary`, `suggest-mapping` — all extensions-OLTP reads, no writes, no AI) now construct on read-only surfaces. Their write siblings (`materialize`, `close-period`/`reopen-period`/`backfill-plan-history`, the registrar mapping writes) still require a writable graph. The unbundled OLTP-backed reads gain an explicit shared-repository exclusion — previously provided only implicitly by the read-only gate — so nothing new is advertised on `sec`. This is a legibility fix for MCP graph viewers independent of the operator.
- **`robosystems/operations/operators/implementations/cypher.py`** — `READ_ONLY_TOOLS` widens from 8 to 34 names (curated financial reads, period/freshness reads, mapping reads, tenant OLTP reads, document reads, `recall`); a `CURATED TOOLS` prompt section, built only from tools the graph actually exposes, routes statement/balance/period questions to `live-financial-statement`, `financial-statement-analysis`, `build-fact-grid`, and the fiscal-calendar/close-status/sync-status reads before raw Cypher; the workflow's first move reflects that. `hit_credit_ceiling` joins the result metadata and lowers confidence like the step cap.
- **`robosystems/operations/operators/tool_loop.py`** — `max_credits`: a soft ceiling checked between model calls against `TrackedAIClient.total_credits`; once reached, no further tool turn starts and the model is nudged (with credit-limit wording) to answer from what it has. The first call always runs — its cost is unknowable up front and the pre-flight already gates the run. The nudge path now also handles a trailing string-content user turn.
- **`robosystems/models/api/graphs/operator.py` / `routers/graphs/operator/execute.py`** — `max_credits` on `OperatorRequest` (positive, capped at 100k), folded into the operator context on both execution paths.

## Breaking Changes

None. `max_credits` is additive and optional — an SDK regen opportunity, generated tier. Read-only MCP viewers on roboledger tenants will see the nine read tools appear; no previously available tool moves or changes shape.

## Testing

- `just test` — full unit suite passed (14,060 passed, 43 skipped).
- New/updated tests: the manager read/write split on read-only tenants and its shared-repo exclusions (two gating tests that pinned the old bundling were deliberately flipped); curated-tool routing built only from exposed tools; ceiling stop between calls with the credit-limit nudge; first-call-always-runs; tenant-supplied garbage `max_credits` ignored; request-model bounds.

## Certification

### DIFF
--- a/robosystems/middleware/mcp/tools/manager.py
+++ b/robosystems/middleware/mcp/tools/manager.py
@@ -202,9 +202,10 @@ class GraphMCPTools:
       self.financial_statement_analysis_tool = FinancialStatementAnalysisTool(
         graph_client
       )
-      # OLTP-backed live statement — tenant entity graphs only. Skipped
-      # on shared repos (no OLTP tenant schema) and on read-only graphs.
-      if not self._is_shared_repository() and not read_only:
+      # OLTP-backed live statement — tenant entity graphs only. Skipped on
+      # shared repos (no OLTP tenant schema). A pure read, so it stays
+      # available on read-only surfaces (graph viewers, read-only operators).
+      if not self._is_shared_repository():
         self.live_financial_statement_tool = LiveFinancialStatementTool(graph_client)
 
       # Semantic enrichment tools (roboledger + manifest flag)
@@ -254,6 +255,8 @@ class GraphMCPTools:
     # Layer 2: Materialization — `materialize` + `get-graph-sync-status`.
     # User entity graphs only (shared repos use their own pipeline and
     # don't track staleness). `materialize` mirrors the REST body shape.
+    # The pair splits on read_only: sync status is a pure read and stays
+    # available on read-only surfaces; materialize is the write half.
     self.get_graph_sync_status_tool = None
     self.materialize_tool = None
     if (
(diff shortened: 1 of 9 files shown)


jfrench9:
@claude please review this PR

PR #1291, opened 2026-08-28
feat(operators): worker execution home, pause/resume, Sonnet 5 default

jfrench9:
## Summary

Makes the background worker the execution home for every AI Operator run, adds the `awaiting_input` pause state a human-gated run needs, exposes queue position and scales the worker on real backlog, and flips the default Bedrock model to Claude Sonnet 5 now that the account's Marketplace agreement is in place. This is Phase 3 of the operator console hardening: the console's operator path ran synchronously inside the API process (or on a FastAPI `BackgroundTasks` runner that lost results), while `OperatorWorkerTask` sat registered with no producer.

## Changes

**Operators run on the worker** (`routers/graphs/operator/execute.py`, `operations/operators/adapters/worker.py`, `progress.py`)
- Both `POST /v1/graphs/{graph_id}/operator[/{operator_type}]` endpoints now gate the request (lifecycle, repository limits, write role, graph scope, credits — a 403/402 comes back on the request, not as a failed operation), enqueue an `operator` task, and answer **202** with the operation's `_links` (stream/status/cancel). `?mode=sync` waits up to 50s (under the ALB idle timeout) and answers 200 with the result when it lands in time, 202 otherwise; a failed or cancelled run under sync answers 200 with `error_details`, the shape the console renders.
- Deleted the three in-process paths and their strategy selector: the sync orchestrator path, the SSE path that buffered progress until the end, and the `BackgroundTasks` runner whose broadcast-only events left `/status` pending forever.
- The worker adapter carries the request's `history` and `context` (so `max_credits` reaches the loop and `mapping_id` still resolves) and returns the response envelope (`content`, `operator_used`, `mode_used`, `metadata`, `tokens_used`, `confidence_score`, `execution_time`) that `/status` returns and the SSE `operation_completed` event carries. The operator's own metadata keys are still merged flat for the mapping operation's consumers. Progress events carry `progress_percentage` alongside `progress_percent` (the name the TypeScript facade reads).
- Reviewers: `_dispatch` and `_wait_for_operation` in `execute.py` are the whole new contract; the worker re-runs the three gates with its own session before the first model call.

**Queue visibility and backlog scaling** (`worker/consumer.py`, `worker/client.py`, `worker/metrics.py`, `routers/operations.py`, `middleware/sse/operation_manager.py`, `cloudformation/worker.yaml`)
- The consumer marks an operation RUNNING on pickup; until now nothing distinguished "queued" from "executing".
- `GET /v1/operations/{id}/status` on a pending operation reports `queue_position` (1-based) and `queue_depth`.
- The metrics thread publishes `InFlight` (sum of the per-worker inflight lists) and `Backlog` (queued + in flight) beside `QueueDepth`; both scaling alarms read `Backlog`, so two tasks waiting behind one busy worker scale out (they never crossed the threshold on queue depth alone).

**Pause state** (`middleware/sse/event_storage.py`, `operation_manager.py`, `streaming.py`, `worker/tasks/base.py`, `worker/consumer.py`, `routers/operations.py`, `models/api/operations.py`)
- `OperationStatus.AWAITING_INPUT` with two events (`operation_awaiting_input`, `operation_resumed`); same priority rung as RUNNING so a run can pause and resume, terminal statuses still win. `OperationMetadata` gains `input_request`.
- `BaseTask.pause_for_input(prompt, checkpoint, details)` records the prompt, the checkpoint and the task's own queue payload on the operation and raises `TaskPaused`; the consumer neither completes nor fails a paused task. `BaseTask.resume` reads the answer back.
- `POST /v1/operations/{id}/resume` (new, owner-only, 409 unless paused) records the answer, moves the operation back to RUNNING and re-enqueues the same operation id with `params["resume"] = {checkpoint, input}`. `/status` exposes the prompt and details (never the checkpoint) plus a `resume` link; a paused operation's stream replays and closes.
- No task pauses yet; the primitive is there for the operators that will stop before an irreversible step.

**Sonnet 5 default** (`config/operators.py`, `cloudformation/api.yaml`)
- `DEFAULT_MODEL_CONFIG` → `SONNET_5`, fallback → `SONNET_4_6`. The family-aware request shaping and the rate-card entry landed in #1288; the Marketplace agreement (product `prod-4ezhkeia6k2cs`) is in place and the model was verified through `AIClient` locally. The API task role's Marketplace subscribe permission gains the Sonnet 5 product ID.

**Docs**: `worker/README.md` (producer, pause-for-input, backlog scaling) and `operations/operators/README.md` (execution paths).

## Breaking Changes

None on the stable SDK tier: the TypeScript `OperatorClient` facade already handles both the 200 and the 202 + stream shapes, so the console and facade callers work unchanged (the console gains real progress). Generated-tier behavior change, for the release notes: the operator POST endpoints now answer **202 by default** where most callers previously got a 200 — a raw generated-client caller that reads `content` off the response should pass `?mode=sync` or follow `_links.stream`. Additive: `POST /v1/operations/{id}/resume`, the `queue_position`/`queue_depth`/`input_request` fields and `resume` link on `/status`, `OperatorResponse.operation_id` now populated, and the `awaiting_input` status value (non-terminal; clients that switch on status should not treat it as done). Worth an SDK regen at the next minor.

Deploy notes: no migrations. `worker.yaml`'s alarms read the new `Backlog` metric; workers still on the old image publish only `QueueDepth`, and the scale-out alarm treats missing data as breaching, so a brief spurious scale-out during the rollout is possible and harmless.

## Testing

- `just test-code` (ruff, format, basedpyright, cf-lint): passed.
- `just test` (full unit suite): 4,565 passed, 1 failed — `test_ai_client.py` asserted a `temperature` on the default request, which Sonnet 5's family gate omits. Fixed in the last commit together with a sibling test; the file re-run is green (42 passed). Targeted suites re-run after each change: `tests/worker`, `tests/routers/test_operations.py`, `tests/middleware/sse`, `tests/routers/graphs/test_operator_router.py`, `tests/routers/graphs/test_operator_subgraph_integration.py`, `tests/operations/operators`, `tests/routers/graphs/test_feature_flags.py`, `tests/security/test_graph_access_control.py` — all green.
- Sonnet 5 exercised end-to-end against Bedrock from the local stack through the app's own `AIClient` and a full operator question on the local demo graph before the flip (see the model-upgrade spec); the worker path itself is exercised by the existing `auto-map-elements` producer and the new router/adapter tests, not by a live run in this session.

## Certification


jfrench9:
@claude please review this PR

jfrench9:
Addressed all three review findings in `528a0028`:

1. **Resume ordering** — `requeue_task` now runs before `resume_operation`. A failed push leaves the operation `awaiting_input` with its resume link intact (test: `TestResumeOrdering`).
2. **Stale `input_request` after a pickup** — `_apply_input_request` clears it on `OPERATION_STARTED` as well as `OPERATION_RESUMED` (test added).
3. **`OperatorWorkerTask` missing `operator_type`** — raises `ValueError` so the consumer records FAILED rather than a completed result with no content (test added).

Also fixed the CI failure: the subgraph integration tests left operator selection to the real registry, whose contents in a given xdist worker depend on import order — an empty registry answered 404. Selection is now mocked there like in the router tests.

PR #1294, opened 2026-08-28
fix(operators): a cancelled run stops calling the model

jfrench9:
## Summary

Found in the post-hotfix prod walk of the worker execution home (#1291/#1293): cancelling an operator run (`DELETE /v1/operations/{id}`) flipped the status to `cancelled` for the client, but the worker kept running the tool loop to completion — 22 more seconds and seven more billed model calls on the cancelled question. The Cypher loop never consulted cancellation; only the MappingOperator did. This closes that, plus the queued case.

## Changes

- `operations/operators/tool_loop.py`: `run_tool_loop` checks `ctx.progress.is_cancelled()` before every model call — including the first, so a run cancelled while it waited in the queue spends nothing — and returns a `cancelled=True` result without the wrap-up ("answer now") call. `ToolLoopResult.cancelled` added.
- `operations/operators/implementations/cypher.py`: `cancelled` in the result metadata; confidence 0.0 for a cancelled run.
- `worker/consumer.py`: a task whose operation is already `cancelled` at pickup is skipped (removed from inflight, never executed) rather than run into a result the status ladder refuses to record anyway.
- Tests: loop cancels between calls with no nudge; a run cancelled while queued makes zero calls; the consumer skips a pre-cancelled task.

On the API path `CallbackProgress.is_cancelled` is always `False`, so nothing changes for the orchestrator/test contexts.

## Breaking Changes

None. Additive `cancelled` key in the Cypher operator's result metadata.

## Testing

- `tests/operations/operators/test_tool_loop.py`, `tests/worker/test_consumer.py`, `tests/operations/operators/test_cypher_orientation.py`, `tests/operations/operators/test_unified_operator.py`: 53 passed; ruff/format/basedpyright clean on the touched files.
- Reproduced on prod before the fix (op `op_01M13CYZ75WYAMZ3YN2KF7N0AA`: cancelled at +4s, worker completed at +22s with 7 billed calls); not re-run on prod — this rides the next release, the console has no cancel control for operator runs yet.

## Certification

### DIFF
--- a/robosystems/operations/operators/implementations/cypher.py
+++ b/robosystems/operations/operators/implementations/cypher.py
@@ -227,6 +227,7 @@ class CypherOperator(Operator):
         "result_count": len(rows),
         "hit_step_limit": result.hit_cap,
         "hit_credit_ceiling": result.hit_credit_ceiling,
+        "cancelled": result.cancelled,
         "loop_iterations": result.iterations,
       },
       # De-dupe preserving order (a tool may be called several times).
@@ -431,6 +432,8 @@ EXAMPLE QUERIES (working patterns for this graph — copy and adapt rather than 
     return value if value > 0 else None
 
   def _calculate_confidence(self, result: ToolLoopResult) -> float:
+    if result.cancelled:
+      return 0.0
     if result.hit_cap or result.hit_credit_ceiling:
       return 0.5
     if result.rows:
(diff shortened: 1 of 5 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Closed the gap in `2c353d19`: the wrap-up call now sits behind the same `is_cancelled` guard as the in-loop call, with a test that reaches the cap with a cancel pending and asserts no second model call.

jfrench9:
Correction: that commit landed after this PR merged, so it is in #1295 instead.

PR #1295, opened 2026-08-28
fix(operators): guard the wrap-up call against a cancel at the cap

jfrench9:
## Summary

Follow-up to #1294 (merged before this landed on it). The review there found the one call the cancel guard didn't cover: after the loop exits via the iteration cap or the credit ceiling, the "answer now" wrap-up call ran unconditionally — so a cancel landing exactly as a run hit its cap still billed one more model call. The wrap-up now sits behind the same `is_cancelled` check and early return as the in-loop call.

## Changes

- `operations/operators/tool_loop.py`: cancel check before the post-loop wrap-up call; returns `cancelled=True` with the rows gathered so far and no further model call.
- `tests/operations/operators/test_tool_loop.py`: reaches the cap with a cancel pending and asserts a single model call.

## Breaking Changes

None.

## Testing

`tests/operations/operators/test_tool_loop.py`: 16 passed; ruff/format/basedpyright clean.

## Certification

### DIFF
--- a/robosystems/operations/operators/tool_loop.py
+++ b/robosystems/operations/operators/tool_loop.py
@@ -277,11 +277,25 @@ async def run_tool_loop(
     else:
       error_retries += 1
 
-  # Iteration cap or credit ceiling reached. Nudge for a final answer,
-  # appending the nudge to the trailing user turn (a second consecutive user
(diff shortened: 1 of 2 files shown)


jfrench9:
@claude please review this PR

PR #1296, opened 2026-08-28
fix(reports): statements stop rendering the cash-flow delta basis

jfrench9:
## Summary

Four correctness fixes on the statement and fact-grid read surfaces, all of the same kind: a number that looked authoritative and wasn't. The largest is the comparative column of every live cash flow statement — the indirect method derives working-capital deltas and the cash tie from period-over-period balance changes, so the earliest period of a cash flow pivot is the delta basis, not a statement. It rendered net income and add-backs with no deltas and no cash reconciliation, footed, and was wrong (on a live tenant the comparative month showed more than twice the balance-sheet cash movement). The close-time stamp already keeps only the close month; the live and saved-report read paths now apply the same rule.

## Changes

**Cash flow renders the current period only** (`operations/roboledger/reads/reports.py`)
- New `rendered_period_indexes(statement_type, periods)`: every period renders except the earliest on a `cash_flow_statement` when two or more were pivoted. Chosen by date, not position, so multi-period saved reports in authored order drop the right column.
- `get_live_financial_statement` still pivots `[Current, Prior]` (the prior month is the delta basis) but renders one column for cash flow; income statement and balance sheet are unchanged. The all-zero filter runs on the rendered values, so a row non-zero only in the basis column no longer appears.
- `get_statement` (saved reports) applies the same rule after `build_periods`, so a `comparative=True` report's cash flow renders one column. The persisted prior-period facts are untouched.
- `LiveFinancialStatementResponse.periods` documents the shape. The roboledger-app statements page maps `periods` and `row.values` generically, so a one-column cash flow renders without a client change.

**`limit` defaults to the whole statement** (`models/api/extensions/reports.py`, `middleware/mcp/tools/financial_statement_tools.py`)
- `LiveFinancialStatementRequest.limit` and both MCP financial-statement tools default to the 1000 ceiling instead of 50. A real chart of accounts exceeds 50 non-zero rows routinely, and a cut statement's visible rows stop footing to its subtotals; the app sends no `limit`.

**`equity_statement` leaves the MCP surface** (`reads/reports.py`, `financial_statement_tools.py`)
- New `MCP_LIVE_STATEMENT_TYPES` (the four minus `equity_statement`) drives the tool's enum and validation. It is still served on REST for the app's Statement of Equity tab. The live path renders equity balances, not a rollforward, and a model reading two rows under a passing validation presents them as the statement; the tool now refuses it with that reason and points at `balance_sheet`.
- The tool description also stops claiming subtotal rows are filtered out — they are kept and flagged `is_subtotal` — and states that cash flow renders the current period only.

**XBRL download declares what it omits** (`models/api/extensions/reports.py`, `reads/reports.py`)
- `ReportBundleDownloadResponse.omitted_content: list[str]` (default `[]`). The XBRL 2.1 presign sets `["disclosure_notes"]`: the emitter strips tenant-authored notes on every generation (`xbrl_21._strip_disclosure_content`), so the file was valid and silently incomplete next to the notes shown on screen. Declared on cache hits too, since it is a property of the flavor.

**Fact-grid summaries stop summing balances across time** (`operations/roboledger/views/fact_grid_builder.py`, `models/api/views/view_response.py`)
- `summarize_by_element` emitted `total` and `average` for instants — its own docstring conceded that summing a balance across periods is not a balance, and both the REST view and the MCP tool surfaced the figure unqualified. Instants (no `period_start`, or an explicit `period_type` of `instant` — verified against how the SEC graph stores them) now carry `count` / `min` / `max` only. `ElementSummary.total` / `.average` become optional to match.

**Tests**
- New: `rendered_period_indexes` (date-not-position, single period, other statement types untouched), live and saved cash flow render one column while both periods reach the generator, instants omit aggregates, MCP enum/limit/equity refusal, `omitted_content` on both flavors and on a cache hit.
- Corrected rather than preserved: two fixtures that built duration facts without a `period_start`, and one that asserted a `total` on `Assets`.

Reviewers: the one judgment call is `rendered_period_indexes` on `get_statement` — it trims the rendered periods, not the persisted facts, so an XBRL/JSON-LD export of a comparative report still carries the prior period's authored facts (net income etc. are legitimate facts on their own). A correct comparative cash flow column would need a three-period pivot; that is a follow-up, not this PR.

## Breaking Changes

None to the stable tier (facades, integration-template emit path).

Generated-tier changes, riding a client minor (SDK regen opportunity — all additive or loosening):
- `LiveFinancialStatementRequest.limit` default 50 → 1000 (range unchanged).
- `LiveFinancialStatementResponse.periods` is one entry for `cash_flow_statement` (two for the others, as before).
- `ElementSummary.total` / `average` become nullable.
- `ReportBundleDownloadResponse.omitted_content` added.
- MCP `live-financial-statement`: `statement_type` enum drops `equity_statement`; `limit` default 1000. Directory-listing note: the tool list is unchanged, so no resubmission on the channels that key on it; the OpenAI portal treats a definition change as a new version, and its review is still pending, so the definitions are not yet locked.

## Testing

- `uv run pytest tests/operations/roboledger tests/middleware/mcp tests/routers/extensions tests/graphql tests/models tests/operations/serialization` — 3,162 passed.
- `just test-code` — ruff, format, basedpyright all clean.
- Full `just test-all` not run in-session (the local run halts on a known test-order pyarrow issue unrelated to this change); CI runs the full suite.

## Certification

### DIFF
--- a/robosystems/middleware/mcp/tools/fact_grid_tool.py
+++ b/robosystems/middleware/mcp/tools/fact_grid_tool.py
@@ -40,7 +40,7 @@ class BuildFactGridTool:
   def get_tool_definition(self) -> dict[str, Any]:
     return {
       "name": "build-fact-grid",
-      "description": """Construct multidimensional fact grid from graph data. Retrieves facts based on elements, periods, and optional dimensions. Returns structured data with element names, values, and periods. Use include_summary=true to add aggregated statistics (count, total, avg, min, max) by element.
+      "description": """Construct multidimensional fact grid from graph data. Retrieves facts based on elements, periods, and optional dimensions. Returns structured data with element names, values, and periods. Use include_summary=true to add per-element statistics: count, min, max for every element, plus total and average for duration elements only — a balance summed or averaged across periods is not a balance, so instants (e.g. total_assets) omit both.
 
 **WHEN TO USE:**
 - When the user asks to compare specific financial metrics across periods or companies
@@ -127,7 +127,7 @@ For income statement items (revenue, net income), always specify period_type='an
           },
           "include_summary": {
             "type": "boolean",
-            "description": "Include summary statistics (count, total, avg, min, max) by element",
+            "description": "Include per-element statistics: count, min, max for every element; total and average for duration elements only (instants omit them — a balance summed across periods is not a balance).",
             "default": False,
           },
           "limit": {
--- a/robosystems/middleware/mcp/tools/financial_statement_tools.py
+++ b/robosystems/middleware/mcp/tools/financial_statement_tools.py
@@ -24,7 +24,7 @@ from robosystems.db.extensions import extensions_session
 from robosystems.middleware.operations import run_off_loop
 from robosystems.operations.roboledger.reads.reports import (
   ANALYSIS_STATEMENT_TYPES,
-  LIVE_STATEMENT_TYPES,
+  MCP_LIVE_STATEMENT_TYPES,
   CoaMappingNotFoundError,
   get_live_financial_statement,
   resolve_reporting_window,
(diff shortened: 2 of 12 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Addressed the one gap from the review in 98986ad1: `build-fact-grid`'s tool description (`fact_grid_tool.py`) and `ViewResponse.summary` now say `total`/`average` are duration-only, and `test_include_summary_omits_total_for_instants` pins the tool surface for an instant element.

PR #1297, opened 2026-08-28
feat(alarms): page when any Dagster run fails

jfrench9:
## Summary

Nothing alarms when Dagster work fails to complete, and the Dagster stack's own alarms have never existed. This PR fixes both: `deploy-dagster.yml` now passes `SNSAlertEmail` to the Dagster stack (it was only ever passed to the worker stack, so the SNS topic and every `HasAlertEmail` alarm — including the stalled-provisioning write-off alarm — were never created), and a new `run_status_sensor` publishes a `RunFailure` CloudWatch metric for every failed run in any job, with one alarm on it.

## Changes

**Deploy workflow (`.github/workflows/deploy-dagster.yml`)**
- Add `ParameterKey=SNSAlertEmail` to the Dagster stack's parameter block, reading the same `aws_sns_alert_email` input the worker block already uses. Widen the input's description to cover both stacks.

**CloudFormation (`cloudformation/dagster.yaml`)**
- `DagsterTaskRole`: the `cloudwatch:PutMetricData` namespace condition becomes a list — `RoboSystems/Graph/${Environment}` (unchanged) plus `RoboSystems/Dagster/${Environment}`, the namespace the stack's two existing metric-filter detectors already alarm on.
- New `DagsterRunFailureAlarm` (`robosystems-dagster-${Environment}-run-failure`): `Sum` of `RunFailure` over 300s, `> 0`, `TreatMissingData: notBreaching`, action `DagsterAlertTopic`. Same shape as the neighbouring Dagster alarms and the worker DLQ-depth alarm.
- `SNSAlertEmail` parameter description lists what it now covers.

**Sensor (`robosystems/dagster/sensors/run_failure.py`, registered in `definitions.py`)**
- `run_failure_metric_sensor`: `@run_status_sensor(run_status=FAILURE, monitor_all_code_locations=True, default_status=RUNNING)`. Logs `DAGSTER RUN FAILED job=… run_id=…` (module logger → daemon CloudWatch stream; `context.log` → the tick log in the UI), then publishes two data points in one `put_metric_data`: a dimensionless `RunFailure=1` for the alarm, and a copy with a `Job` dimension for triage in the console. CloudWatch never aggregates a custom metric across dimensions, which is why the alarm reads the dimensionless one.
- Publishes only when `env.is_aws_environment()`; any publish error is logged at warning and never raised, so the publisher's own failure cannot turn one failed run into a second failure.
- Defaults to `RUNNING` — a detector that ships stopped is the gap this closes. Run-status sensors start from the cursor at enable time, so the deploy does not replay historical failures.
- `run_monitoring` is already enabled in `dagster_home/dagster_prod.yaml`, so a run whose Fargate Spot task is reclaimed does transition to FAILURE and is covered.

**Docs**: `robosystems/dagster/README.md` sensor table gains the row.

**Reviewer notes**
- Adding the parameter creates `DagsterAlertTopic` with an email subscription on the next Dagster deploy; the confirmation email must be clicked before anything delivers (see below).
- `monitor_all_code_locations=True` rather than `monitored_jobs=[…]`: the point is every job, including the SEC adapter's, and the list would rot.

## Breaking Changes

None. No API surface changes; nothing for the SDKs to regenerate.

## Testing

- `just cf-lint dagster` — clean.
- `uv run pytest tests/dagster` — 271 passed, including the five new tests in `tests/dagster/test_run_failure_sensor.py` (registered + `RUNNING` + all code locations; alarm point dimensionless, triage point carries the job; publishes to `RoboSystems/Dagster/prod`; no CloudWatch call outside AWS; a publish error never fails the tick).
- Full `just test-all` unit run: not run in this session.

**Post-deploy verification (owner, after the next Dagster deploy)**
1. Confirm the SNS subscription email for `robosystems-dagster-prod-alerts` — until then the topic delivers nothing.
2. Prove the delivery path with one forced transition: `aws cloudwatch set-alarm-state --alarm-name robosystems-dagster-prod-run-failure --state-value ALARM --state-reason "delivery test"`. A green alarm proves it exists; only a transition proves it works.
3. Prove the sensor→metric path once with a real failed run (any job launched with a bad partition/config that fails at execution), then check `RunFailure` in `RoboSystems/Dagster/prod`.

## Certification

### DIFF
--- a/.github/workflows/deploy-dagster.yml
+++ b/.github/workflows/deploy-dagster.yml
@@ -148,7 +148,7 @@ on:
         type: string
         default: "false"
       aws_sns_alert_email:
-        description: "Email for worker alerts (e.g. DLQ depth); empty disables alarm notifications"
+        description: "Email for Dagster and worker alerts (run failures, DLQ depth, backup gaps, stalled-provisioning write-offs); empty disables alarm notifications"
         required: false
         type: string
         default: ""
@@ -347,6 +347,7 @@ jobs:
                 ParameterKey=BastionSecurityGroupId,ParameterValue=${{ inputs.bastion_sg_id }} \
                 ParameterKey=ApiSecurityGroupId,ParameterValue=${{ steps.resolve-api-sg.outputs.api_sg_id }} \
                 ParameterKey=ContainerInsightsEnabled,ParameterValue=${{ inputs.container_insights_enabled }} \
+                ParameterKey=SNSAlertEmail,ParameterValue=\"${{ inputs.aws_sns_alert_email }}\" \
                 ParameterKey=SharedRawBucketArn,ParameterValue=\"${{ inputs.shared_raw_bucket_arn }}\" \
                 ParameterKey=SharedProcessedBucketArn,ParameterValue=\"${{ inputs.shared_processed_bucket_arn }}\" \
                 ParameterKey=DeploymentBucketArn,ParameterValue=\"${{ inputs.deployment_bucket_arn }}\" \
--- a/cloudformation/dagster.yaml
+++ b/cloudformation/dagster.yaml
@@ -14,7 +14,7 @@ Parameters:
   SNSAlertEmail:
     Type: String
     Default: ""
-    Description: Email for Dagster alerts (e.g. backup coverage gaps); empty disables alarm notifications
+    Description: Email for Dagster alerts (run failures, backup coverage gaps, stalled-provisioning write-offs); empty disables alarm notifications
 
   # Infrastructure & Networking
   VpcId:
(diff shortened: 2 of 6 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Took the one nit: the doubled blank line before `Outputs:` is gone in 8b617a7d.

jfrench9:
CI's `test / test` failure was test isolation in `test_run_failure_sensor.py`, not the sensor: under `ENVIRONMENT=test` an earlier test's monkeypatch leaves an instance attribute on the shared `env`, so patching the class attribute flipped `is_aws_environment()` but not the namespace read. 36512f81 patches the sensor module's `env` binding with an `EnvConfig` subclass instead; reproduced the CI ordering locally (`tests/config/test_graph_tier_config.py` before the sensor tests) and it passes.

PR #1300, opened 2026-08-28
chore(commands): add /hotfix as the explicit trigger for the two-PR release flow

jfrench9:
## Summary

Adds `/hotfix`, the explicit trigger for shipping a fix to production ahead of the next release via the two-PR pattern (a reviewed PR to `main` plus a same-commits twin into the current `release/<x.y.z>` branch). The procedure previously existed only as prose guidance, which is easy to pattern-match into from an urgency word; a hotfix is a deploy-timing decision with a real cost, so it should only ever start from the user invoking the command. The file says so, and lists what does not count as an invocation.

## Changes

- `.claude/commands/hotfix.md` — new command: preflight (resolve the highest `origin/release/*`, confirm the defect is actually on that branch, security-disclosure sequencing), the main PR via `/create-pr`, the twin via `just create-feature hotfix … release/x.y.z` + ordered cherry-pick (stop on conflict), the twin opened without the `@claude` review request (the one deliberate exception to `/create-pr` §5 — identical, already-reviewed commits; its job is CI on the deploy ref), and a hand-off that leaves merge, deploy and branch deletion to the user.

## Breaking Changes

None

## Testing

Prose-only change; nothing to run. Structure mirrors the existing commands (`create-pr`, `deploy`), and the flow matches the two-PR pattern already used on this repository.

## Certification

### DIFF
--- a/.claude/commands/hotfix.md
+++ b/.claude/commands/hotfix.md
@@ -1,0 +1,83 @@
+---
+description: Ship a fix to production ahead of the next release — the two-PR pattern (main + release-branch twin). Runs only when the user invokes it explicitly.
+argument-hint: "<name> [release/x.y.z] [main-pr-number-or-branch]"
+---
+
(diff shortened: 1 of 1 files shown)


jfrench9:
@claude please review this PR

PR #1301, opened 2026-08-28
fix(mcp): keep the authored tool descriptions; carry the SEC query rules on the surface

jfrench9:
## Summary

`MCPHandler.get_tools()` overwrote the authored descriptions of `read-graph-cypher` and `get-graph-schema` with a single scope sentence on every graph, shared and tenant, discarding the WHEN-TO-USE / NOTES / worked-pattern guidance each client writes its Cypher from. Reproduced live on `sec`: a textbook query written from the degraded description returned eight rows all stamped the same date — segment facts, restatements and cumulative figures blended into what reads as a quarterly revenue trend. The guidance that fixes it lived only in `get-example-queries` (pull-only) and the client-side `sec-filing-analysis` skill, which reaches only our own Claude; on ChatGPT, Cursor, Docker, VS Code, Glama and Grok there is no skill. The tool description is the one surface every client reads, so this PR keeps the authored text and puts the SEC rules on it.

## Changes

**`robosystems/routers/graphs/mcp/handlers.py`**
- `get_tools()` prepends a one-line graph scope (`**GRAPH:** shared repository `sec` — public, read-only data.` / `**GRAPH:** private graph `kg…`.`) to the two core tools instead of replacing their descriptions. Tool dicts are copied before mutation.
- For a shared repository, the manifest's new `cypher_query_guidance` is appended to `read-graph-cypher`'s description (resolved the same way `get_instructions()` resolves `agent_instructions`).
- `get-graph-info` gets a real description: what it returns (graph id, approximate node count, node labels, relationship types, read-only flag), when to use it, no arguments. Schema unchanged.

**`robosystems/adapters/base.py`** — `SharedRepositoryManifest.cypher_query_guidance: str | None = None`, with a comment on why the tool description (not `agent_instructions`, not a skill) is where this knowledge has to live.

**`robosystems/adapters/sec/manifest.py`**
- `cypher_query_guidance`: the six rules — `has_dimensions: false` for consolidated totals; `Element.canonical_concept` over a single qname (with the `us-gaap:Revenues` vs `…ExcludingAssessedTax` example and the `resolve-element` pointer); pinned period shape (`duration_type` for flows, `period_type: 'instant'` for balances — quarterly/YTD/annual share end dates); anchor on Entity/Report first, `Structure`/`Element`/`Period` last, always `LIMIT`; `RETURN DISTINCT` because restated facts sit in more than one FactSet; run `get-example-queries` first. Wording mirrors the skill's "Rules of the road" so the server and the skill teach the same thing.
- `agent_instructions` NOTES gains a one-line pointer to those rules.

**Docs**: `middleware/mcp/README.md` paragraph on the scope-prepend rule and `cypher_query_guidance`.

**Tests** (`tests/routers/graphs/test_mcp_handlers.py`): the authored text survives on a tenant graph and carries no SEC rules; on `sec` the scope line, the authored text and each rule are present, and the rules are on the Cypher tool only; two `get_tools()` calls yield identical output and leave the manager's definitions untouched.

**Reviewer note — directory listings.** No tool is added or removed (the `sec` surface stays at 10 tools, the invariant the open reviews captured). OpenAI treats a definition change as a new version, but the v1.0.0 review is still pending, so there is no published definition to diverge from; the reviewer sees this text. Noted in the listings runbook.

## Breaking Changes

None. No REST/GraphQL/operation surface changes; the manifest field is additive with a `None` default.

## Testing

- `uv run pytest tests/routers/graphs/test_mcp_handlers.py tests/routers/graphs/mcp tests/middleware/mcp tests/adapters/sec tests/config` — 2,769 passed, 13 skipped, including the three new tests.
- Full `just test-all` unit run: not run in this session.

## Certification

### DIFF
--- a/robosystems/adapters/base.py
+++ b/robosystems/adapters/base.py
@@ -46,6 +46,13 @@ class SharedRepositoryManifest:
   # falls back to generating instructions from the graph's live tool surface.
   agent_instructions: str | None = None
 
+  # Query guidance appended to `read-graph-cypher`'s tool description for this
+  # repository: the data-model rules a raw Cypher query must honour to return
+  # a correct number (consolidation, period shape, deduplication). The tool
+  # description is the one place every MCP client reads — `agent_instructions`
+  # is the routing layer, and client-side skills reach only our own Claude.
+  cypher_query_guidance: str | None = None
+
   # Status
   status: str = "available"  # available, coming_soon
 
--- a/robosystems/adapters/sec/manifest.py
+++ b/robosystems/adapters/sec/manifest.py
@@ -44,8 +44,37 @@ SEC_MANIFEST = SharedRepositoryManifest(
     # not served: the replica stack carries `sec` only, master reads are off,
     # and the master is parked at desired 0, so every read for it fails. Do not
     # restore this line until a read actually succeeds against that graph.
+    "- Raw Cypher on this graph has four rules that decide whether a number is "
+    "right: `Fact {has_dimensions: false}` for consolidated totals, "
+    "`Element.canonical_concept` over a single qname, a pinned `Period` shape "
+    "(`duration_type` for flows, `period_type: 'instant'` for balances), and "
+    "`RETURN DISTINCT` anchored on the Entity. `read-graph-cypher`'s "
+    "description spells them out.\n"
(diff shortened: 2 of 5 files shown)


jfrench9:
@claude please review this PR

PR #1302, opened 2026-08-28
feat(sec): publish the SEC dump to Hugging Face and add just sec-dump

jfrench9:
## Summary

Publishes the SEC shared-repository snapshot as a public Hugging Face dataset ([robosystems/sec-xbrl-knowledge-graphs](https://huggingface.co/datasets/robosystems/sec-xbrl-knowledge-graphs)) and gives a checkout a one-command way to pull it. Two halves: a **manual-only Dagster job** that copies the existing R2 artifact (`sec.lbug.zst`) to the Hub without the bytes ever leaving AWS a second time, and **`just sec-dump`**, which downloads, verifies, and decompresses the file into `data/lbug-dbs/` so the local stack serves the full corpus with no pipeline run. The first snapshot and dataset card are already live (uploaded from a workstation); this PR is the repeatable path.

## Changes

**Publish job (Dagster)**
- `dagster/assets/shared_repositories/huggingface.py` — `publish_to_huggingface(context, graph_id, repo_id, path_in_repo, …)`, generic over shared repositories like its R2/S3 siblings. Presigns the R2 object (12 h), launches a **Hugging Face Job** (`cpu-basic` by default) whose container `curl`s the presigned URL and `hf upload`s to the dataset — so the copy runs on the Hub's compute and R2's free egress, never through the worker or the graph instance. Polls `inspect_job` to a terminal stage (hub 1.19 has no `wait_for_job`), fails loud with the job's log tail, verifies the published size against R2's `ContentLength`, and optionally prunes superseded LFS objects at the same path (`permanently_delete_lfs_files`, history rewritten) so storage stays one snapshot deep. Fixed `path_in_repo` keeps the download URL stable; the commit title carries the snapshot date and the pinned `ladybug` version.
- `adapters/sec/pipeline/hf_publish.py` — thin asset `sec_lbug_hf_published` (deps: `sec_lbug_r2_published`) with `SECHFPublishConfig` (`job_flavor`, `job_timeout`, `prune_previous`); `jobs.py` — `sec_lbug_hf_publish` job on the light ECS profile; registered in `__init__.py`. **Manual only** — no sensor or schedule launches it, matching the existing R2 publish, because the dump is deliberately monthly while the hosted graph stays nightly.
- `config/env.py` / `secrets_manager.py` / `.env.example` — `HF_TOKEN` (base secret, same slot as the `R2_*` keys) and `HF_SEC_DATASET_REPO` (defaults to the dataset above). The token and the presigned URL travel to the job as **secrets**, never plain env.

**Consumer path**
- `scripts/sec_dump.py` + `just sec-dump` — `hf_hub_download` (resumable, `HF_XET_HIGH_PERFORMANCE`), then streams the decompression through `zstandard` (checksum verified as it reads; no `zstd` binary needed), checks the `LBUG` magic and the frame's recorded size, writes to `.partial` and renames atomically, drops a stale `.wal`, deletes the archive (`--keep-archive` to keep it). Refuses to overwrite an existing `.lbug` without `--force`, checks free space before download and before decompression, and **warns when the engine that wrote the dump (read from the archive's commit title) differs from the `ladybug` version this checkout pins**.
- `justfile` — `sec-dump` is composed from `sec-dump-no-restart` (the script) and `graph-api-restart` (two guarded one-liners: restart the container only if it is running, then wait for `:8001/health`), because a running graph-api keeps serving the replaced file until restarted. Reviewers: `graph-api-restart` uses `docker ps --filter … -q` rather than a Go template on purpose — `just` only escapes the opening braces, so `{{{{.Names}}}}` renders as `{{.Names}}}}`.
- `README.md` / `CLAUDE.md` — pointers to the dataset and the recipe.

**Dependencies** — `huggingface-hub` (was only transitive via `fastembed`/`tokenizers`) and `zstandard` become declared dependencies.

## Breaking Changes

None. No API, GraphQL, or operations-envelope surface changes; nothing for the client SDKs to regenerate.

## Testing

- `just test-code` green on every commit (pre-commit hook: ruff, format, basedpyright, cfn-lint). Full `just test-all` unit run **not** executed in-session; the new modules' tests were run directly: `tests/adapters/sec/pipeline/test_hf_publish.py` (helper happy path with mocked R2/Hub, secrets-not-env, job failure with log tail, size mismatch, prune on/off, enum stage normalisation) and `tests/scripts/test_sec_dump.py` (round-trip, corrupt archive, wrong magic, size mismatch, `--force`/stale `.wal`/`--keep-archive`, engine-mismatch warning) — 40 tests, all passing, plus `test_init.py` counts updated.
- End to end on the published file: `just sec-dump --force` replaced a demo `sec.lbug` with the 128.5 GiB corpus (12 min download, 100 s decompress, exact byte count, `.wal` removed, archive deleted), after which the direct read and the local API both returned 8,444 filers; the refusal, `--no-restart`, and standalone restart branches were exercised against a small staged archive.
- The Hugging Face Job mechanism (`curl` → `hf upload` from inside Hub compute, secrets as env) was smoke-tested against a private scratch dataset, then deleted. The Dagster asset itself has not yet run in prod — it needs the `HF_TOKEN` secret (in place) and a deploy.

## Certification


jfrench9:
@claude please review this PR

jfrench9:
Addressed the two actionable observations in 2a082b55: the post-upload verification read (`_published_file`) now retries briefly so a lagging Hub listing can't fail a multi-hour run (still never prunes on a miss; tests cover both the lag and the exhausted-retries case), and the `SECHFPublishConfig` docstring no longer describes cpu-basic as disk-bound — the job filesystem observed in the smoke test was ~1.7 TB. The `CLAUDE.md` line is committed in 432bc49 and present in the branch diff; that was a checkout artifact on the review side.

PR #1303, opened 2026-08-28
fix(reports): validate every rendered column and the live statement; report inconclusive instead of a vacuous pass

jfrench9:
## Summary

Two gaps in what a green validation result actually meant:

- **The live statement ran no validation.** `validate_report` had two call sites — the saved-report read and the IB envelope — and neither is the surface the app renders or the MCP `live-financial-statement` tool returns. `LiveFinancialStatementResponse` had no `validation` field, so the accounting equation, net-income equation, totals footing, and the operating-plug warning were all off on the primary surface. Concretely: a cash flow whose reconciling plug on `IncreaseDecreaseInOtherOperatingCapitalNet` was larger than operating cash rendered with no warning.
- **`passed: true` overstated what was checked.** Every structural check read `values[0]` only, so a comparative statement was validated on one column — the oldest on the envelope path (periods sort ascending), `Current` on the saved-report path — and both reported one unqualified boolean. And a block type with no rules (`equity_statement`) returned `passed=True, checks=["no_validation_rules"]`, which the close UI renders as a green "Validation Passed" badge.

## Changes

- `guard_rails.py` — every check runs once per rendered column; on a multi-column statement each failure/warning is prefixed with the column it was found in (`[Prior] Balance sheet does not balance …`). Checks are recorded once; column-independent findings (a missing classification rollup) are stated once. A comparative column that is empty end to end is reported once by `comparative_data` instead of per section. `validate_report` takes optional `period_labels`; missing labels fall back to `column N`.
- `ValidationResult` / `ValidationCheckResponse` / `ValidationLite` gain `status`: `passed | failed | inconclusive`. A block type with no rules returns `inconclusive` with `passed=False` and a warning saying nothing was checked — not a vacuous pass.
- `get_live_financial_statement` validates exactly the rendered columns (the cash flow delta basis is projected out first) and carries the result as `LiveFinancialStatementResponse.validation`; the MCP tool's description tells the model to read it before quoting a number.
- Saved-report and envelope paths pass column labels (`Current`/`Prior`; period end date on the envelope).

## Client contract

Additive on the generated tier — new `validation` field on `LiveFinancialStatementResponse`, new `status` on `ValidationCheckResponse` / `ValidationLite` (REST + GraphQL). Rides the next client minor. The one behavior change consumers can see: `passed` is now `false` for an equity statement (nothing was checked). Follow-on in `roboledger-app`: read `status` and render a neutral "Not validated" badge for `inconclusive`, and show `validation` on the live statements page.

## Tests

- `test_guard_rails.py` — per-column failures name the column; single-column messages unchanged; checks recorded once; column-independent inconclusive stated once; empty prior column reported once; operating plug per column; `status` tracks failures; equity/unknown types are `inconclusive` (the old test asserted the vacuous pass and was rewritten).
- `test_reports_extra.py` — live path validates both rendered columns, passes a clean statement, validates only the rendered CF column, and surfaces a plug larger than operating cash.
- `test_financial_statement_tools.py` — the MCP payload carries `validation`.
- `test_statement_handlers.py` — envelope passes `status` through.

`just test-code` clean; `tests/operations`, `tests/graphql`, `tests/routers/extensions`, `tests/models`, `tests/middleware/mcp` green.

### DIFF
--- a/robosystems/middleware/mcp/tools/financial_statement_tools.py
+++ b/robosystems/middleware/mcp/tools/financial_statement_tools.py
@@ -62,6 +62,8 @@ class LiveFinancialStatementTool(BaseTool):
 
 **RETURNS:**
 Facts with element qnames, names, classifications, and values aligned with `periods` — current + prior (equal duration) for income_statement and balance_sheet, current only for cash_flow_statement. Abstract scaffolding rows and all-zero rows are filtered out; subtotals are kept and flagged `is_subtotal`. Capped at `limit` rows (default 1000 — the whole statement).
+
+`validation` is the guard-rail outcome for the rendered columns (`status`: passed | failed | inconclusive; `failures`; `warnings`). Read it before quoting a number: a `failed` status means the statement does not foot or balance, and a warning about 'Other operating capital, net' means the cash flow was made to foot by a reconciling plug — say so rather than presenting the figure as clean.
 """,
       "inputSchema": {
         "type": "object",
--- a/robosystems/models/api/extensions/reports.py
+++ b/robosystems/models/api/extensions/reports.py
@@ -253,9 +253,29 @@ class FactRowResponse(BaseModel):
 
 
 class ValidationCheckResponse(BaseModel):
-  """Aggregate result of running reporting rules over a structure."""
+  """Aggregate result of running reporting rules over a structure.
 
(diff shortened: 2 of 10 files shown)


jfrench9:
@claude can you review

PR #1305, opened 2026-08-28
fix(graph): refuse change-tier without target capacity; make the drain fail loudly

jfrench9:
## Summary

`change-tier` is the designated upsell route from Standard to Large/XLarge, and those tiers are provisioned on request (their writer ASGs sit at desired 0 by policy). The operation did not check that first: it moved the customer's Stripe price, set the subscription `upgrading`, and enqueued a migration whose worker raised the target ASG's desired capacity itself and then waited at most 300 s for a cold instance to boot, register, and claim the volume — a window a cold boot does not reliably fit, ending in a rollback that has only ever run in a test. Separately, the migration's drain step polled admin routes the graph API does not serve, timed out after 120 s, and detached the volume anyway.

This PR makes `change-tier` apply the same refuse-the-sale rule checkout already applies — a healthy writer on the target tier with a free slot, or a 409 before anything moves — and makes the drain refuse rather than time out into a detach.

## Changes

- **`operations/graph/capacity.py`** (new) — `tier_capacity_status(tier)`: `ready` when the tier has a healthy writer with a free slot; `scalable` and any lookup failure read as `at_capacity` (fail closed). Lifted verbatim from the checkout router so both sale paths share one definition.
- **`routers/billing/checkout.py`** — imports the shared helper under its existing private name; behaviour unchanged.
- **`operations/graph/commands/tier.py`** — after tier validation and before any write: `tier_capacity_status(new_tier) != "ready"` → `409` *"No capacity is currently available on the '<tier>' tier. Request access from the tier picker or contact support."* Nothing is committed, Stripe is not called, no task is enqueued.
- **`operations/graph/tasks/graph_tier_upgrade.py`** — `_drain_instance` now has three outcomes instead of one:
  - graph API unreachable → treated as drained (the maintenance-window procedure stops the container before a tier change, and every write reaches the volume through that API);
  - graph API up but `404` on `/admin/drain`, or any error status → `DrainRefusedError` before the detach;
  - drain timeout with connections still open → `DrainRefusedError` before the detach.
  The existing `except Exception` in `run()` catches it, rolls back the registry state (nothing has been detached), and re-raises. Module docstring updated to say so. Reviewers: the "unreachable counts as drained" branch is the judgment call — it is what lets the documented stop-the-container procedure work without a real drain endpoint.
- **Tests** — `tests/operations/graph/test_capacity.py` (new, 5); `test_tier_command.py` gains an autouse "capacity ready" fixture and a 409 test asserting no commit / no Stripe / no enqueue; `test_graph_tier_upgrade.py` gains three drain tests (unreachable, 404, timeout) and gives the existing happy-path stub a real status code.

## Breaking Changes

None. The only observable change is a new `409` from `POST /v1/graphs/{g}/operations/change-tier` when the target tier has no capacity; the operation envelope and request model are unchanged, so nothing in either SDK tier moves.

## Testing

- `just test` — 14,185 passed, 42 skipped.
- `just test-code` — clean (ruff, format, basedpyright, cfn-lint); also re-run by the pre-commit hook on both commits.
- Not exercised against a live fleet; the change-tier walk on a throwaway graph is the next step and is documented separately.

## Certification

### DIFF
--- a/robosystems/operations/graph/capacity.py
+++ b/robosystems/operations/graph/capacity.py
@@ -1,0 +1,35 @@
+"""Tier capacity as the sale paths see it.
+
+Both places the platform commits money against a writer slot — checkout and
+``change-tier`` — ask the same question first: is there a healthy writer on
+the target tier with a free slot *right now*? Nothing on either path raises
+desired capacity on purpose (the high tiers are provisioned on request), so
+``scalable`` is not good enough, and any failure to determine capacity reads
+as none: refuse the sale rather than collect against a slot that may not
+exist.
+"""
+
+from __future__ import annotations
+
+from robosystems.config import env
+from robosystems.logger import logger
+
+
+async def tier_capacity_status(tier: str) -> str:
+  """``ready`` when a writer for ``tier`` has a free slot; otherwise
+  ``at_capacity``. ``scalable`` (no slot, ASG below max) counts as
+  ``at_capacity`` because nothing on the sale paths raises desired capacity.
+  Any failure to determine capacity reads as ``at_capacity``.
+  """
+  try:
+    from robosystems.middleware.graph.allocation_manager import (
+      LadybugAllocationManager,
+    )
+    from robosystems.middleware.graph.types import GraphTier
+
+    manager = LadybugAllocationManager(environment=env.ENVIRONMENT)
+    status_value = await manager.check_tier_capacity(GraphTier(tier))
+  except Exception as e:
+    logger.warning(f"Could not determine capacity for tier {tier}: {e}")
+    return "at_capacity"
+  return "ready" if status_value == "ready" else "at_capacity"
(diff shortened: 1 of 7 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Follow-up in ace5b80a: (1) no private IP now raises `DrainRefusedError` instead of returning — nothing could be asked, so nothing is detached; (2) "unreachable" is confirmed across three attempts (`DRAIN_UNREACHABLE_CONFIRMATIONS`) before it is read as a stopped container, with a test that a first-attempt blip followed by a live API still refuses. On the question: the maintenance-window precondition is procedural (the runbook), but the code enforces its effect — `graph_api` mounts no drain endpoint, so an API that is up always hits the 404 branch and refuses; a stopped container is the only path through, and now a brief blip cannot impersonate one. The 404-then-`>= 400` overlap is deliberate for the more specific message.

PR #1306, opened 2026-08-29
fix(reports): a zero credit balance renders as $0.00, not -$0.00

jfrench9:
## Summary

On a live balance sheet, `Accounts Payable, Current` rendered as **−$0.00** when the account's debits exactly equalled its credits. The value leaving the API was a literal `-0.0`: `_natural_sign` negates credit-normal balances, and `-(0.0)` is `-0.0` — equal to zero for every comparison and filter in the renderer, but formatted with a sign by the app. Subtotals happened to come out clean because their rollup starts from an integer accumulator; leaf rows carried the raw value through.

## Changes

- **`operations/roboledger/reports/fact_grid.py`**
  - `_natural_sign` returns an unsigned `0.0` for a zero credit balance instead of negating it.
  - New `_unsigned_zero` applied where rows leave `_build_rows`, so no other sign-flipping path (cash-flow deltas, equity reducers) can reintroduce `-0.0` on a rendered row.
- **`tests/operations/roboledger/reports/test_fact_grid.py`** — two regressions that check the sign with `math.copysign` (the existing `test_zero` passed on `-0.0`, since `-0.0 == 0.0`): the helper directly, and a `_build_rows` case with a zero-balance credit leaf that stays visible because its prior period is non-zero — the exact shape seen on prod.

## Breaking Changes

None. Response shapes are unchanged; only a signed zero becomes an unsigned zero.

## Testing

- `just test` — full unit suite: 14,193 passed, 42 skipped (223 in the reports/reads modules specifically).
- `just test-code` — clean; re-run by the pre-commit hook.
- Reproduced from prod: `live-financial-statement` (balance sheet, YTD) on the Harbinger ledger returned `"values": [-0.0, 195.94]` for AP before the fix.

## Certification

### DIFF
--- a/robosystems/operations/roboledger/reports/fact_grid.py
+++ b/robosystems/operations/roboledger/reports/fact_grid.py
@@ -2811,7 +2811,8 @@ def _build_rows(
       # rolled-up sum of all descendants for parent nodes, so subtotal
       # rows get the correct aggregate instead of zeros.
       vals = [
-        computed_per_period[i].get(node.element_id, 0.0) for i in range(n_periods)
+        _unsigned_zero(computed_per_period[i].get(node.element_id, 0.0))
+        for i in range(n_periods)
       ]
     # A row is a subtotal if it aggregates other rows by either path:
     # (a) it has child summands in the disclosure DAG, or (b) it is a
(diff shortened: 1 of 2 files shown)


jfrench9:
@claude please review this PR

PR #1307, opened 2026-08-29
feat(operators): prefetch recalled memories into the question; say when document search should go semantic

jfrench9:
## Summary

The console's Cypher operator has had `recall` on its read-only allowlist since fa2267a8, but nothing in the prompt routed to it, and under "act on your first turn" with a 2/5/12-turn step budget the model never spent a turn on it — access without use. This prefetches the memories most similar to the question up front and renders them ahead of the question in the user turn, so the operator reads what earlier sessions stored at zero tool-turn cost. It also corrects the prompt's description of `search-documents`, which called it hybrid semantic search while the tool defaults to keyword-only.

## Changes

- `operations/operators/implementations/cypher.py`
  - New `_fetch_memories`: when the graph exposes `recall`, calls it with the question as the query (`k=5`). Any failure, an error-shaped result, or an empty store returns `None` and the question goes in bare — a memory miss never costs the question.
  - New `_build_user_message`: renders the hits as a `REMEMBERED CONTEXT` block (whitespace collapsed, 800 chars per hit, tags appended) framed as background data rather than instructions, with the question last. **Placement is the point reviewers should check:** the block goes in the user turn, not the system prompt, because it is per-question tenant data and the system prefix is cached per graph.
  - System prompt gains a `MEMORY` section, gated on `recall` being available (static per graph, so it stays inside the cached prefix), telling the model the relevant memories are already in the user message and to call `recall` only for follow-up lookups. `recall` stays on the offered tool list.
  - The `search-documents` guidance in both the SEC and tenant branches now says it is BM25 by default and when to pass `semantic=true`.
- `operations/operators/tool_loop.py` — `run_tool_loop` gains a keyword-only `user_message: str | None` that replaces `ctx.query` as the opening user turn. Defaults to the existing behavior; the Cypher operator is the only caller.
- Tests — `test_cypher_orientation.py` (memories land in the user turn and not the prefix; empty store, raised error, error dict, blank text, and non-dict results all fall back to the bare question; graphs without `recall` make no call and get no `MEMORY` section; per-hit cap; the semantic nudge is present) and `test_tool_loop.py` (the override replaces the opening turn; the default is still the question).

Graphs without `recall` — the `sec` shared repository, and deployments with `SEMANTIC_MEMORY_ENABLED` / `MCP_SEMANTIC_MEMORY_ENABLED` off — are unaffected beyond the `search-documents` wording.

## Breaking Changes

None. No API, GraphQL, or SDK surface changes; the new `run_tool_loop` parameter is optional and internal.

## Testing

- `uv run pytest tests/operations/operators` — 207 passed (includes the 12 new tests).
- `just test-code` — all checks passed (ruff, format, basedpyright 0 errors, cf-lint).
- `just test-all` — not run; the full unit suite was not executed this session.

## Certification

### DIFF
--- a/robosystems/operations/operators/implementations/cypher.py
+++ b/robosystems/operations/operators/implementations/cypher.py
@@ -42,6 +42,13 @@ from robosystems.operations.operators.tool_loop import (
 _ORIENTATION_TOOLS = ("get-graph-schema", "get-example-queries")
 _MAX_ORIENTATION_CHARS = 48000
 
+# Semantic memory is the other prefetch, with the opposite placement: what is
+# relevant depends on the question, so the hits go in the user turn rather
+# than the cached prefix. Small caps — a memory is a fact or a decision, and
+# five of them is context, not a document dump.
+_RECALL_K = 5
+_MAX_MEMORY_CHARS = 800
+
 
 @register_operator("cypher")
 class CypherOperator(Operator):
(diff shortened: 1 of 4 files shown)


jfrench9:
@claude please review this PR

PR #1308, opened 2026-08-29
refactor(operators): rename the cypher operator to analyst, keep cypher as an alias

jfrench9:
## Summary

The operator registered as `cypher` answers natural-language questions from a 34-tool read-only surface — curated statements and fact grids, close and mapping status, documents, memory (#1307), GraphQL — and its own prompt tells the model to reach for Cypher only when no curated tool fits. The name had outgrown it, and its self-description (spec, docstring, the prompt's opening line) still said "answers by writing Cypher", contradicting the routing two paragraphs later. This renames it to `analyst` and keeps `cypher` as a registry alias so nothing that stored or hardcoded the old name breaks.

## Changes

- `operations/operators/implementations/cypher.py` → `analyst.py` — `CypherOperator` → `AnalystOperator`, registered as `analyst`. Spec name/description, module docstring, and the system prompt's opening line now describe the operator as it is: curated reads where one fits, documents and memory where the graph has them, read-only Cypher for everything else. `operator_type` on its model calls and the loop's operation description follow. No change to the allowlist, the loop, or the structured `cypher`/`rows` output the console renders.
- `operations/operators/operator_registry.py` — **the part to look at.** New `_ALIASES = {"cypher": "analyst"}` and `resolve_operator_type()`; `get_operator`, `get_operator_class`, and `is_registered` resolve through it. A direct registration always beats an alias; `list_operators()` shows canonical names only.
- `routers/graphs/operator/execute.py` — `_dispatch` resolves the alias before enqueue so the queued task and `operator_used` carry the canonical name; the metadata endpoint resolves it too. Endpoint descriptions name `analyst` and note `cypher` is accepted as its former name. `POST /v1/graphs/{g}/operator/cypher` keeps working.
- `config/operators.py` — `fallback_operator` and the `OPERATOR_CAPABILITIES` key renamed.
- Docs — `CLAUDE.md`, the operators/operations/config READMEs, and docstring mentions in `base.py`, `tool_loop.py`, `adapters/worker.py`.
- Tests — `test_cypher_orientation.py` → `test_analyst_prefetch.py`; registry alias tests (former name resolves, listing stays canonical, direct registration wins); router test that `/operator/cypher` queues `operator_type == "analyst"`; string updates across the worker adapter, orchestrator, config, operations, and worker-client tests.

[...]

PR #1310, opened 2026-08-29
chore(deps): dev client floor to 1.12.2

jfrench9:
## Summary

Adopt `robosystems-client` 1.12.2 in the dev dependency group (the client is used only by the demo scripts under `examples/`). The release carries the regenerated operator endpoint docs for the `analyst` rename (#1308) and removes three POC facade methods (`analyze_financials` / `research` / `rag`) that nothing here called.

## Changes

- `pyproject.toml` — dev floor `robosystems-client>=1.12.0,<2.0` → `>=1.12.2,<2.0`, matching the convention from the 1.12.0 bump (`a4c6307a`).
- `uv.lock` — resolves to 1.12.2.

No application code changed; the client is not a runtime dependency of the API.

## Breaking Changes

None.

## Testing

- `just test-all` — not run; dependency floor and lockfile only, and the client is dev-group-only.

## Certification

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,7 +112,7 @@ dependencies = [
 [project.optional-dependencies]
 dev = [
     # RoboSystems client for demo and testing.
-    "robosystems-client>=1.12.0,<2.0",
+    "robosystems-client>=1.12.2,<2.0",
     # Testing framework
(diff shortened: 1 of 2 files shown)


jfrench9:
@claude please review this PR

PR #1311, opened 2026-08-29
fix(ledger): retract a fulfilled event while its rows are still drafts

jfrench9:
## Summary

A handler can target `fulfilled` while its ledger rows are still `draft`, because drafts post at close — `asset_disposed` does. (`journal_entry_reversed` also targets `fulfilled`, but posts its reversing entry immediately, so the guard below refuses it from the start; the correction there is another reversal.) Three shipped rules then closed on that pair with no way out: `delete_journal_entry` refuses to delete the last draft of a live event and tells the caller to void or supersede it, while `fulfilled` had an empty transition set, so that advice could not be followed. The result was that **a disposal draft could never be discarded** — only posted and then reversed.

This allows `fulfilled` → `voided` / `superseded`, guarded so a retraction can never orphan an effect the books already hold.

## Changes

**`operations/event_block/commands.py`** — the fix.

- `_VALID_TRANSITIONS["fulfilled"]` gains `voided` and `superseded`. `voided` and `superseded` stay empty — a retraction is final.
- New `_assert_retractable`, consulted only for `fulfilled`. Retraction drops the event from `is_live`, which materialization, QuickBooks write-back and every reporting read filter on. That is the right answer while the rows are drafts — nothing has hit the books, so removing the event removes the whole story. It is the wrong answer once a row has posted or the event has published to QuickBooks: the effect stays where it landed while the event explaining it disappears, leaving a GL balance no query can attribute. Past that line the correction is a reversal, the same split `delete_journal_entry` already draws for entries.
- The guard refuses on three signals: `metadata.qb_external_id` set, any `Entry` in `posted`/`reversed`, any `Transaction` in `posted`. `reversed` counts — the original posted and a reversing entry stands against it, so retracting would strand both halves.
- The guard runs on **every** retraction, from any status. Scoping it to `fulfilled` was wrong: `classified` and `committed` handlers (`schedule_entry_due`, `journal_entry_recorded`, `schedule_created`) write their ledger rows when they fire, and close promotes any draft whose event is not already retracted — it filters on voided/superseded, not on fulfilled. A `committed` event could therefore sit across a close, collect posted entries, and be voided with no guard at all. Those transitions were already reachable, so this closes a live gap rather than a hypothetical one.
- The retraction takes the **shared period fence** on its rows' posting dates before the event row lock, in the order close and `delete_journal_entry` already use. The row lock alone does not order a retraction against a close: close promotes drafts with a bulk `UPDATE` over `entries` filtered on posting date and never touches the owning `Event` row, so it does not contend on that lock, and the guard could read zero posted rows in the window where close was about to post them. Fenced on the rows rather than the event, so an event that never produced any stays retractable even where its date falls in a closed period.
- New `EventEffectsAlreadyLandedError`, exported from the package.

**`routers/extensions/roboledger/operations.py`** — maps the new error to **422** on `update-event-block`. It subclasses `Exception`, so without this it would have fallen through to the generic handler; the caller needs an answer they can act on, not a 500.

**`models/api/event_block.py`** — `status` field description now says `fulfilled` is retractable while its ledger rows are drafts. Description text only; no shape change.

**`examples/roboledger_demo/validate.py`** — the case that surfaced this. The disposal smoke test's cleanup had been failing since the delete guard landed (2026-08-18; the cleanup dates from 2026-07-10), leaving an orphan $100 draft in the very period the demo then instructs the user to close, where it would post as a spurious closing entry. It now voids the event before deleting the draft, and asserts the close target is clear afterwards. 42 → 44 checks.

**`tests/operations/event_block/test_commands_update.py`** — `test_supersede_from_terminal_state_rejected` encoded the old rule and was replaced by `test_supersede_from_retracted_state_rejected`, which keeps the coverage for `voided`/`superseded`. New `TestRetractFulfilled` covers: void allowed while rows are drafts, refused once an entry has posted, refused once published to QuickBooks, the guard applying from `captured`/`classified`/`committed`/`pending` and not only `fulfilled`, the fence being taken before the row lock, and no fence taken when the event has no ledger rows. The pre-existing `test_void_does_not_take_period_fence` still passes unchanged.

## Breaking Changes

None. The transition set widens and no shape changes, so no client SDK coordination is needed — nothing in the stable tier is touched, and there is no GraphQL schema or operations-envelope change.

Two behavior notes for reviewers rather than breaks:

- A `fulfilled` → `voided` call that previously always returned 422 `terminal state` may now succeed. That is the bug being fixed.
- `update-event-block` gains a new 422 case for a retraction of a `fulfilled` event whose effects have landed.

The OpenAPI `status` description changed, so the generated tier of both SDKs will pick up new docstring text on the next regen. Cosmetic; no symbol or shape movement.

## Testing

- `just test-all` — **14214 passed, 42 skipped**; ruff, format, basedpyright and cf-lint all clean.
- Verified against the local stack, not only unit tests. The event that was actually stuck now voids and its draft deletes, taking the close target to `[]`. A `fulfilled` event with a posted entry in an **open** period is refused by the guard (`422 … 1 of its journal entries have posted. Reverse the posted entries instead`); one whose entries fall in a **closed** period is refused earlier by the fence (`422 Cannot write to closed period '2025-06'`).
- `just demo-roboledger` re-run end-to-end on a freshly reset environment: **44/44 checks**, with the close target confirmed empty afterwards.
- The other seven demos (`demo-user`, `coffee-roaster`, `saas-startup`, `roboinvestor`, `custom-graph`, `seattle-method`, `world-online`, `sec`) were run on the same reset environment and were unaffected — RoboInvestor 31/31, Seattle Method 18/18 exact, World Online 7/7 statement anchors tie at Δ 0.00.

## Review response

Both findings from the `@claude` review were verified against the code and were correct; both are fixed in `8ae96e06`.

1. **The concurrency claim in this PR's original description was wrong.** I had written that the event row lock kept a concurrent close from posting a draft between the guard's check and the write. It does not — close reaches the entries through a bulk `UPDATE` that never touches the `Event` row. The shared period fence is the mechanism that orders the two, and the retraction path now takes it. The original claim has been corrected above rather than left standing.
2. **The guard was under-scoped**, and by more than the review said — `classified` is affected too, not only `committed`/`pending`, since `schedule_entry_due` and `journal_entry_recorded` both create entries at that status. The guard now runs on every retraction.

3. **Second review pass** (`docs(ledger)` follow-up commit): `UpdateEventBlockRequest.transition_to`'s description still called `fulfilled` terminal and omitted the new transition, while the updated `status` description defers to it for the graph — now consistent. The transition-table comment had named `journal_entry_reversed` as landing with drafts; it does not (see above), so the comment and this description were corrected. Router and handler-type comments updated to say the guard runs from every status.

## Certification


jfrench9:
@claude please review this PR

jfrench9:
Both findings verified and fixed in `8ae96e06`. Both were correct.

**1. The concurrency claim was wrong** — mine, in the PR description, not in the code. You're right that `with_for_update()` on the `Event` row doesn't serialize against close: `PeriodCloseService.close()` promotes drafts with a bulk `UPDATE` over `entries` filtered on posting date, excluding retracted events through a plain subquery, and never touches the `Event` row. So it never contends on that lock, and `_assert_retractable` could read zero posted rows in exactly the window where close was about to post them.

The retraction path now takes the shared period fence before the row lock, in the fence-then-rows order close and `delete_journal_entry` already use. One refinement on the suggested fix: it fences on the **rows' own posting dates**, not the event's date, and skips the fence when the event has no ledger rows. That keeps `test_void_does_not_take_period_fence` passing on its original intent — an event that never produced rows has nothing to strand and stays retractable even where its date falls in a closed period.

I've corrected the claim in the PR description rather than leaving it standing.

**2. The guard was under-scoped — and by more than you flagged.** `classified` belongs in the list too: `schedule_entry_due` and `journal_entry_recorded` both target `classified` and return `entry_ids`, so the handler has run and the rows exist. My docstring's reasoning ("a handler has not run and there is nothing behind the event yet") was simply false for `classified`, `committed` and `pending` alike.

To your question about whether it was intentionally out of scope — it wasn't, and there's no invariant making those statuses safe. Close's exclusion filter is `status IN ('voided','superseded')`, so a `committed` event can sit across a close, collect posted entries, and be voided today with no guard. The guard now runs on **every** retraction; what decides is whether anything landed, never which status it's retracted from. The status-based rationale is gone from the docstring.

Verified live, both layers: a `fulfilled` event with a posted entry in an open period is refused by the guard; one whose entries fall in a closed period is refused earlier by the fence. `just test-all` is 14214 passed / 42 skipped, and the roboledger demo is still 44/44 with the close target empty.

Noted on the static-only review — the test results above are from local runs.

PR #1312, opened 2026-08-29
refactor(operators): remove the orchestrator response cache

jfrench9:
## Summary

Removes the orchestrator's opt-in in-memory response cache. It was never enabled in production and could not have worked as written, but its key omitted the graph scope, which made it a trap for whoever switched it on next. The Bedrock prompt cache that carries the operators' actual cost savings lives in `AIClient` and is untouched.

## Changes

- `robosystems/operations/operators/orchestrator.py`
  - Drop `OrchestratorConfig.enable_caching`, the `_cache` dict, `_get_cache_key`, the read/write around `route_query`, and the `cache_hits` / `cache_misses` counters from `get_metrics()`.
  - Drop the `force_extended` parameter on `route_query` — it existed only to bypass this cache. (The API's `force_extended_analysis` is unrelated: it selects the mode at the router and is unchanged.)
  - Why removal rather than a better key: the cache was off by default, the one production constructor never enabled it, and `route_query` has no production caller (the router uses `get_operator_recommendations` and hands execution to the worker). The key also used `hash(str)`, which is salted per process, so no entry would have survived a worker boundary.
- `tests/operations/operators/test_orchestrator.py` — remove the `enable_caching` config assertions and the `test_caching_enabled` case.
- `robosystems/models/api/graphs/operator.py` — drop `cache_hits` / `cache_misses` from `OperatorMetricsResponse`, which mirrors `get_metrics()`. The model is exported but bound to no router, so it is absent from the OpenAPI spec and both SDKs (verified by grep of the generated clients).

Net −49 lines, no behavior change on any reachable path.

## Breaking Changes

None. Nothing here reaches the OpenAPI spec or either SDK tier; `OrchestratorConfig` and `route_query` are internal Python surface.

## Testing

- `just test-code` — all checks passed (ruff, format, basedpyright, cf-lint).
- `uv run pytest tests/operations/operators/ tests/routers/graphs/test_operator_router.py` — 240 passed; `tests/models/api/test_operator_models.py` + `test_orchestrator.py` re-run after the model change — 47 passed.
- `just test-all` not run in-session; the PR's `test / test` job passed on the first commit (10m49s) and is re-running on the second.

## Certification

### DIFF
--- a/robosystems/models/api/graphs/operator.py
+++ b/robosystems/models/api/graphs/operator.py
@@ -268,8 +268,6 @@ class OperatorMetricsResponse(BaseModel):
   average_response_time: float = Field(
     ..., description="Average response time in seconds"
   )
-  cache_hits: int = Field(0, description="Number of cache hits")
-  cache_misses: int = Field(0, description="Number of cache misses")
   errors: int = Field(0, description="Number of errors")
   timestamp: datetime = Field(
     default_factory=datetime.utcnow, description="Metrics timestamp"
--- a/robosystems/operations/operators/orchestrator.py
+++ b/robosystems/operations/operators/orchestrator.py
@@ -59,7 +59,6 @@ class OrchestratorConfig:
 
   routing_strategy: RoutingStrategy = RoutingStrategy.BEST_MATCH
   enable_rag: bool = False
-  enable_caching: bool = False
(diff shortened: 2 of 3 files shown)


jfrench9:
@claude please review this PR

PR #1314, opened 2026-08-30
fix(graph-api): require the cluster key except on /health

jfrench9:
## Summary

Require the Graph API cluster key on every route except `/health`, matching the documented contract. Load-balancer and container probes stay unauthenticated; everything else uses the key the platform client already sends.

## Changes

- `graph_api/middleware/auth.py` — exempt only `/health` in prod/staging (dev still unauthenticated).
- `graph_api/app.py` — OpenAPI security scheme matches that exemption.
- `middleware/mcp/client.py` — cluster probe goes through `GraphClient` so it carries the key.
- Tests pin the exemption and the prod deny on non-health cluster routes.

## Breaking Changes

None for the public API / SDKs. Graph API callers in prod/staging that hit non-health routes without `X-Graph-API-Key` (or Bearer) will get 401. The platform Graph client and the volume-monitor Lambda already send the key.

## Testing

`uv run pytest tests/graph_api/test_auth_middleware.py tests/middleware/mcp/test_client.py tests/graph_api/routers/test_info.py tests/graph_api/routers/test_metrics.py tests/graph_api/routers/test_health.py` — 77 passed. Pre-commit (ruff + basedpyright) passed on commit. `just test-all` not run.

## Certification

### DIFF
--- a/robosystems/graph_api/app.py
+++ b/robosystems/graph_api/app.py
@@ -223,16 +223,7 @@ def create_app() -> FastAPI:
         }
       }
 
-      exempt_paths = {
-        "/health",
-        "/status",
-        "/info",
-        "/metrics",
-        "/openapi.json",
-        "/docs",
-        "/redoc",
-        "/",
-      }
+      exempt_paths = {"/health"}
       for path, methods in openapi_schema["paths"].items():
         if path not in exempt_paths:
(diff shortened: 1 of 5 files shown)


jfrench9:
@claude can you review

PR #1315, opened 2026-08-31
chore(docs): stop hardcoding the SEC dump's per-snapshot sizes

jfrench9:
## Summary

The SEC dump's archive and expanded sizes were written out in four places in this repo, and every one of them went stale with each new snapshot — the 2026-08-31 publish moved them from ~35/~128 GiB to ~39/~130 GiB. This replaces the exact figures with a bracket that survives corpus growth and points readers at the dataset card, which is the one place the numbers are already maintained per snapshot.

`robosystems/scripts/sec_dump.py` already reads both real sizes from the dataset at run time and gates the download and the decompression on free space against them, so the hardcoded numbers were duplication that no reader depended on being precise.

## Changes

- `robosystems/scripts/sec_dump.py` — drop the size figures from the usage block; add a docstring paragraph noting that both sizes are read from the dataset at run time and checked against free space, so none is repeated in the source.
- `justfile` — `sec-dump` recipe comment.
- `README.md` — SEC shared-repository section; now also says the dataset card carries each snapshot's exact sizes.
- `CLAUDE.md` — SEC quick-reference block.

All four now read *tens of GiB down, >100 GiB on disk*, which keeps the "this is not a laptop-minutes download" warning without a number that rots monthly.

No behavior change: nothing outside comments, a module docstring, and Markdown prose is touched.

## Breaking Changes

None. No API surface, no SDK-visible symbols, no runtime code paths.

## Testing

`just test-code` (ruff + format + basedpyright + cf-lint) — all checks passed, 0 errors / 0 warnings / 0 notes.

The unit-test half of `just test-all` was not run to completion: it needs the local Postgres, which was not up in this session, and all 14 collection errors were `psycopg2.OperationalError: connection refused` on `localhost:5432`. They are environmental and unrelated to this diff, which contains no executable code.

## Certification

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -156,7 +156,7 @@ just lbug-query GRAPH_ID "CYPHER_QUERY"        # Direct LadybugDB query (bypass 
 just sec-load NVDA 2025    # Load company filings
 just sec-health            # SEC database health
 just sec-reset             # Reset SEC database
-just sec-dump              # Pull the public SEC .lbug dump from Hugging Face (~35 GiB down, ~128 GiB on disk)
+just sec-dump              # Pull the public SEC .lbug dump from Hugging Face (tens of GiB down, >100 GiB on disk)
 ```
 
 ### Demo Scripts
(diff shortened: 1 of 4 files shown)


jfrench9:
@claude please review this PR

PR #1317, opened 2026-08-31
fix(graphs): suspend the graph inline on immediate delete-graph

jfrench9:
## Summary

An immediate `delete-graph` cancels the subscription and returns, but never touches `Graph.status`. That status is a projection of billing state, reconciled by `expired_graph_subscription_sensor` on a 300s tick — so until the next tick, every read reports the graph as `active` and it stays fully usable, despite being irreversibly destroyed the moment the cancel committed.

This sets the status inline on the immediate path, so the API stops advertising a destroyed graph as healthy.

## Changes

- **`robosystems/routers/graphs/operations.py`** — in `delete_graph_op`, the `immediate` branch now transitions the graph to `suspended` before returning.
  - Guarded on `status == ACTIVE`, because `transition_status` rejects `suspended -> suspended` and the sensor may already have won the race.
  - Skipped entirely when `at_period_end=true` — that graph is paid through the period and must stay usable until it closes. Only the sensor may suspend it, once `ends_at` actually passes.
  - A missing registry row is tolerated rather than raised: the billing cancel has already committed, and turning a successful teardown into a 500 would be worse than leaving the sensor to reconcile.

## Why this is safe

`subscription.cancel(immediate=True)` writes exactly what the sensor keys off — `status=canceled`, `ends_at=now`, `cancellation_type=immediate` — so the operation already holds the conclusion the sensor would reach 0-300s later. It is applying a decision, not making a new one.

`suspend_expired_graphs`'s entire body is one `transition_status(SUSPENDED)` call, so the inline write is a complete reproduction of what the job does, not a partial one.

**The sensor is retained and not weakened.** It still uniquely owns:

- period-end cancellations, where `ends_at` is in the future and becomes eligible later;
- `status = "failed"` (payment failure), which never routes through this operation;
- provider-originated cancellations arriving by webhook;
- any run that dies between the cancel commit and this write.

That last case is why ordering matters — provider cancel, then `cancel()`, then the status write — and why the sensor must remain rather than be replaced. Its query filters on `Graph.status == "active"`, so once this write lands the row simply stops matching; there is no double transition.

## Measured behaviour

Two production teardowns, timed against the API:

| | `delete-graph` accepted | reported `suspended` | row gone |
| --- | --- | --- | --- |
| graph A | 05:57:48Z | ~06:02:5xZ (~5.1 min) | 06:05:37Z |
| graph B | 06:05:00.6Z | 06:07:51Z (2m50s) | 06:10:36Z |

The two suspend events are 300s apart and the two deprovision events are 299s apart, confirming a fixed-phase 300s cadence on both sensors. The wait is therefore not a predictable lag but a uniform 0-300s draw depending on where the request lands in the cycle — a caller could equally see 10 seconds or 4m59s with no way to tell which.

Deprovisioning timing is unchanged by this PR and remains sensor-driven, which is fine: by then the graph already reads as `suspended`.

## Breaking Changes

None. No request or response shape changes — `delete-graph` returns the same `deprovisioning_queued` envelope. The only observable difference is that a subsequent `GET /v1/graphs` reports `suspended` immediately instead of after a delay, which is the intended fix. No SDK regeneration required.

## Testing

- `just test` — 14218 passed, 42 skipped.
- `just test-code` — ruff, format, basedpyright and cf-lint all clean.
- Four new tests in `tests/routers/graphs/test_delete_graph_op.py` covering the immediate transition, the `at_period_end` non-transition, the already-suspended guard, and the missing-row path. The first was confirmed to fail against the unmodified handler.

## Related

Frontend companion: RoboFinSystems/robosystems-app#350 renders a `Suspended` badge and disables actions on a non-active graph. That PR made the gap visible; this one closes it.

### DIFF
--- a/robosystems/routers/graphs/operations.py
+++ b/robosystems/routers/graphs/operations.py
@@ -363,6 +363,7 @@ async def delete_graph_op(
     BillingSubscription,
     SubscriptionStatus,
   )
+  from robosystems.models.core.graph import Graph, GraphStatus
   from robosystems.models.core.graph.graph_user import GraphUser
   from robosystems.operations.providers.payment_provider import get_payment_provider
   from robosystems.security import SecurityAuditLogger, SecurityEventType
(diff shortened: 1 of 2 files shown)


jfrench9:
@claude please review this PR

PR #1332, opened 2026-09-01
chore(deps): pin uv, hold two majors, take the focused dbt upgrade

jfrench9:
## Summary

Follow-up to this month's dependabot pass. Three of the ten proposed bumps were majors or carried freight we don't want, so they were closed in favour of deliberate decisions recorded in-tree; while checking them, the uv binary itself turned out to be unpinned in five places. This PR pins uv, records the two holds where dependabot will otherwise re-propose them, and takes the narrow dbt upgrade that delivers what 1.12 was wanted for without its new dependencies.

## Changes

**Pin the uv binary (`pyproject.toml`, `Dockerfile`, `Dockerfile.lambda`)**

uv was unpinned everywhere it is installed — `setup-uv` with no `version:` in `test.yml` and `create-release.yml`, two `curl -LsSf https://astral.sh/uv/install.sh | sh` installs in `Dockerfile`, and `ghcr.io/astral-sh/uv:latest` in `Dockerfile.lambda`. Nothing tracked any of them, so every CI run and image build took whatever uv was newest: the last green Test CI run used 0.12.8, published the day before.

`[tool.uv] required-version = "==0.12.8"` pins it in one place. `setup-uv` reads that key on its own — its `version-file` input defaults to searching `uv.toml` then `pyproject.toml` — so both workflows pin with no workflow change, and uv refuses to run on a mismatch, which makes local/CI skew visible instead of silent. The Dockerfiles now `COPY --from=ghcr.io/astral-sh/uv:0.12.8`, replacing the two curl-pipe-shell installs and the `:latest` tag.

**Hold python-ulid at 3.x (`pyproject.toml`)**

`python-ulid>=3.0.0` was unbounded, so 4.x will be re-proposed indefinitely. 4.0 reworks generation customization only — the new `ULIDGenerator`, monotonicity policies, and the removal of `ValueProvider`/`ULID.provider`. `robosystems/utils/ulid.py` uses `ULID()`, `from_str`, `.timestamp` and `.bytes`, none of which changed, so the major carries risk with no corresponding gain.

**Hold astral-sh/setup-uv at v8 (`.github/dependabot.yml`)**

v9 flipped `prune-cache` to `false`, which grows the Actions cache to the full unpruned uv cache on a dependency set including onnxruntime, arelle, scipy and dagster. v10's cache-poisoning guard applies only to `pull_request_target`, `workflow_run` and `release`; `test.yml` sets `enable-cache: true` explicitly and `create-release.yml` is `workflow_dispatch`, so none of our workflows benefit. Majors are ignored; minors and patches still flow.

The `ghcr.io/astral-sh/uv` image is also ignored for the docker ecosystem, since a Dockerfile-only bump would split the pin from `required-version`.

**Take dbt-core 1.11.14 instead of 1.12.3 (`pyproject.toml`, `uv.lock`)**

Both releases lift dbt's sqlparse ceiling to `<0.7.0`, which is the only thing the 1.12 bump was wanted for. 1.12 additionally adds `dbt-core-experimental-parser` (resolving to a `2.0.0b2` beta) as a hard runtime dependency and swaps `dbt-semantic-interfaces` for `metricflow`, pulling in sqlglot, rapidfuzz, tabulate, referencing and opentelemetry-api for a semantic layer the QuickBooks dbt project doesn't use — all of it into the runtime image, since dbt is a main dependency.

1.11.14 keeps the existing tree and retires the `[tool.uv] override-dependencies = ["sqlparse>=0.6.0"]` block, whose own comment said to delete it once dbt lifted the ceiling. sqlparse stays at 0.6.0 (so the CVE fixes that motivated the override are retained) but now resolves natively rather than being forced past a constraint dbt declared incompatible. `dbt-duckdb` moves to 1.11.0, which requires only `dbt-core>=1.8.0`.

Reviewers may want to confirm the lock: `dbt-core-experimental-parser`, `metricflow`, `sqlglot` and `rapidfuzz` should all be absent, and `sqlparse` should still be 0.6.0.

## Breaking Changes

None. No API surface, GraphQL schema, operations envelope or REST shape is touched, so there is no client SDK impact in either tier.

One local-workflow consequence, by design: `required-version` makes uv exit with an error if the running binary isn't 0.12.8, so contributors on an older uv must upgrade (`brew upgrade uv`) before `uv run` works.

## Testing

`just test-all` passed on this change set (exit 0). Edits after that run were limited to comment wording and dependency-list ordering; the pre-commit gate (ruff, format, basedpyright) passed on the committed tree, and `uv lock --check` confirms the lock matches `pyproject.toml`.

Additionally verified out of band: all 244 locked packages scanned against the OSV advisory database with zero hits, confirming no security driver was deferred by the two holds.

Not exercised locally: the Docker image builds, which are the surface the `COPY --from` change affects. CI build is the check for those.

## Certification

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,10 @@ updates:
       interval: "monthly"
     labels:
       - "dependencies"
+    ignore:
+      # hold: v9's prune-cache flip bloats the cache; v10's guard covers events we don't use
+      - dependency-name: "astral-sh/setup-uv"
+        update-types: ["version-update:semver-major"]
 
   # Python — uv dependencies in pyproject.toml and uv.lock
   - package-ecosystem: "uv"
@@ -47,3 +51,5 @@ updates:
         update-types: ["version-update:semver-major", "version-update:semver-minor"]
       - dependency-name: "lambda/python"
         update-types: ["version-update:semver-major", "version-update:semver-minor"]
+      # Must move in lockstep with [tool.uv] required-version in pyproject.toml
+      - dependency-name: "ghcr.io/astral-sh/uv"
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,9 +19,10 @@ RUN apt-get update && apt-get upgrade -y && apt-get install -y --no-install-reco
     curl \
     unzip \
     file \
-    && rm -rf /var/lib/apt/lists/* \
-    && curl -LsSf https://astral.sh/uv/install.sh | sh \
-    && mv /root/.local/bin/uv /usr/local/bin/uv
+    && rm -rf /var/lib/apt/lists/*
+
+# uv pinned to [tool.uv] required-version in pyproject.toml
+COPY --from=ghcr.io/astral-sh/uv:0.12.8 /uv /usr/local/bin/uv
 
 # Copy LadybugDB extensions from official extension repository
 # Extensions pulled from ghcr.io/ladybugdb/extension-repo:latest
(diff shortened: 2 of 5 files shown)


jfrench9:
@claude please review this PR

PR #1340, opened 2026-09-01
chore(image): drop curl and git from the runtime stage

jfrench9:
## Summary

Remove `curl` and `git` from the API image's runtime stage and move every in-container health probe to a stdlib-only Python script. Neither package is used by the service at runtime, and they pull `libcurl`, `libssh2`, `git` and `liberror-perl` into an image the container scanners keep flagging for advisories no code path here can reach. Removing them clears those findings on rebuild instead of re-explaining them every cycle.

## Changes

- **`Dockerfile`** — runtime stage no longer installs `curl` or `git`. `git` is only needed in the builder (where `arelle_cache_manager.py` fetches the EDGAR subtree) and no dependency in `uv.lock` is git-sourced. `postgresql-client` stays because `entrypoint.sh` waits on the database with `psql`. Comment explains the choice so the packages don't get helpfully restored.
- **`bin/healthcheck.py`** (new) — `GET url [timeout]`, exit 0 on 2xx, 1 otherwise, standard library only. Replaces `curl -f` in every probe.
- **`compose.yaml`** — API and graph-api health checks call the script.
- **`cloudformation/dagster.yaml`** — Dagster webserver ECS container health check calls the script (`CMD` form, no shell).
- **`bin/userdata/common/run-graph-container.sh`** — LadybugDB graph container `--health-cmd` calls the script with the same 10s budget.
- **`bin/userdata/common/graph-lifecycle.sh`** — the ASG drain hook exec'd `curl` inside the graph container; it now calls the published port from the host. Reviewers should note: the Graph API exposes neither `/admin/drain` nor `/admin/connections`, so both calls already fell through to their fallbacks. Behavior is unchanged; the script now says so in a comment. Implementing those endpoints is a separate piece of work.

Not changed: `perl` remains in the image. `postgresql-client` depends on it and `perl-base` is Essential on Debian, so scanner matches against the `perl` source package are a base-image question, not a package-list one.

## Breaking Changes

None. No API surface or SDK-visible shape changes; the SDK tiers are untouched.

## Testing

- `just test-code` — passed (ruff, format, basedpyright, cf-lint incl. the edited `dagster.yaml`).
- `bin/healthcheck.py` exercised inside the running local containers (`robosystems-api`, `robosystems-graph-api`): 2xx → exit 0, 404 → exit 1, connection refused with a 2s timeout → exit 1. `python3` resolves to the venv interpreter for `appuser`, which the `CMD`-form probes rely on.
- Built a trimmed `python:3.13-slim` + runtime package set locally and confirmed with `dpkg-query` that `libssh2-1t64`, `libcurl4t64`, `git` and `liberror-perl` are absent; Trivy on that image reports no libssh2 findings.
- `bash -n` on both edited userdata scripts.
- Full image build and the unit-test half of `just test-all` were not run locally; CI's build + Trivy scan covers the real image.

## Certification

https://claude.ai/code/session_012mdzNFGhSkgCy8eSmsJjJD

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -129,13 +129,15 @@ ENV PYTHONUNBUFFERED=1 \
 # Install runtime dependencies, apply security patches, and install uv.
 # CACHE_DATE (set per-build in build.yml) busts this layer so the upgrade
 # re-runs despite GHA layer caching — a cached layer keeps stale OS packages.
+# No curl or git here: git is only needed in the builder (EDGAR subtree), and
+# health probes use bin/healthcheck.py — curl would drag libcurl and libssh2
+# into the image, which the container-image scanners then flag.
+# postgresql-client stays: entrypoint.sh waits on the database with psql.
 ARG CACHE_DATE
 RUN echo "os-refresh ${CACHE_DATE}" && apt-get update && apt-get upgrade -y && apt-get install -y --no-install-recommends \
     libpq5 \
     libatomic1 \
     postgresql-client \
-    curl \
-    git \
     zstd \
     && rm -rf /var/lib/apt/lists/*
 
--- a/bin/healthcheck.py
+++ b/bin/healthcheck.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python3
+"""HTTP health probe for container health checks.
+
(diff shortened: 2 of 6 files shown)


jfrench9:
@claude please review this PR

PR #1343, opened 2026-09-02
CI sweep: fix release-notes generation, share composite actions and the Claude workflow

jfrench9:
## Why

Release bodies have been bare commit stats since early August. Investigating that surfaced a broader problem: the same workflow code is copy-pasted across the org's repos, so a fix lands in one repo and rots in the others. Three separate defects here had each landed on only one side of a copy.

## What changed

**Release notes were being generated and thrown away.** The response parser read `content[0].text`, assuming the first content block is text. Once `CLAUDE_MODEL` moved to a model that thinks by default, responses lead with a thinking block, so the extraction returned `null` and tripped the empty-content fallback on every release. Verified against the run log for v1.11.12: the failure line is "returned empty", not an API error — the call was succeeding and being discarded.

Text blocks are now selected by type. Also raised `max_tokens` (3000 was shared between thinking and output) and the curl ceiling (60s was tuned for a non-thinking model), and both fallback paths now emit `::error::` with `stop_reason` and the returned block types — this degraded silently for a month, which is the part worth not repeating.

**`select-runner`** — token handling brought in line with the copy the three app repos already use.

**`claude.yml`** — split into triggers (`claude.yml`) and implementation (`claude-review.yml`, a `workflow_call` target). Eleven repos carry this file in two shapes that differ only in runner choice, so every `claude-code-action` pin bump is currently eleven PRs. No input was needed to cover both shapes: `select-runner` treats an unset `RUNNER_LABELS` as `github-hosted` and returns `["ubuntu-latest"]` without an API call.

## Companion PRs

The three app repos switch to the shared composite actions and delete their local copies (~467 lines each). They have no merge-order dependency on this PR — all three actions already resolve on `main`.

Fanning `claude-review.yml` out to the other repos waits until this merges, since it must exist on `main` first.

## Testing

The release-notes fix can only be exercised by a real release — `create-release.yml` has no dry-run path. The jq change was verified locally against both response shapes (thinking-first returns the notes, text-only still works). The next release is the end-to-end test; if it comes back with real notes, it's confirmed.

### DIFF
--- a/.github/actions/select-runner/action.yml
+++ b/.github/actions/select-runner/action.yml
@@ -31,6 +31,10 @@ runs:
     - name: Check if self-hosted runners are available
       id: check-runners
       shell: bash
+      # Pass the token via env rather than inline ${{ }} in the run body, so the
+      # credential is never spliced into the rendered script source.
+      env:
+        GH_API_TOKEN: ${{ inputs.github_token }}
       run: |
         RUNNER_LABELS="${{ inputs.runner_labels }}"
         RUNNER_SCOPE="${{ inputs.runner_scope }}"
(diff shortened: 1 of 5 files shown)


jfrench9:
@claude please review this PR

PR #1344, opened 2026-09-02
feat(ci): parameterize tag-release so other repos can share it

jfrench9:
Adds `product_name`, `product_description`, `project_kind` and `live_url` as optional inputs to `tag-release.yml`, so one copy can serve every repo instead of fourteen near-identical ones.

**All four default to this repo's current values** — the release body here is byte-identical. The only difference is one extra line of prompt context naming the product, which the callers need anyway.

## Why this matters beyond deduplication

The `.content[0].text` defect fixed in #1343 exists in all 14 copies. Rather than patch each one, the other repos now delete their copy and call this. Consolidating also fixes a second problem: several copies tell Claude they are the RoboSystems web app regardless of what the repo actually is — `robosystems-mcp-client`'s prompt says `- App: RoboSystems (Comprehensive financial platform hub web application)` and links to robosystems.ai. Those repos would have generated notes describing the wrong product even after the parser fix.

## ⚠️ Merge this first

Six companion PRs pass the new inputs. If any merges before this one, its next release fails with an undefined-input error rather than merely degrading:

- robosystems-app#368 · roboledger-app#348 · roboinvestor-app#294
- robosystems-python-client#207 · robosystems-typescript-client#221 · robosystems-report-components#49

`robosystems-mcp-client` is deliberately excluded (pending deprecation). `robosystems-core` and `robosystems-xbrl-holon` still carry their own copy — outside this session's working set.

## Testing

Still no dry-run path in `create-release.yml`, so the next real release is the test. The identity logic was verified locally across all three shapes (service / app / SDK): defaults reproduce this repo's title and links exactly, and the optional live link is omitted when `live_url` is empty.

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -7,6 +7,26 @@ on:
         description: 'Branch or commit ref to create tag from'
         required: true
         type: string
+      product_name:
+        description: 'Name used in the release title, e.g. "RoboSystems Service", "RoboLedger App"'
+        required: false
+        type: string
+        default: 'RoboSystems Service'
+      product_description:
+        description: 'One line saying what this repo is. Given to Claude so the notes describe the right product.'
+        required: false
+        type: string
+        default: ''
+      project_kind:
+        description: 'Kind of release, used in the prompt, e.g. "software", "Next.js frontend application", "Python SDK"'
+        required: false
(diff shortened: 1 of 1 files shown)


jfrench9:
@claude please review this PR

PR #1347, opened 2026-09-04
chore(bootstrap): set the frontend repos' OIDC variables from the backend

jfrench9:
## Summary

The frontend apps deploy through the OIDC stack this repo owns, assuming the separate `RoboSystemsGitHubActionsFrontendRole`. Nothing they need from AWS is per-repo — one role ARN, one account, one region — yet each app repo carried a 486-line `bin/bootstrap.sh` to re-derive values this side already holds, and needed its own AWS session to do it. That script could never *create* anything either: it hard-failed when the stack was missing and told you to come back here and run `just bootstrap`.

This adds `configure_frontend_repos()`, which reads `GitHubActionsFrontendRoleArn` from the stack and fans `AWS_ROLE_ARN` / `AWS_ACCOUNT_ID` / `AWS_REGION` out to each frontend repo with `gh variable set --repo`.

## Details

- **Repo names come from the stack's own parameters** (`GitHubAppRepoName`, `GitHubLedgerAppRepoName`, `GitHubInvestorAppRepoName`, `GitHubHolonViewerRepoName`), not a list hardcoded in the script. A fork that renamed its apps stays consistent with the trust policy instead of drifting from it silently.
- Runs in **both** the full and `--oidc` paths — `just bootstrap-oidc` is now the re-run path for frontend identity variables too.
- Skips a repo the `gh` account cannot see, and skips the whole step against a stack that predates the frontend role output, pointing at `just bootstrap-oidc` to update it.
- `set_variable_if_changed` grows an optional target-repo argument and keeps its read-only-when-unchanged behavior, so a re-run against a correct account still writes nothing.
- Written without bash arrays, matching the rest of the script — it stays runnable on a stock macOS bash 3.2.

## Companion PRs

The app repos delete their bootstrap machinery and end up with **no AWS surface at all** — nothing in their `bin/` touches the aws CLI, and their VS Code task lists lose the duplicated SSO logins:

- RoboFinSystems/robosystems-app#373
- RoboFinSystems/roboledger-app#352
- RoboFinSystems/roboinvestor-app#300

Merge this one first: it takes over setting the variables those repos stop setting for themselves.

## Testing

- `bash -n` clean; `shellcheck -S warning` reports only the four pre-existing SC2155 warnings elsewhere in the file.
- Repo-name parsing and the skip paths exercised offline against stubbed `describe-stacks` output, including a stack with no holon-viewer parameter and one with no parameters at all.
- Not yet run against the live stack — the SSO session here has expired. Worth a `just bootstrap-oidc` before merging the app PRs.

## Docs

`Bootstrap-Guide.md` in the wiki needs the matching edit (the "Multi-Repository Setup" and frontend "Bootstrap" sections still say `npm run setup:bootstrap`). Prepared locally, not pushed — the wiki has no PR flow.

https://claude.ai/code/session_01U23eqA9V5R7VYSorFs23Rm

### DIFF
--- a/bin/setup/bootstrap.sh
+++ b/bin/setup/bootstrap.sh
@@ -15,14 +15,17 @@
 #   1. Configures AWS CLI SSO profile (if not exists)
 #   2. Deploys the GitHub OIDC CloudFormation stack (the deploy roles)
 #   3. Sets GitHub variables (role ARN, etc.)
-#   4. First run only: ECR repository, SES identity, application secrets
+#   4. Sets the same identity variables on the frontend app repos, which own
+#      no AWS access of their own — the repos come from the stack parameters
+#   5. First run only: ECR repository, SES identity, application secrets
 #
 # MODES:
 #   full (default)  Everything above. On a re-run the application-config
 #                   step (Secrets Manager, GitHub variables) is skipped unless
 #                   --with-app-config is given — those are the two steps that
 #                   can rewrite a live account, so they never run by accident.
-#   --oidc          Only the OIDC stack + the three identity variables. Use this
+#   --oidc          Only the OIDC stack + the identity variables, here and on
+#                   the frontend app repos. Use this
 #                   after editing cloudformation/bootstrap-oidc.yaml. The stack
 #                   update is previewed as a change set: you see exactly which
 #                   resources change before anything is applied, and a stack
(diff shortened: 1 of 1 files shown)


jfrench9:
@claude please review this PR

PR #1348, opened 2026-09-04
fix(sec): follow full iXBRL continuation chains and dedup facts by precision

jfrench9:
## Summary

Two data defects on the SEC surface, both found by replaying the Filing Ladder benchmark questions against the live `sec` MCP tools.

- **Text index: multi-page notes were cut after one continuation.** The iXBRL disclosure parser read each continuation's `continuedAt` pointer from its content instead of its tag, so every chain stopped after the first hop. On the 3M FY2024 10-K the whole filing yielded 110K characters of narrative; with the fix it yields 343K, Note 6 recovers the PFAS exit-actions paragraph, and Note 19 goes from 570 to 130,272 characters. Reading the pointer from the tag also stops a nested element's pointer from splicing another note's text into the chain.
- **Views: dedup kept the first duplicate, not the most precise.** `financial-statement-analysis` and `build-fact-grid` collapse facts by period identity and kept whichever row the engine returned first. A figure reported on the statement face and again, rounded, in the narrative shares that identity, and the engine's row order is not stable, so the tools answered the rounded one on a fraction of identical calls (3M FY2024 R&D: 1,100 for 1,085). Both queries now project `decimals` and keep the most precise fact per period; equal precision keeps the first row seen.

## Testing

- New parser tests: a four-link chain, and a nested element's pointer that must not hijack the chain. Both fail on the previous parser.
- New dedup tests for both views, plus the shared helper: the 3M R&D pair in either row order, `INF`, and missing precision.
- `just test-code` clean; 121 targeted and 615 downstream tests pass.
- End to end on a local `just sec-load MMM 2025`, comparing the pre-fix and post-fix API process on the same graph:
  - Pre-fix process: 100 identical `financial-statement-analysis` calls returned 10 distinct fact sets, flipping between 1,085 and 1,100 for FY2024 R&D and between 1,154 and 1,200 for FY2023.
  - Post-fix process: 100 identical REST calls returned one fact set, and 20 MCP tool calls all returned 1,085 / 1,154 / 1,160 for FY2024/23/22. The Filing Ladder R&D question scores correct on the shaped-tools rung, and the PFAS narrative question scores correct now that the index holds the paragraph.
  - The local index holds Note 6 at 5,769 characters with the PFAS paragraph. Note 19 is the one section over the existing 50K section cap and is indexed truncated with the marker; the cap is unchanged here.

## Follow-up

The OpenSearch iXBRL disclosure index carries the truncated sections until it is rebuilt: after deploy, backfill the `sec_ixbrl_index` job with `force_reindex: true` over the populated quarters. Facts, the narratives index, and the CDN section copies (which come from Arelle's resolved value and were always complete) are unaffected.

### DIFF
--- a/robosystems/adapters/sec/ixbrl_parser.py
+++ b/robosystems/adapters/sec/ixbrl_parser.py
@@ -9,13 +9,17 @@ ARE the disclosure sections. They wrap the narrative content and contain nested
 `ix:nonFraction` numeric facts. By extracting both the text and the element names,
 we enable bidirectional navigation between graph facts and their filing context.
 
-iXBRL continuation pattern:
+iXBRL continuation pattern (a note that spans pages is a chain, and each
+link's pointer to the next is an attribute on its own tag):
   <ix:nonNumeric name="us-gaap:GoodwillDisclosureTextBlock" continuedAt="f-571-1">
     Goodwill
   </ix:nonNumeric>
-  <ix:continuation id="f-571-1">
+  <ix:continuation id="f-571-1" continuedAt="f-571-2">
     <p>The following table summarizes changes in goodwill...</p>
     <ix:nonFraction name="us-gaap:Goodwill">32,431</ix:nonFraction>
+  </ix:continuation>
+  <ix:continuation id="f-571-2">
+    <p>...and the rest of the note after the page break.</p>
   </ix:continuation>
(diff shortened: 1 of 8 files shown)


jfrench9:
@claude please review this PR

PR #1350, opened 2026-09-04
chore(tunnels): bind offset local ports so a tunnel never collides with the local stack

jfrench9:
## Summary
- SSM tunnels now bind the service port with a `1` prefixed locally — postgres `15432`, valkey `16379`, api `18000`, dagster `18002` — overridable through `LOCAL_PORT_*`. The local Docker stack can stay up while a tunnel is open, and a command left on a default port can never silently reach production.
- The admin CLI targets `localhost:8000` in dev and `localhost:18000` for staging and prod automatically; `--api-url` still overrides both and `--direct` is unchanged.
- Deploy summaries (prod + staging), the `/deploy` command, the CloudFormation and admin READMEs, and the two demo scripts' reset-guard comments describe the same ports. The guards themselves are unchanged (they check host + not-RDS on the database side).

## Test plan
- [ ] `just tunnel prod dagster` binds `127.0.0.1:18002` while the local stack still holds `8002`; the prod UI answers there
- [ ] `just tunnel prod api-internal` + `just admin prod stats` reaches the API on `18000` with no flag
- [ ] `just tunnel prod postgres` + `psql -h localhost -p 15432` connects

### DIFF
--- a/.claude/commands/deploy.md
+++ b/.claude/commands/deploy.md
@@ -62,7 +62,7 @@ A green workflow means the pipeline finished, not that the app is serving. Confi
 
 ```bash
 curl -sf https://<api-host>/v1/status && echo OK      # public API over HTTPS
-# if the API is in internal mode: tunnel first, then curl http://localhost:8000/v1/status
+# if the API is in internal mode: tunnel first, then curl http://localhost:18000/v1/status
 ```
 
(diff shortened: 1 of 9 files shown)


jfrench9:
@claude please review this PR

PR #1351, opened 2026-09-05
feat(sec): keep text blocks in the graph behind a local-only flag

jfrench9:
## Summary

Adds `XBRL_KEEP_TEXTBLOCKS_INLINE`, default `false`, which makes the SEC graph processor store an externalized text-block value in the Fact as the text itself (`value_type=inline`) while still uploading the CDN copy. It exists for one purpose: a local control experiment for the Filing Ladder's raw-Cypher rung (7b), which on the shipped graph cannot reach note narrative because text blocks live in the document index by design.

It must never be set in a real environment. Text blocks are most of a filing's bytes, and moving them out of the graph into OpenSearch is what keeps the shared `sec` graph its size — that decision is the reason the document index exists. The flag is documented as such in `env.py` and `.env.example`.

## Changes

- `config/env.py`: `XBRL_KEEP_TEXTBLOCKS_INLINE = get_bool_env(..., False)` with the rationale.
- `adapters/sec/config.py`: exposes it beside the other XBRL constants.
- `adapters/sec/processors/textblock.py`: `TextBlockExternalizer(keep_inline=...)`; every result now carries `stored_value` (the URL, or the text when kept inline) and the matching `value_type`, built in one `_result` helper so the three return paths agree.
- `adapters/sec/processors/xbrl_graph.py`: stores `stored_value` instead of `url`.
- `.env.example`: the flag, commented out, with the warning.

Known consequence, by design: the iXBRL index's `content_url` resolution keys on `value_type=external`, so a filing processed with the flag on gets no CDN link in search results. Nothing else reads `value_type` for SEC facts.

## Breaking Changes

None. The default path stores exactly what it stored before.

## Testing

`uv run pytest tests/adapters/sec/processors`: 64 passed, including three new tests (default path unchanged; keep-inline stores the text, uploads once, and survives the content cache). `just test-code`: ruff, format, basedpyright and cfn-lint clean. The flag verified to read `true` from the shell through `UV_ENV_FILE=.env.local` and `false` without it.

## Certification

https://claude.ai/code/session_012pEqR1iCHTmMfJJX9bLTE5

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -271,6 +271,9 @@ SECURITY_AUDIT_ENABLED=false
 
 ## Adapter Pipelines
 # SEC_PIPELINE_ENABLED=true
+# Local control experiments only: keep text-block values in the graph as well as on S3.
+# Never in a real environment — text blocks are most of a filing's bytes.
+# XBRL_KEEP_TEXTBLOCKS_INLINE=false
 
 
 
--- a/robosystems/adapters/sec/config.py
+++ b/robosystems/adapters/sec/config.py
@@ -65,6 +65,10 @@ XBRL_EXTERNALIZE_LARGE_VALUES = True
 # Character threshold for externalizing values
 XBRL_EXTERNALIZATION_THRESHOLD = 1024
 
+# Keep the externalized value in the graph too (local control experiments only;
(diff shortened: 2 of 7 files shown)


jfrench9:
@claude please review this PR

PR #1353, opened 2026-09-05
refactor(sec): build the graph from xbrlkit's model, projection and schema

jfrench9:
## Summary

The SEC adapter now consumes xbrlkit for everything that reads a filing, completing the cutover that 0.3.0 was cut for. `XBRLGraphProcessor` loads the filing through `ArelleClient` as before, then hands the `ModelXbrl` to `xbrlkit.parse.to_xbrl_model` and projects the result with `xbrlkit.serialize.lpg.to_graph_tables` — the same ids, columns and rows that `xbrlkit build --format lpg` writes into a single-filing `.lbug`. The platform's own steps (text-block externalization to the CDN, semantic enrichment, the schema-aware parquet writer, association classification) run on top of the projected tables unchanged. The XBRL tables in the graph schema are declared from `xbrlkit.schema`, so the DDL the platform compiles for the shared `sec` graph and the DDL xbrlkit creates locally are one asset. The SEC inline-XBRL transforms come from xbrlkit's vendored registry, which retires the EDGAR plugin fetch from the Docker build, and `arelle-release` is no longer pinned here — it resolves through xbrlkit.

Spec: `local/RoboSystems/specs/adapters/xbrlkit-extraction.md` (phase 4, the adapter re-base).

## Changes

**Processor** (`adapters/sec/processors/xbrl_graph.py`, 2,195 → 775 lines)
- `process()` is now: resolve the instance path, load with `ArelleClient`, `to_xbrl_model(mx, FilingMeta, entity=EntityIdentity)`, `to_graph_tables(model)`, bind the rows as the per-table DataFrames, externalize, enrich, write parquet, classify. The `make_*` walk over Arelle is gone; `FilingMeta` and `EntityIdentity` are built from the same `sec_filer` / `sec_report` dicts the metadata loader supplies, with the same normalization the old code applied (padded CIK, header CIK over `entityId`, empty header strings kept, website falling through to `investorWebsite`, malformed dates to null).
- Externalization is one pass over the Fact rows instead of a branch inside `make_fact`; `XBRL_SKIP_TEXTBLOCK_FACTS` filters the model before projection. `enrich_dataframes` and `classify_associations` are unchanged and read the filing's coordinates from the projected Entity and Report rows.
- `processors/ids.py` keeps the naming helpers only; the twelve `create_*_id` functions go (xbrlkit mints the ids from the same UUID5 namespace). `enrichment.parse_structure_definition` is re-exported from xbrlkit. Dead `fetch_filing` removed.
[...]

PR #1354, opened 2026-09-05
refactor(sec): take the EDGAR, EFTS and Arelle clients from xbrlkit

jfrench9:
## Summary

The SEC adapter's client layer — EDGAR, EFTS and the Arelle load — is now xbrlkit's, which closes phase 2 of `local/RoboSystems/specs/adapters/xbrlkit-extraction.md`: every function that needs only the filing lives in the toolkit, and the adapter keeps what needs the platform (S3, Dagster, the enrichment stack). The part that mattered in production goes with it and gets better: the Arelle schema cache. Arelle re-validates every cached DTS file older than a week with a conditional request, and its check times do not survive the process, so even a warm cache cost one request per file per process — hundreds per filing, against xbrl.org and w3.org. That recheck, not the first fetch, is what throttled the corpus runs. xbrlkit's loader turns it off (a taxonomy URL is versioned; the file never changes), spaces fetches per host, waits out a `Retry-After`, and fails a filing loudly when a document still cannot be resolved instead of returning it with the missing schema's concepts silently gone. Netflix's FY2024 10-K on a warm cache: 34 s and four hosts contacted before, 0.7 s and none after.

## Changes

**Arelle load** (`adapters/sec/client/arelle.py`, 638 → 100 lines)
- `load_filing(path)` calls `xbrlkit.parse.load_model` with the platform's cache directory (`env.ARELLE_CACHE_DIR`, else the repo's `arelle/cache`; a directory the process cannot write falls back to `/tmp` seeded from it), the per-fetch timeout and the offline flag from `adapters/sec/config.py`. `close_filing` releases the model and its controller. The processor calls these two; the `ArelleClient` class, its bundle-copying, its `XDG_CACHE_HOME` mutation, its opener wrapper and the dead EFM disclosure-system selection are gone.
- A `DtsResolutionError` from the load propagates: `process_single_filing_to_memory` records the filing as failed and the source file stays retryable, where the old opener returned an empty body and the filing was indexed with holes.

**EDGAR and EFTS** (`client/edgar.py` 446 → 25 lines, `client/efts.py` removed)
- `edgar_client()` returns xbrlkit's `EdgarClient` on the platform's User-Agent and timeouts (`xbrlkit_config()` in `adapters/sec/config.py`). The metadata loader's API fallback, the download step's ticker map, complete-submissions build and incremental refresh, and the pipeline script's top-companies list use it; the incremental refresh no longer carries its own aiohttp GET and 429 handling.
- EFTS discovery is xbrlkit's synchronous `EftsClient`, run in a thread from the async download step and the bulk downloader so the event loop that drives the parallel S3 uploads stays free. Hits are `EftsHit` (`accession`, `form`).
- `client/downloader.py` and `client/rate_limiter.py` stay: the parallel download to S3 needs the platform.

**Schema cache** (`Dockerfile`, `arelle/bundles/`, `scripts/arelle_cache_manager.py` removed)
- The image seeds `ARELLE_CACHE_DIR` with `xbrlkit cache extract` from the bundle, in Arelle's own layout (the old bundle's `host/path` layout was one Arelle never read directly; only the removed opener wrapper bridged it).
- The bundle is rebuilt as the throttling hosts' complete set — every xbrl.org and w3.org file the 26-filing corpus and the previous bundle needed, 125 files, 124 KB (was 58 files across four hosts, 984 KB). The tolerant hosts fill in on first use. `adapters/sec/README.md` carries the refresh procedure.

**Examples and docs**
- `examples/_common/validate.py`: `validate_arelle` loads through `load_filing` and runs Arelle's `ValidateXbrl` itself (it imported the removed `ArelleClient`; caught by the review). A DTS the loader cannot resolve is reported as load errors in the validation report rather than a traceback. The two README usage snippets that still named `SECClient` are updated.

**Tests**
- `tests/adapters/sec/arelle/test_client.py` rewritten for the function-based load (cache-directory selection, fallback seeding, the settings handed to xbrlkit, error propagation); `test_edgar.py` and `test_efts.py` removed with their modules (their subjects are tested in xbrlkit); the downloader, metadata and processor tests adapted; the unused `mock_sec_client` fixture removed from `conftest.py`.

**Dependency**
- `xbrlkit>=0.4` (RoboFinSystems/xbrlkit#23): the cache-first loader, `DtsResolutionError`, `EftsClient`, `company_tickers`, `complete_submissions`.

## Breaking Changes

None. `robosystems.adapters.sec` no longer exports `SECClient`, `ArelleClient`, `EFTSClient`, `query_efts` or `enable_test_mode`; nothing outside the adapter imported them. No API surface is touched.

## Testing

- Identity gate, previous processor vs this branch, 26-filing benchmark corpus: 777 table comparisons, row-identical with identical arrow types; the only difference is the duplicate-arc collapse recorded in #1353. Every load went through `load_filing` with nothing unresolved.
- `just test`: 14,108 passed, 42 skipped.
- Live, not in the suite: the seed bundle extracts through `xbrlkit cache extract` and reports seeded; the 26-filing gate above ran every load through `load_filing` against the harness's cache with nothing unresolved.

## Certification

https://claude.ai/code/session_012pEqR1iCHTmMfJJX9bLTE5


jfrench9:
@claude please review this PR

PR #1355, opened 2026-09-05
feat(roboledger): read journal entries on their own terms

jfrench9:
## Problem

The ledger's list surfaces are transaction-centric: `transactions` walks the `transactions` table and hangs entries off each row. An entry with **no parent transaction** therefore appears on no list surface in the product.

Parentless entries are not an anomaly. `Entry.transaction_id` is nullable by design, the schedule engine (`create_closing_entry` / `create_manual_closing_entry`) and the event handlers never set it, and `Entry.triggered_by_event_id` exists precisely to carry the event chain for entries that have no transaction to carry it.

The practical consequence: **everything a period close posts was invisible.** The close-review outbox filters `status = 'draft'`, so posting an entry removed it from the only surface that ever showed it. The trial balance and the statements saw those entries the whole time — so the numbers were right and the audit trail behind them could not be browsed, which is the worst combination for trust.

Found on live books: two closed months, 66 posted schedule-derived entries, none of them reachable from any list view.

## Change

Adds a `journalEntries` GraphQL field — the entry-centric read:

- `transactionId` is **projected**, so null means *standalone*, not *missing*.
- Filters on date range, `status`, `type`, `provenance` and parent `transactionId`. `provenance: "schedule_derived"` answers "what did the close post?" in one query.
- Amounts in dollars, matching the transaction reads it sits beside. Pagination counts **entries**, not line-item rows.

The entry projection and row→entry grouping move into `reads/journal_entries`, shared with `list_period_drafts`, which becomes a caller. The close-review query and the journal query now cannot drift apart. Drafts keep reporting cents — the shared fetch is unit-neutral and that response's contract is unchanged.

## Verification

- Full suite green: 14120 passed, 42 skipped; ruff, format, basedpyright, cf-lint all clean.
- The unit tests are mock-based (matching `test_period_drafts`), which means they never execute SQL — so the statement was **also run against real Postgres**: both orderings, every filter bound simultaneously, and a seeded parentless posted entry returned alongside a parented one, with the drafts outbox still returning only the draft in cents. That check mattered here because the whole point of this read is a row shape no existing query returned.
- New tests cover what nothing exercised before: a posted entry with no parent transaction, and parentless + parented entries returned together. That gap is why this shipped invisible.

## Scope

Backend only. The `LedgerClient` facade method and the transactions-page tab ride the next client cut — generated-tier surface, additive, no deprecation cycle. Design doc is `specs/roboledger/journal-entry-read-surface.md` (private vault).

https://claude.ai/code/session_01UdMBS1zchAGU4JszgXtMXY

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,7 +112,7 @@ dependencies = [
 [project.optional-dependencies]
 dev = [
     # RoboSystems client for demo and testing.
-    "robosystems-client>=1.13.1,<2.0",
+    "robosystems-client>=1.14.0,<2.0",
     # Testing framework
     "pytest>=9.0.0,<10.0",
     "pytest-asyncio>=1.4.0,<2.0",
--- a/robosystems/graphql/resolvers/ledger.py
+++ b/robosystems/graphql/resolvers/ledger.py
@@ -44,6 +44,7 @@ from robosystems.graphql.types.ledger import (
   EventBlock,
   FiscalCalendar,
   LedgerEntity,
+  LedgerJournalEntryList,
   LedgerSummary,
   LedgerTransactionDetail,
   LedgerTransactionList,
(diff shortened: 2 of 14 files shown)


jfrench9:
@claude please review this PR

jfrench9:
@claude can you review the latest changes

PR #1356, opened 2026-09-05
fix(roboledger): derive Entry.provenance, and give non-economic events a class

jfrench9:
Two origin-tracking corrections on the ledger spine, both surfaced by the journal-read work in #1355. Migration **0035**, fanned out over every tenant schema.

## 1. `Entry.provenance` was hardcoded

`create_journal_entry` set `provenance="manual_entry"` for every entry regardless of origin. So every entry that arrived over a connection claimed to be hand-entered, and `source_sync` — a value the CHECK constraint allows — had **never been written by anything**. On the live books that's **1726 QuickBooks-synced entries**.

This directly weakened the filter that just shipped: the Journal's Source filter could answer *"what did the close post?"* but not *"what came from QuickBooks?"*

The column exists to answer "where did this come from", and the write already receives the answer — `body.source`, which the event handler sets from `Event.source`. It was being discarded. Now derived:

| Source | Provenance |
|---|---|
| `manual` / `native` / absent | `manual_entry` |
| `system` | `system_computed` |
| `schedule` | `schedule_derived` |
| **anything else** | `source_sync` |

Closed on the platform side, open by default on the adapter side — because `Event.source` deliberately carries no CHECK (registering a connection opens a source name with no schema change), so the adapter half can't be an enumerated list.

**Backfill** reads each entry's linked event. Conservative and idempotent: only entries whose provenance is exactly `manual_entry`, only where a linked event exists, and a `manual` event maps straight back — so genuinely hand-entered rows are untouched and re-running is a no-op. Entries with no linked event are left alone; there's nothing to derive from.

Two related touches: `create_manual_closing_entry` now takes `provenance` instead of hardcoding it, and the disposal handler passes `event_handler` ("manual" there means *not schedule-derived*, a different question from who wrote it). The reversal path deliberately keeps `manual_entry` — a reversal is an act by whoever asked for it, not a re-sync — with a comment, since it otherwise looks like the bug being fixed.

## 2. `event_class` had no home for a non-economic occurrence

The vocabulary was `economic | support`. Economic asserts a resource flow; support means *supporting* an economic event — controls, approvals, reconciliations, inquiries. Anything that is neither had nowhere honest to go.

That isn't hypothetical. On the live books **`support` has never been used once**, and `schedule_created` — setting a schedule up, which moves nothing and posts nothing — is filed as `economic/other`. The gap was already being papered over.

Adds **`operational`** with `pipeline | engagement | schedule | other`, and re-files the existing `schedule_created` rows. The recognition events a schedule later emits stay `economic/recognition` — those do drive the GL. Filing a non-economic occurrence as economic asserts a resource flow that did not happen, which is exactly what this column is for.

Design context: `specs/ledger/non-financial-events.md` §3 (private vault). This clears the blocker it names; the measurement half is separate and unbuilt.

## Verification

Migration applied locally and checked against Postgres rather than only through mocks:

- constraints land on `public` and fan out per tenant;
- `lead_received` / `operational` / `pipeline` is **accepted** where it was previously impossible;
- `support` / `pipeline` is **still correctly rejected** — the vocabulary kept its shape rather than being opened up;
- backfill maps `quickbooks→source_sync`, `manual→manual_entry`, `system→system_computed`, leaves event-less entries alone, and is a no-op on a second run.

`just test-all` green: 14131 passed, 42 skipped; ruff, format, basedpyright, cf-lint clean.

**Note on the autogenerated migration**: `just migrate-create` missed the CHECK constraint change entirely (as CLAUDE.md warns it does) and emitted 24 spurious FK/type operations that would have been destructive. It was discarded and 0035 written by hand off the `0033` fanout precedent.

**Note on a changed test**: one existing assertion pinned the misclassification — `all(e.event_class == "economic" for e in events)` over both schedule event types at once. It now distinguishes them, which is the thing actually worth asserting.

## Downgrade

The constraint change reverts. The provenance repair deliberately does **not** — the repaired values are the true origins, and restoring the mislabel would be re-introducing the bug.

https://claude.ai/code/session_01UdMBS1zchAGU4JszgXtMXY

### DIFF
--- a/migrations/extensions/versions/0035_operational_events_and_entry_provenance.py
+++ b/migrations/extensions/versions/0035_operational_events_and_entry_provenance.py
@@ -1,0 +1,172 @@
+"""operational event class, and repair Entry.provenance
+
+Two related corrections to origin-tracking on the ledger spine, both
+fanned out over every tenant schema.
+
+**1. `event_class = 'operational'`.** The class vocabulary was
+`economic | support`, where economic asserts a resource flow and support
+means *supporting an economic event* (controls, approvals,
+reconciliations, inquiries). A business occurrence that is neither had
+nowhere honest to go: `schedule_created` — setting up a schedule, which
+moves nothing and posts nothing — was filed as `economic/other`, and a
+CRM lead had no legal category under either class. Adds a third class
+with `pipeline | engagement | schedule | other`, and moves the existing
+`schedule_created` rows onto it. The recognition events a schedule later
+emits stay `economic/recognition`; those do drive the GL.
+
+**2. `Entry.provenance` repair.** `create_journal_entry` hardcoded
+`manual_entry` regardless of origin, so every entry that arrived over a
+connection claimed to be hand-entered and `source_sync` — a value the
+CHECK constraint allows — had never been written by anything. The write
+already receives the origin (`body.source`, which the event handler sets
+from `Event.source`); it was simply discarded. The code now derives it,
+and this backfills the rows already written by reading each entry's
+linked event.
+
+The backfill is conservative and idempotent: it only touches entries
+whose provenance is exactly `manual_entry`, only where a linked event
+exists, and it maps a `manual` event straight back to `manual_entry`, so
+genuinely hand-entered rows are unchanged and re-running is a no-op.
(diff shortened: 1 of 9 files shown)


jfrench9:
@claude please review this PR

PR #1352, opened 2026-09-05
refactor(sec): read filing text through xbrlkit.text; index long sections in parts

jfrench9:
## Summary

The SEC adapter's filing text layer — the iXBRL disclosure parser, the 10-K/10-Q Item extractor and the HTML-table converter — now lives in the public `xbrlkit` package (`xbrlkit.text`, 0.3.0), which the platform depends on for the first time. The three modules and their tests leave this repo; the two index assets import the parsers from xbrlkit. The move carries the parser fixes measured on the Filing Ladder's 26-filing corpus (nested `ix:continuation`s registered, nested text blocks as sections, table-of-contents rows and cross-references rejected as section starts, a 10-Q's Part I and Part II Items kept apart), and long sections are now indexed in parts instead of truncated at 50,000 characters. Full record: `local/RoboSystems/specs/adapters/xbrlkit-extraction.md` (vault).

## Changes

**adapters/sec** — `ixbrl_parser.py`, `html_table.py`, `narrative_extractor.py` and their three test files removed; `pipeline/text_index.py` imports `iXBRLParser` / `NarrativeExtractor` from `xbrlkit.text`. Both index assets: `part_size` (default 25,000) replaces `max_section_length` on `SECNarrativeIndexConfig` / `SECiXBRLIndexConfig`; each part is its own document with `part`, `part_count`, a `parent_document_id` shared by the section's parts and a `next_document_id`; an unsplit section keeps the document id it always had (`_part_document_ids`); a `force_reindex` run deletes the accession's documents before writing, since the part count decides the ids; the CDN copy is one object per part. READMEs updated.

**operations/search** — `client.py`: the four part fields in the mapping, added to an existing index on startup with `put_mapping` (additive, idempotent); `delete_by_accession`; the `section` filter matches the value as given and lower-cased, so an iXBRL disclosure's element qname can match — lower-casing alone never could. `service.py`: the part fields on hits and sections; `parent_document_id` from the document when present, else the user-doc derivation.

**models/api/search.py** — `SearchHit` and `DocumentSection` gain `part`, `part_count`, `next_document_id` (and `parent_document_id` on the section); `section` filter description names the qname form.

**middleware/mcp/tools/search_tools.py** — `search-documents` and `get-document-section` descriptions say a long section is paged and how to follow `next_document_id`; the `section` parameter documents the qname form.

**pyproject / uv.lock** — `xbrlkit>=0.3`.

## Breaking Changes

[...]

PR #1357, opened 2026-09-05
fix(sec): index a ticker-less filer under its CIK, not the string "<NA>"

jfrench9:
## Summary

A filer without a ticker was indexed with `entity_ticker` set to the string `"<NA>"` instead of its CIK. The two text index assets build their entity lookup with `str()` on each parquet cell, and `str()` of a pandas missing value is `"<NA>"` — truthy, so the `ticker or cik` fallback never fired. Found auditing the local index for the Filing Ladder corpus (U.S. Steel's 119 documents); on prod it is every ticker-less filer.

## Changes

**adapters/sec/pipeline/text_index.py** — `_cell()` turns a parquet cell into text with a missing value (`None`, NaN, pandas NA) as an empty string; both assets' entity lookups (`sec_narratives_indexed`, `sec_ixbrl_disclosures_indexed`) use it for ticker, name and CIK, so a missing ticker falls through to the CIK as intended.

**tests/adapters/sec/pipeline/test_text_index.py** — `TestCell`: missing values of each kind are empty; values are their text.

## Breaking Changes

None. No API surface changes; the indexed `entity_ticker` value for ticker-less filers becomes the CIK, which is what the `entity` search filter and the hit models already expect.

## Testing

- `tests/adapters/sec/pipeline/test_text_index.py`: 13 passed.
- Live, on the local stack: the Filing Ladder corpus (26 filings) force re-indexed with this change: U.S. Steel's 119 documents carry `entity_ticker` `0001163302` (its CIK), 0 of 2,484 documents carry `<NA>`, and the other 25 accessions are unchanged (section sets equal to the parser's output, embeddings on every document). The `entity` search filter finds the filer by name and by CIK as before.

Existing documents on prod keep `"<NA>"` until their filings are re-indexed; the corpus re-index owed after #1352 covers them.

## Certification

https://claude.ai/code/session_014adkqDsj2Mb2XEri6HDSNa

### DIFF
--- a/robosystems/adapters/sec/pipeline/text_index.py
+++ b/robosystems/adapters/sec/pipeline/text_index.py
@@ -31,6 +31,7 @@ import zipfile
 from typing import Any
 
 import boto3
+import pandas as pd
 from dagster import AssetExecutionContext, BackfillPolicy, MaterializeResult, asset
 
 from robosystems.config import env
@@ -111,8 +112,6 @@ def _cell(value: Any) -> str:
   ``str()`` on a pandas missing value is the truthy ``"<NA>"``, which once
   reached the index as the ticker of every filer without one.
   """
-  import pandas as pd
-
   if value is None or pd.isna(value):
     return ""
   return str(value)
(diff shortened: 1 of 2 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Both observations taken in 0e60c535: `filing_date` goes through `_cell()` at both sites, and pandas is imported once at module level.

PR #1358, opened 2026-09-05
feat(examples): load a filing corpus and audit the text index against the parsers

jfrench9:
## Summary

A repeatable check that a known set of filings is loaded and indexed completely: `examples/sec_corpus_check` runs the pipeline over a corpus directory in the Filing Ladder's layout, then audits what OpenSearch holds for each accession against what the parsers produced. With the ladder's own `bin/check_text_layer.py` (parsers against each filing's text-block facts), a parse defect and an index defect are told apart. The two together found the three defects fixed today: the repeated-tag and `ix:exclude` parser defects (xbrlkit #24) and the `"<NA>"` ticker (#1357).

## Changes

**examples/sec_corpus_check/main.py** — four subcommands on `--corpus <dir>`:
- `load` — the pipeline's stages as `just sec-load` runs them, with the download stage selecting filers **by CIK per quarter** (a `SECPipeline` subclass): a delisted filer no longer resolves through the SEC ticker map, and an empty filter would download the whole quarter, so a quarter with no corpus filer is skipped outright. Resets the local SEC graph first, as the pipeline does.
- `audit` — per accession: section sets and part counts against the ladder check's JSON (`--check`), part chains (`part` 1..n, shared `parent_document_id`, `next_document_id` links), the metadata the index assets add (ticker, name, CIK, form, fiscal year uniform across the filing's documents; a ticker that is empty or a missing-value string is flagged), embeddings on every document, one CDN copy per source fetched. Disclosure documents whose fact was stored inline have no CDN copy by design and are not flagged.
- `probe` — two searches per filing through `POST /v1/graphs/sec/search` (the MD&A Item and the accounting-policies disclosure), reading the top hit's section through `GET …/search/{id}`.
- `reindex` — force re-indexes both source types for the corpus quarters.

**examples/sec_corpus_check/README.md** — what each command checks. **justfile** — `sec-corpus-check *args`.

## Breaking Changes

None. Tooling under `examples/`; no platform code.

## Testing

- Live, on a fresh local stack with the Filing Ladder's 26 filings (24 10-Ks, 2 10-Qs, 2024–2025): `load` — 26 of 26 processed, 0 failed stages, 29.6 minutes; `audit --check` against the ladder's check on xbrlkit #24 — 26 filings, 0 problems (2,484 documents, every parsed section and part present, embeddings on all, chains intact); `probe` — 52 of 52 hits, U.S. Steel found by its CIK; `reindex` — 12 jobs, all succeeded. Before #1357 the audit flagged U.S. Steel's 119 documents with `entity_ticker` `"<NA>"`.

## Certification

https://claude.ai/code/session_014adkqDsj2Mb2XEri6HDSNa

### DIFF
--- a/examples/sec_corpus_check/README.md
+++ b/examples/sec_corpus_check/README.md
@@ -1,0 +1,33 @@
+# SEC corpus check
+
+Load a known set of filings through the SEC pipeline on the local stack, then
+audit what the text index holds against what the parsers produced. Built for the
+Filing Ladder's 26-filing corpus, whose `bin/check_text_layer.py` checks the
+parsers (`xbrlkit.text`) against each filing's own text-block facts; this checks
+the platform's index against those parsers, so a parse defect and an index
+defect are told apart.
+
+```bash
+just sec-corpus-check load    --corpus ../filing-ladder/data
+just sec-corpus-check audit   --corpus ../filing-ladder/data --check text-layer.json
+just sec-corpus-check probe   --corpus ../filing-ladder/data
+just sec-corpus-check reindex --corpus ../filing-ladder/data
+```
+
+The corpus directory holds one folder per accession with a `meta.json`
+(`accession`, `cik`, `form`, `filing_date`, `ticker`) — the layout
+`filing-ladder materialize` writes.
+
+| Command | What it does |
+|---|---|
+| `load` | `just sec-load`'s stages, with the download stage selecting filers **by CIK per quarter** — a delisted filer no longer resolves through the ticker map, and an empty filter would download the whole quarter. Resets the local SEC graph first, as the pipeline does. |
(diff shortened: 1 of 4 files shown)


jfrench9:
@claude please review this PR

jfrench9:
All three taken in 5104cce2: a probe hit is the filer's when its ticker field or its name matches (the ticker field carries the CIK for a ticker-less filer after #1357; the name covers an index written before it), the host copy of each run config is unlinked after the `docker cp`, and documents of a source type the audit does not check are reported instead of skipped.

One correction to the premise: the CIK path was exercised — U.S. Steel (0001163302) has no ticker in the corpus, and its two probes hit because the index under test already carried #1357. Re-run after the change: 52 of 52 probes, audit 26 filings with 0 problems.

PR #1361, opened 2026-09-06
feat(serialization): Tavi compiled model as a fourth report download flavor

jfrench9:
## Summary

Adds the Project Tavi compiled model as a fourth download flavor for a published RoboLedger report (`tavi`, GraphQL `TAVI`, file `g{n}.tavi.json`) beside the JSON-LD, the holon and the XBRL 2.1 zip. It is the same standards form the SEC pipeline publishes per filing, so `@robosystems/report-components` renders it through the adapter it already has, with no RDF step — the first consumer is roboledger-app's report page.

The emitter is xbrlkit's, not a second one: a new bundle → `XbrlModel` bridge (`operations/serialization/model.py`) re-expresses the `StatementBundle` as the model xbrlkit's projections read, so a filing and a ledger report go through one implementation and come out byte-comparable. Design and review: `local/RoboSystems/specs/reporting/tavi-serialization.md` (§3 mapping, §6.2 this repo's half, §11 why the bridge is the first brick of moving the encoders into xbrlkit).

Depends on xbrlkit 0.4.2 (RoboFinSystems/xbrlkit#25, released) and robosystems-client 1.15.0 (the `tavi` format), both pinned here. Every report demo's artifact set now includes `{stem}.tavi.json`, downloaded through the published client beside the other three flavors.

## Changes

- `operations/serialization/model.py` (new) — `bundle_to_xbrl_model`: concepts and headings, the wire item-type table (`ratio`/`multiple` → pure, `days` → decimal, untyped → the XBRL 2.1 monetary/string rule), periods, units, facts with XBRL lexical values, networks from the presentation and calculation links with roots computed per link. Role definitions are composed SEC-shaped (`0001 - Statement - Balance Sheet`, `0100 - Disclosure - <note>`) so the adapter titles, kinds and orders sections as the holon renderer does; the raw structure name rides as documentation; a structure with no `role_uri` gets `https://robosystems.ai/role/{structure_id}`. Markdown notes become HTML text blocks with raw HTML escaped (`markdown-it-py`, now a direct dependency). Entity under `http://robosystems.ai/entity`.
- `operations/serialization/xbrl/tavi.py` (new) — `serialize_to_tavi` (compact JSON, the SEC writer's serialization; gap report logged at debug) and `TAVI_OMITTED_CONTENT`.
- `flavors.py`, `xbrl/__init__.py`, `serialization/__init__.py` — `XbrlFlavor.TAVI`, the dispatch arm, exports.
- `bundle.py` — `BundleElement.namespace_uri`, resolved once by the producer (`namespace_uri_for`: concept IRI minus local name → canonical context → minted), since `namespace` has always held the prefix.
- `reads/reports.py` — `_materialize_and_presign_tavi`, gated on publication like the holon (a received copy carries the sender's Tavi and no flat bundle); `_refuse_local_rederivation` for received reports; `.tavi.json`, `application/json`, `omitted_content` declared on hits and misses.
- `commands/reports.py` — a share builds whichever derived artifact (holon, Tavi) is missing off one bundle and copies each under its own media type (`PUBLICATION_MEDIA_TYPES`); a failing encoder no longer loses the others. Docstring `v{n}` → `g{n}`.
- `graphql/types/ledger.py` — `ReportDownloadFormat.TAVI`; `models/api/extensions/reports.py` — `omitted_content` documents the Tavi's names.
- Tests — `test_model_bridge.py` (the mapping), `test_tavi_emitter.py` (what a consumer sees), `test_cross_encoder_equivalence.py` now asserts the fact tuple across three projections, `test_report_download.py::TestTavi` plus the received-report parametrizations, `test_publication_artifacts.py` (new; the share path), the revoke test covers the copied `.tavi.json`.

## Breaking Changes

None. Additive: a new enum value and download flavor, a new optional bundle field, a new direct dependency already in the lock. SDK impact is generated-tier — `ReportDownloadFormat` gains `TAVI` — so both clients ride a minor with the regen (Python: `just refresh-schema` will flag the enum; TS: the union gains `'TAVI'`).

## Testing

`just test-code` green (lint, format, basedpyright). Targeted suites green on the published xbrlkit 0.4.2: `tests/operations/serialization` (bridge, emitter, three-way equivalence, existing encoder suites), `tests/operations/roboledger/reads/test_report_download.py`, `tests/operations/roboledger/commands` including the DB-backed share tests — 228 + 319 passed. Full `just test` runs in CI.

End to end on a rebuilt local stack: `just demo-roboledger` published the demo FY2025 report and downloaded all four flavors through robosystems-client 1.15.0 (Tavi 66 KB beside the 194 KB holon; SHACL conforms, Arelle valid). Parsed through `@robosystems/report-components` 0.5.1, the Tavi and the holon of the same report yield the same four sections, the same entity name, and the same balance-sheet rows; the Tavi parses in 2 ms to the holon's 36 ms in node. That file is now the fixture in report-components and roboledger-app.

## Certification

https://claude.ai/code/session_01TBKj1VsfBpKsFY7PPcrCFu


jfrench9:
@claude please review this PR

jfrench9:
Review follow-ups, all on the branch now:

- **Blocker** — resolved by `793b4ca8` (`xbrlkit>=0.4.2` + lock) once 0.4.2 was on PyPI; `7330b198` pins `robosystems-client>=1.15.0` with it. CI is green on both. The eager import in `serialization/__init__.py` is deliberate: the package's exports are the encoders' public surface, and with the lock pinned to the release the package imports or it doesn't, rather than one flavor failing at first use.
- **Suggestion 1** — `79c51275` documents in `model.py` that the bridge writes the entity id into the model's `cik` field and the report id into `accession`, and that the entity `scheme` is what keeps the wire output off `cik:`. Neutral names are xbrlkit's to add (spec §11.3).
- **Suggestion 2** — deliberate: a structure's presentation and calculation links share its role by construction (the producer binds calc arcs to the rendered network), so the minted role is the same for both and they land in one group. Now said in `network_role`'s docstring.
- **Question** — the follow-up landed on this branch rather than a separate PR, so `main` never sees the gap.

PR #1362, opened 2026-09-06
fix(reports): persist report facts at cents precision

jfrench9:
## Summary

Report facts now persist at cents precision. The ledger stores integer cents; the report engine converts them to dollars and adds subtotals in float, and `facts.value` is `double precision`, so a subtotal such as `52585 + 5400.02` was stored as `57985.020000000004` and every published flavor — JSON-LD, holon, Tavi, XBRL 2.1 — printed it faithfully. In the demo tenant 12 of 335 numeric facts carried a tail, all subtotals (Assets, Net Income, Stockholders' Equity, operating cash flow); the leaves were clean. Found in the first Tavi off a ledger report (#1361); traced to the database rather than the serialization layer, which was printing what it was given.

## Changes

- `operations/roboledger/reports/statement_sets.py` — `to_cents_precision(value)` and its use in `_stamp_facts_into_sets`, the one choke point both report publication and canonical stamping write rows through. Every value there is dollars (the row is stamped `unit="USD"`), so rounding to cents removes float noise and no real precision: the ledger has none below a cent. The published files keep claiming `decimals="INF"` (the bundle producer hardcodes it), while the row's `decimals` stays NULL and the graph coalesces that to `-2` — a pre-existing three-way disagreement about precision metadata, filed separately; this PR is about `value`.
- `tests/operations/roboledger/reports/test_statement_sets.py` — a float-summed subtotal persists as `57985.02` (and `0.1 + 0.2` as `0.3`); negative and whole amounts are untouched; the helper's contract.

No column change, no migration, no encoder change. Existing reports keep their tails until regenerated.

## Breaking Changes

None.

## Testing

`tests/operations/roboledger` green (1046 passed); ruff, format and basedpyright clean on the changed files. Then on the restarted local stack, `just demo-roboledger` regenerated the demo's FY2025 report and re-downloaded all four flavors: the new report's 41 numeric facts have zero tails in the database, the JSON-LD, holon and Tavi contain no float noise, the totals read `57985.02` / `7385.02` / `57185.02`, SHACL conforms and Arelle validates. Full `just test` runs in CI.

## Certification

https://claude.ai/code/session_01TBKj1VsfBpKsFY7PPcrCFu

### DIFF
--- a/robosystems/operations/roboledger/reports/statement_sets.py
+++ b/robosystems/operations/roboledger/reports/statement_sets.py
@@ -370,6 +370,11 @@ def _pre_create_report_fact_sets(
   session.flush()
 
 
+def to_cents_precision(value: float) -> float:
+  """A dollar amount rounded to cents — the ledger's own precision."""
+  return round(value, 2)
+
+
 def _stamp_facts_into_sets(
   session: Session,
   facts,
@@ -391,6 +396,15 @@ def _stamp_facts_into_sets(
   Fact row per owning structure, each pinned to that structure's
   FactSet. This is what lets the CF block's calc walker resolve
   NetIncome locally without cross-structure fact lookup.
+
+  Every value through here is dollars (the row is stamped ``unit="USD"``)
+  derived from the ledger's integer cents, so it is rounded to cents
+  before it is written. The pivot and the subtotal derivations add in
+  float, and a sum such as ``52585 + 5400.02`` lands on
+  ``57985.020000000004``; ``facts.value`` is ``double precision`` and
+  would keep that tail, and every published flavor prints the stored
+  double faithfully. Rounding here removes float noise, never real
+  precision — the ledger has none below a cent.
   """
   for fact in facts.facts:
     for structure_id in element_to_structures.get(fact.element_id, ()):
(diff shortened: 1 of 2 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Good catch on `decimals`, and worth stating precisely because it is three different claims for the same fact:

- `facts.decimals` is NULL on every numeric fact in the demo tenant (335/335) — the stamp path never sets it, and `ReportFact` has nothing to source it from.
- The graph coalesces that NULL to `-2` at materialize (`materialize.py:1052`), which claims accuracy to the nearest hundred.
- The bundle producer hardcodes `decimals="INF"` (`serialization/bundle.py:1157`), so the JSON-LD, holon, Tavi and XBRL 2.1 all claim infinite precision. That is what the PR body's "stays INF" referred to — the files, not the row — and I've reworded it.

After this fix the true precision of a statement fact is `2`. Reconciling the three (stamp `decimals='2'` on statement facts; have the bundle read the row instead of hardcoding) is a pre-existing precision-metadata item, filed as its own row in the reporting README rather than widened into this PR, which stays about `value`.

PR #1363, opened 2026-09-08
chore(examples): regenerate demo samples with the Tavi bundle

jfrench9:
## What

The committed `sample_output/` sets predated the Tavi serialization (#1361), so none of the five carried a `.tavi.json`. The demos emit one — the reference copies never picked it up.

Regenerates all five sample sets from a clean run against v1.11.17, and fixes the reason they'd have stayed stale.

## Why they were missing

The refresh command documented in three READMEs is `cp output/*.{md,jsonld,zip} sample_output/` — a glob that structurally excludes `.tavi.json`. The seattle README documents the file in its `download-bundles` row while its own refresh command drops it, so a by-the-book refresh would have kept missing it indefinitely. Widened to `*.{md,jsonld,tavi.json,zip}`.

Also corrected two stale artifact counts in the surrounding prose ("both bundles", "all five artifacts") that predated the holon and DataBook landing.

## Verification

All five demos re-run green; every set SHACL-conformant and valid XBRL 2.1:

| Demo | tavi.json | SHACL | Arelle |
| --- | --- | --- | --- |
| roboledger | 66 KB | ✅ | ✅ |
| coffee-roaster | 88 KB | ✅ | ✅ |
| saas-startup | 84 KB | ✅ | ✅ |
| seattle-method | 68 KB | ✅ | ✅ |
| world-online | 70 KB | ✅ | ✅ |

`tests/operations/serialization/` + `test_report_download.py` — 224 passed. The samples are gated in CI by `test_sample_bundles_shacl.py`, which globs `*.jsonld` and validates each against `frameworks/ontology/v1/shapes.ttl`.

Promotion was stem-scoped, so the uncommitted `preclose` / `postclose` / `driftline-prod` variants in the coffee-roaster and saas-startup output dirs stayed out.

## Notes for the reviewer

The diff is large but almost entirely regenerated artifacts — each run stamps fresh graph/report IDs, so every sample churns. The reviewable surface is the three README files plus the five new `.tavi.json`.

Reproducing this locally needed two environment fixes, **neither of which required a source change**:

- `just restart` — the running stack served a GraphQL schema older than its own bind-mounted code, so `ReportDownloadFormat` exposed no `TAVI` despite it being present in `robosystems/graphql/types/ledger.py:724`.
- `just rebuild` — the image carried xbrlkit 0.4.1 against the `>=0.4.2` pin in `pyproject.toml:62`, so `to_tavi_report()` raised `TypeError` on the `description` kwarg.

Worth a second opinion on the second one: I did not establish why the image had 0.4.1 when the lock pins `>=0.4.2`. If that image came from a registry pull rather than a local build, the same skew could hit anyone running `just start` today.

Two things deliberately not done:

- No test pinning the Tavi samples. `test_sample_bundles_shacl.py` globs `*.jsonld` only, so the new files are reference evidence with no CI gate. The emitter itself is covered by `test_tavi_emitter.py` and `test_cross_encoder_equivalence.py`, so a sample-level gate is optional — but it is an asymmetry with the JSON-LD bundles.
- No changes to the coffee-roaster or saas-startup READMEs, which don't document a refresh command at all.

https://claude.ai/code/session_012gBgndhHN1pg24apjVraGH


jfrench9:
@claude please review this PR

PR #1367, opened 2026-09-09
fix(graph): exempt deleted rows from the orphan-registration sweep

jfrench9:
## Summary

The `OrphanedGraphRegistrations` alarm has been firing on graph-registry rows for graphs that no longer exist. `cleanup_stale_graphs` removes a deleted row only once it is older than `STALE_GRAPH_DAYS`; a deleted row still inside that window fell through to the orphan check, which had no `status` predicate, and was stamped and counted as soon as the instance it named left the instance registry.

That instance is always about to leave — a deleted graph's writer is released with the graph and recycled by the ASG well inside seven days — so every rolling fleet replacement that followed a deletion raised the alarm for a week and then healed itself. This scopes the orphan check to rows that are actually routing.

## Changes

**`robosystems/operations/graph/infrastructure.py`**

- `cleanup_stale_graphs`: add `status != "deleted"` to the `instance_missing` predicate, alongside the existing shared-repository exemption. Both exemptions say the same thing — the row is not a routing pointer — so they sit together and are documented together.
- Docstring and inline comment updated to record why, including the failure mode the predicate prevents (a page that clears itself after `STALE_GRAPH_DAYS`).
- The clear path needed no change and now covers the other half for free: a row stamped while it was live and subsequently deleted drops its `instance_missing_since` marker on the next sweep, rather than carrying it until the row ages out.

**`tests/operations/graph/test_infrastructure.py`** — three cases in `TestCleanupStaleGraphs`, mirroring the shared-repository exemption trio that already exists:

- `test_recently_deleted_graph_is_not_orphaned` — a deleted row whose instance is absent is neither counted nor stamped, and the published metric is 0.
- `test_deleted_graph_marker_from_before_the_exemption_is_cleared` — a marker left from a sweep that predates this rule is removed.
- `test_live_graph_is_still_orphaned_alongside_a_deleted_row` — the exemption is scoped; an active graph swept at the same time is still marked and counted.

Worth a close look: the first test is a negative control. Verified that it fails without the predicate, and that when it fails it emits the same warning line the production sweep did — so it reproduces the defect rather than asserting around it.

No behaviour change to routing, allocation, or any request path. `cleanup_stale_graphs` runs only from the daily `instance_registry_cleanup_job` and the weekly `full_instance_maintenance_job`; what changes is which rows those sweeps count and stamp.

## Breaking Changes

None. No API surface is touched — no GraphQL schema, operations envelope, or REST shape changes — so there is no client-SDK impact in either tier and no regen is required.

## Testing

- `just test-code` — clean (ruff, format, basedpyright, cf-lint).
- `uv run pytest tests/operations/graph/test_infrastructure.py` — 33 passed.
- `uv run pytest tests/operations/graph tests/dagster` — 992 passed.
- Negative control: reverted the predicate, confirmed `test_recently_deleted_graph_is_not_orphaned` fails and emits the production warning line, restored.

Full `just test-all` was not run; the unit suite above covers the changed module and its callers, and CI runs the full gate.

## Certification

https://claude.ai/code/session_01PBkpeEv3gBUzL812xqfXnS

### DIFF
--- a/robosystems/operations/graph/infrastructure.py
+++ b/robosystems/operations/graph/infrastructure.py
@@ -407,10 +407,13 @@ class InstanceMonitor:
     when the instance reappears. Reconciling from on-disk truth is the
     follow-up, not this sweep.
 
-    Shared repositories are exempt from the orphan check — their rows are
-    bookkeeping, not routing, and their master is deliberately parked to zero
-    between ingestion runs. An exempt row that still carries a stamp from
-    before this rule is cleared on the next sweep.
+    Shared repositories and rows already marked ``deleted`` are exempt from
+    the orphan check — neither is routing. A shared repository's master is
+    deliberately parked to zero between ingestion runs; a deleted graph's
+    instance is recycled long before the row ages out, so counting it means
+    every fleet replacement pages for ``STALE_GRAPH_DAYS`` and then heals
+    itself. An exempt row that still carries a stamp from before these rules
+    is cleared on the next sweep.
     """
     logger.info("Starting graph registry cleanup")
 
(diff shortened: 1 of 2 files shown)


jfrench9:
@claude please review this PR

PR #1368, opened 2026-09-09
refactor(graph): drop the HNSW reader from the vector/search route

jfrench9:
## What

Removes the HNSW branch of `POST /databases/{g}/tables/{t}/vector/search`. HNSW indexes are searched in Cypher via `CALL QUERY_VECTOR_INDEX`; that route now returns 400 for `backend="hnsw"` with the query to send instead. The Lance branch is untouched, and so is `/vector/build`, which is the live half of the module.

## Why

The HNSW reader read its rows through `result.get_as_list()` — a method `ladybug` 0.18.1's `QueryResult` does not have (it exposes `get_all`, `rows_as_dict`, `get_next`/`has_next`). The `hasattr` guard around it had no `else`, so the function fell through with an empty list:

```python
result = conn.execute(query)          # ANN runs, rows come back
results = []
if hasattr(result, "get_as_list"):    # False
    ...
return {"results": results, "total": 0}   # 200 OK, zero rows, no error
```

Every HNSW search returned `200 {"results": [], "total": 0}` with a plausible latency and nothing logged. The sibling call sites in the same module (lines 263, 386, 401) and `tables/materialize.py:153` all carry the `else list(result)` fallback — only the reader lacked it.

Nothing consumes it. `GraphClient.vector_search()` is the route's only client and never sends `backend`, so it defaults to `lance` and cannot reach the branch; its sole caller is a unit test. `/vector/build` with `backend="hnsw"` **is** live (`chunked_materialization.py:169`) and is unaffected. The module docstring already said HNSW is *"searched in Cypher via CALL QUERY_VECTOR_INDEX, not the /search route"* — the code contradicted the docs shipping beside it.

Repairing a reader with no consumers would leave two ways to search one index, one of them undocumented. Removing it leaves the way that works, and the 400 hands the caller the working query rather than an empty list.

## Tests

The existing tests could not have caught this — they set `get_as_list` on a `MagicMock`, so `hasattr` was true under test and false against the engine. That's the same shape of blind spot either way, so the replacements are route-level and go through the real request path: the 400, the pointer's contents, an explicit assertion that a 200-with-no-rows can no longer happen, and that `backend="lance"` is still served.

## Verification

Beyond the suite, on the local stack against a subgraph with a `FLOAT[384]` embedding column and 8 nodes of real fastembed vectors:

- `backend="hnsw"` → `400` naming `CALL QUERY_VECTOR_INDEX('Document', 'document_vec_index', ...)` and `POST /databases/{g}/query`
- `column` (removed from the request model) → `422`, per `extra = "forbid"`
- `/vector/build` `backend="hnsw"` → still indexes, `row_count: 8`
- `CALL QUERY_VECTOR_INDEX` through `/v1/graphs/{g}/query/cypher` → correct nearest neighbour on all three probes (goodwill → Goodwill impairment 0.316; coffee roasting → Espresso roasting 0.255; buyback → Share repurchase 0.351)

## Follow-up, not in this PR

The `hasattr(result, "get_as_list")` pattern remains at three sites in this module plus `materialize.py:153`. All are correct today because they have the fallback, but all four are guarding against a method the engine no longer has, so all four silently take the fallback path. Worth collapsing to a single reader helper separately.

https://claude.ai/code/session_01HHvoJGP23TsTZZ8njeRTWj

### DIFF
--- a/robosystems/graph_api/client/client.py
+++ b/robosystems/graph_api/client/client.py
@@ -910,6 +910,9 @@ class GraphClient(BaseGraphClient):
     Serves from master or replica instances. The index must exist already —
     build it with :meth:`vector_build`. Each result carries the selected
     columns plus a ``distance``; ``select=None`` returns all columns.
+
+    LanceDB only. An HNSW index built by :meth:`build_vector_index` is
+    searched in Cypher with ``CALL QUERY_VECTOR_INDEX(...)``, not here.
     """
     json_data: dict[str, Any] = {"embedding": embedding, "limit": limit}
     if select is not None:
--- a/robosystems/graph_api/routers/databases/vector_search.py
+++ b/robosystems/graph_api/routers/databases/vector_search.py
@@ -5,7 +5,8 @@ Two backends via the `backend` field:
 
   - **hnsw** — the live path. LadybugDB-native HNSW index on a materialized
     table. Built here (the materialize path calls build with backend='hnsw')
-    but *searched in Cypher* via CALL QUERY_VECTOR_INDEX, not the /search route.
+    but *searched in Cypher* via CALL QUERY_VECTOR_INDEX, not the /search
+    route, which rejects backend='hnsw' and says so.
 
   - **lance** — no live consumer today. LanceDB IVF-PQ built from a DuckDB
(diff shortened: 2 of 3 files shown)


jfrench9:
@claude please review this PR

PR #1369, opened 2026-09-09
refactor(graph): one reader for LadybugDB QueryResult rows

jfrench9:
Follow-up to #1368, which removed a reader that silently returned nothing. This removes the pattern that produced it.

## What

Four call sites read rows this way:

```python
rows = result.get_as_list() if hasattr(result, "get_as_list") else list(result)
```

`QueryResult` has no `get_as_list` — ladybug 0.18.1 exposes `get_all`, `rows_as_dict`, `get_as_arrow` and a `has_next`/`get_next` cursor — so the guard is always false and all four have been taking the fallback since the rename. They work today. Replaced with `result_rows()` in `core/ladybug/results.py`, built on the documented cursor, matching the row-to-dict readers already in `service.py` and `engine.py`.

- `vector_search.py:263` (build, COUNT), `:386` (info, TABLE_INFO), `:401` (info, COUNT)
- `tables/materialize.py:153` (`_get_target_columns`, TABLE_INFO)

No behavior change: the fallback every site was silently taking is now the only path.

## Why, given they all work

The shape is the defect, not the four instances. It reads as though either branch might fire, so a fifth copy written without the `else` looks like a style choice rather than a bug. That fifth copy is exactly what #1368 deleted — the HNSW reader that answered `200 {"results": []}` for every query, with nothing logged, because a missing method behind `hasattr` is not an error. Two files, four sites, one demonstrated failure is enough evidence to collapse it.

The honest limit: a helper only prevents recurrence if the next person reaches for it. It's exported from `core.ladybug` and documented as the one way in, but that's a convention, not a guarantee.

## Deliberately left alone

- `_copy_result_rows` in `materialize.py` guards on `get_as_arrow`, which `QueryResult` **does** have. Live guard, correct as written.
- `service.py` and `engine.py` build column-mapped dicts with row caps for API responses. Different job, correct, and riskier to touch than to leave.

## The tests were half the bug

The old tests set `get_as_list` on a `MagicMock`, so `hasattr` was true under test and false against the engine — the mock manufactured the very method whose absence was the defect.

Swapping in a fake with the engine's real surface didn't just flip those tests to passing; **it made them hang.** `MagicMock.has_next()` returns a truthy `Mock`, so the cursor never terminates. That's the same lie pointing the other way, and it's why `FakeQueryResult` now lives in `tests/graph_api/conftest.py` for anything reading rows off a result.

`test_results.py` covers order, cursor exhaustion, unnormalized rows (`TABLE_INFO` callers distinguish tuple rows from dict rows), and a trap fake that raises if anything reaches for `get_as_list` again.

1393 graph_api tests pass; `just test-code` clean.

## Verification

On the local stack, against the subgraph from #1368:

- `GET /vector` → `row_count: 8` — exercises TABLE_INFO **and** COUNT through the new reader
- `POST /vector/build` `backend="hnsw"` → `row_count: 8`
- table with no embedding column → `200` with null, not a 500

The `TABLE_INFO` read is the same query and the same reader `_get_target_columns` uses during materialization, so that row shape is covered live rather than only under mocks.

https://claude.ai/code/session_01HHvoJGP23TsTZZ8njeRTWj

### DIFF
--- a/robosystems/graph_api/core/ladybug/__init__.py
+++ b/robosystems/graph_api/core/ladybug/__init__.py
@@ -4,6 +4,7 @@
 - ``LadybugConnectionPool`` — shared, thread-safe connections per database
 - ``LadybugDatabaseManager`` — database lifecycle and the blue-green swap
 - ``LadybugService`` — the query, health and metrics surface the routers use
+- ``result_rows`` — the one way to read rows out of a ``QueryResult``
 """
 
 from .config import get_database_memory_config
@@ -14,6 +15,7 @@ from .pool import (
   get_connection_pool,
   initialize_connection_pool,
 )
+from .results import result_rows
 from .service import (
   LadybugService,
   get_ladybug_service,
@@ -38,5 +40,6 @@ __all__ = [
   "get_ladybug_service",
   "init_ladybug_service",
   "initialize_connection_pool",
+  "result_rows",
   "validate_cypher_query",
 ]
--- a/robosystems/graph_api/core/ladybug/results.py
+++ b/robosystems/graph_api/core/ladybug/results.py
@@ -1,0 +1,34 @@
+"""Reading rows out of a LadybugDB ``QueryResult``.
+
+One helper, because the alternative has already cost us a bug. Callers used
+to write::
+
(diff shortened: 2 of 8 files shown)


jfrench9:
@claude please review this PR

PR #1373, opened 2026-09-09
fix(serialization): emit the holon from xbrlkit through the bundle bridge

jfrench9:
## Summary

The platform's `rdf/holon.py` carried its own copy of the scene/boundary/projection partition, forked from the kernel xbrlkit now owns. It seeded the scene from facts alone, so every element reached only by an arc — an empty statement line, an abstract header — left the holon without its `rs:Element` and rendered without a label: **56 of the SaaS demo's 98 elements**. xbrlkit's kernel had since fixed this (arc endpoints carried into `#projection`, dimensions, the `rs:Report` node); the fork never inherited it.

- `serialize_to_holon_jsonld` is now `to_holon(bundle_to_xbrl_model(bundle))` — the waist the Tavi flavor already reads through. The platform partition is deleted.
- The DataBook converter (`examples/_common/databook.py`) reads the flat JSON-LD back through xbrlkit's reader and calls the same writer, and stamps the report id as the accession so the derived holon reads back like the API's.
- The flat JSON-LD encoder — the synchronous publish path — is untouched.
- The tracked demo sample `saas-startup-demo.holon.jsonld` is regenerated through the new path.

## Verification

| | before | after |
|---|---|---|
| `rs:Element` declared | 42 | 98 |
| arc endpoints referenced but undeclared | 56 | 0 |
| xbrlkit load — concepts / labelled | 42 / 38 | 98 / 93 |
| `rs:Report` node in `#scene` | no | yes |
| root IRI | `https://robosystems.ai/report/{id}` | unchanged |

Facts are identical before and after (86, 84 numeric, no value differences). New tests pin the regression — every arc endpoint declared, xbrlkit reads back every bundle concept, the named graphs hang off the report IRI. `tests/operations/serialization` plus the download and publication-artifact tests: 230 passed.

This is the holon half of `serialization-waist` phase 3, pulled early by the defect; the JSON-LD half stays parked.

### DIFF
--- a/examples/_common/databook.py
+++ b/examples/_common/databook.py
@@ -37,10 +37,8 @@ from pathlib import Path
 import rdflib
 from rdflib import RDF, Graph, URIRef
 from rdflib.namespace import Namespace
-
-from robosystems.operations.serialization.rdf.holon import (
-  serialize_holon_jsonld_from_graph,
-)
+from xbrlkit.deserialize.holon import from_holon_json
+from xbrlkit.serialize import to_holon
 
 REPO_ROOT = Path(__file__).resolve().parents[2]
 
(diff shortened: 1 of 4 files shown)


jfrench9:
@claude please review this PR

PR #1375, opened 2026-09-10
fix(serialization): carry structure and fact-set ids into the holon; refresh the demo samples

jfrench9:
## Summary

Follow-up to #1373, which moved the holon onto xbrlkit's writer and with it lost the identity the flat JSON-LD and the holon of one report had shared: the flat bundle names structures, associations, Information Blocks and fact sets by the platform's ids; xbrlkit slugged the role URI; nothing on the bridge carried the ids across. Periods, units and facts were fine.

- **xbrlkit 0.11.0** (`chore(deps)`), carrying [xbrlkit #48](https://github.com/RoboFinSystems/xbrlkit/pull/48): a producer's structure and fact-set ids are written verbatim, and a pinned fact belongs to its own structure.
- The bridge passes each network's `structure_id` and fact-set id, and each fact's structure pin.
- `test_holon_alignment` pins the invariant: every shared node class identical between the two projections, every fact on the same structure and fact set, IBs named by their structure. The only arcs the holon lacks are the definition arcs the bridge does not carry (documented in `model.py`).
- All five demo sample sets are **re-downloaded from one run each** on a clean rebuild from the 0.11.0 lock — the four artifacts and the three reports per demo, one report id each. They replace #1373's single-file regen and the four older samples.

## Second-round checks, on the published package

| demo | IRI identity flat↔holon | per-fact structure / fact set | arc endpoints declared | four representations agree | round-trip fixed point | SHACL · Arelle |
|---|---|---|---|---|---|---|
| saas-startup | all 9 classes | 86/86 · 86/86 | 98/98 | 0 disagreements | yes | clean |
| coffee-roaster | all 9 classes | 90/90 · 90/90 | 101/101 | 0 | yes | clean |
| roboledger | all 9 classes | 41/41 · 41/41 | 93/93 | 0 | yes | clean |
| seattle-method-case-1 | all 9 classes | 48/48 · 48/48 | 93/93 | 0 | yes | clean |
| world-online | all 9 classes | 56/56 · 56/56 | 93/93 | 0 | yes | clean |

The old samples fail the same checks. The same demos were first run against a local wheel of #48 overlaid on the image, then again on the published 0.11.0; keyed by element, period dates, structure and fact set, zero facts differ between the two runs — the same report modulo per-run ids. Holon and Tavi carry the same report id per demo; the XBRL 2.1 zip carries fewer facts by the documented disclosure strip and dedupe, with no value differences on what it shares.

One residual, pre-existing: the two `@context`s compact an extension concept differently (`concept/cadence:X` vs `cadence:X`), so raw Element IRIs match only for demos without extension concepts.


jfrench9:
@claude please review this PR

PR #1376, opened 2026-09-10
feat(mcp): trim search hits to decision fields and page section reads

jfrench9:
## Summary

A tool result is context the model pays for on every later turn. Measured across both published Filing Ladder runs (2,280 records, 9,900 tool calls), `search-documents` was a median 10K characters per call, 44% of it the per-section `xbrl_elements` list that no transcript used to choose a hit, and `get-document-section` returned whole ~25K-character parts to read one paragraph. This PR shrinks what those two MCP tools say while answering, without changing what they can answer. The REST search models and the graph-side retrieval (ranking, chunking, what is indexed) are untouched.

Design and measurement: `local/RoboSystems/specs/ai-operators/mcp-response-economy.md` (changes 1–4 of the six there; 5 and 6 are re-scoped in the spec rather than built).

## Changes

**`search-documents` — `middleware/mcp/tools/search_tools.py`**
- New pure function `compact_search_response`: drops `xbrl_elements` from hits (it stays on `get-document-section` for the hit the model picks, and the `element` filter still runs the other direction), drops null fields, and hoists the filing fields (`entity_ticker`, `entity_name`, `form_type`, `filing_date`, `fiscal_year`) to the result root when every hit shares the same value. Hits that span filings keep them per hit.
- New `snippet_chars` argument (default 400, min 80, max 1500), clamped in the tool.
- Description rewritten to match: a hit no longer promises `xbrl_elements`; a snippet is described as an excerpt to be read from the section before a figure is quoted from it (the description half of the *snippets get quoted as answers* row in the ai-operators README); the hoist and the new parameter are stated.

**`get-document-section` — same file**
- New `offset` + `length` arguments (default 4,000, max 8,000 — xbrlkit `read_text`'s caps, so the hosted and local document tools page the same way) via `window_section`. The result carries `content_length` for the whole part and `next_offset` while more follows; an offset past the end is an error. `xbrl_elements` and the part chain (`next_document_id`) are unchanged.

**Snippet budget plumbing** — `models/api/search.py`, `operations/search/client.py`, `operations/search/service.py`
- `SearchRequest.snippet_chars: int | None` (additive; `None` keeps today's REST behaviour).
- `OpenSearchClient._highlight_config(snippet_chars)`: fragments stay at the standard 200 characters and the budget sets how many a hit may carry, so a smaller budget shows fewer matches, not shorter ones. Threaded through both `search` and `search_hybrid`.

**Prompts and docs**
- `operations/operators/implementations/analyst.py`: the analyst's two `search-documents` guidance lines now say the section tool returns a window and to page on with `offset` rather than answering from a partial read.
- `middleware/mcp/README.md` Layer 3: describes the new hit and section shapes.

**Measured on the local MMM FY2024 10-K** (same query, before → after):

| Call | Before | After |
|---|---|---|
| 10-hit iXBRL disclosure search | 17,092 chars | 8,407 |
| 4-hit MD&A search | 4,590 | 3,240 |
| One MD&A part read (default window) | 25,227 | 4,813 |

The re-run that proves the spec's cost claim is the Filing Ladder `7a-tagged` control against a deploy carrying this change — accuracy must not move, cost should fall. Not run here; the ladder reads whatever the tools return, so it measures the change rather than breaking on it.

## Breaking Changes

None. The MCP response shape is a tool-layer change; the REST `/v1/graphs/{graph_id}/search` models are unchanged apart from the additive `snippet_chars` request field (generated tier, free — an SDK regen opportunity, not a coordinated release). Nothing outside the generated SDK models reads `xbrl_elements` from a search hit (checked the three apps, `robosystems-mcp-client`, and the Filing Ladder runner, which reads only document ids).

## Testing

- Targeted suites: `tests/middleware/mcp`, `tests/operations/search`, `tests/operations/operators`, `tests/routers/graphs/mcp`, `tests/routers/graphs/test_search.py`, `tests/routers/graphs/test_documents.py`, `tests/models` — 2,324 passed. New tests cover the compaction rule (drop, hoist, span-filings, missing-on-one-hit, empty, part fields, payload halves on the measured shape), the section window (whole-when-short, paging, capped length, offset past end, empty content), the snippet clamp, and the highlight config for both search modes.
- The tool-description contract test passes on the rewritten descriptions.

## Certification

### DIFF
--- a/robosystems/middleware/mcp/README.md
+++ b/robosystems/middleware/mcp/README.md
@@ -198,10 +198,18 @@ Gated by `SEMANTIC_SEARCH_ENABLED`.
 | `create-document` / `update-document` | no |
 
 Document management additionally skips shared repositories — SEC searches
-OpenSearch directly and has no Postgres document rows. `search-documents`
-results on iXBRL disclosures carry `xbrl_elements`, the XBRL fact tags in that
-section, which cross-reference into the graph via `resolve-element` or
-`read-graph-cypher`.
+OpenSearch directly and has no Postgres document rows.
+
+A tool result is context the model pays for on every later turn, so these two
+tools return less than the REST models they wrap (`search_tools.py`). A
+`search-documents` hit carries what a model needs to choose it — id, label,
+score, snippet (`snippet_chars`, default 400) — with null fields dropped and
+the filing fields hoisted to the result root when every hit shares them; the
+per-section `xbrl_elements` list is left to `get-document-section`, which
+carries it for the hit the model picked and cross-references into the graph
+via `resolve-element` or `read-graph-cypher`. `get-document-section` returns
+a window (`offset` + `length`, default 4,000, max 8,000) and a `next_offset`
+while more of the part follows, the same paging as xbrlkit's `read_text`.
 
 ### Layer 4 — graph lifecycle, subgraph writes, memory
 
--- a/robosystems/middleware/mcp/tools/search_tools.py
+++ b/robosystems/middleware/mcp/tools/search_tools.py
@@ -6,12 +6,98 @@ Two-tool pattern:
 
 All searches use hybrid mode combining keyword matching (BM25) with vector
 similarity (KNN) via a normalization pipeline for balanced scoring.
+
+A tool result is context the model pays for on every later turn, so a hit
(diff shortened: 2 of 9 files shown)


jfrench9:
@claude please review this PR

PR #1377, opened 2026-09-10
feat(mcp): cap financial-statement-analysis at the columns a filing presents

jfrench9:
## Summary

The second half of the MCP response economy (`local/RoboSystems/specs/ai-operators/mcp-response-economy.md` §3, change 5), following #1376. `financial-statement-analysis` was a median 24K characters per call in the Filing Ladder runs because it returned every period its hypercube holds: on the 3M FY2024 10-K income statement, 8 of 11 period keys were the quarterly figures from the notes, and the balance sheet carried 2021 and 2022 instants from the equity roll-forward. Each fact row also repeated the qname as `name` plus three null fields. The tool now answers with the columns a filing presents, and says what it left out.

## Changes

**`financial-statement-analysis` — `middleware/mcp/tools/financial_statement_tools.py`**
- `default_period_type`: when the caller gives no `period_type`, an annual form (10-K / 20-F / 40-F, per the resolver's `ANNUAL_FORMS`) defaults the filter to `annual`. The balance sheet keeps the query's instant default; a 10-Q is left unfiltered because its quarter and year-to-date columns share end dates, so the cap below bounds it without a filter. An explicit `period_type` still wins.
- `cap_periods`: a new `periods` argument caps the distinct period end dates kept, newest first — default 2 for the balance sheet, 3 for the flow statements, max 100. The result carries `periods` (the keys the facts span, newest first) and, when older end dates were cut, `periods_omitted` with a tip to raise the cap.
- `compact_fact`: rows drop `name` (always the local part of `qname` — verified on the live output) and null fields.
- `limit` applies within the kept periods (dedup → cap → limit).
- Description and `inputSchema` state all of the above.

**Docs** — `middleware/mcp/README.md`: the tool table cell and the paragraph beneath it describe the cadence default, the cap, and the row shape.

**Measured on the local MMM FY2024 10-K** (before = every period, old row shape):

| Statement | Before | After |
|---|---|---|
| Income statement | 50,043 chars / 174 facts / 9 end dates | 18,531 / 78 / 3 |
| Balance sheet | 25,617 / 99 / 4 | 15,837 / 94 / 2 (2 omitted, said so) |
| Cash flow | 30,938 / 106 / 4 | 23,952 / 102 / 3 |

Not changed: the REST view op at `routers/extensions/roboledger/views.py` still returns every period and the full row model. This is a tool-layer change for the reader with the budget, as in #1376.

## Breaking Changes

None. MCP response body and tool definition only; no REST or SDK surface moves. Under the ChatGPT plugin's rule the new `periods` parameter is a definition change that rides the next plugin version's review, alongside #1376's parameters.

## Testing

- `tests/middleware/mcp/tools/test_financial_statement_tools.py` plus the description contract, manager helpers, MCP execute, and roboledger views suites — 335 passed. New tests cover the cadence default per form and statement, the cap (newest end dates, keys sharing an end date counting once, empty input), row compaction, and the tool end to end (annual default sent to the query, explicit `period_type` winning, balance sheet not forced annual, the 3-period and 2-period defaults with `periods_omitted`, the `periods` argument and its bound, `limit` within the kept periods).
- Live check against the local `sec` graph produced the table above.

## Certification

### DIFF
--- a/robosystems/middleware/mcp/README.md
+++ b/robosystems/middleware/mcp/README.md
@@ -137,7 +137,11 @@ period filter to `annual` (a 10-K's statement hypercube also carries the
 quarterly figures from its notes), `periods` caps the distinct end dates kept
 (two for the balance sheet, three for the flow statements, newest first) and
 the result names the keys it kept and how many it cut, and fact rows carry no
-`name` (the local part of `qname`) and no null fields.
+null fields and a `name` only when it is a label rather than the local part of
+`qname` (SEC rows repeat the qname there; a tenant's rs-gaap rows carry a
+readable label). The graph fetch always takes the query's full row budget so
+that `limit`, applied after the cap, can never hide a period from it; a fetch
+that hits the ceiling is flagged `rows_truncated`.
 
 ### Layer 2b — roboledger OLTP
 
--- a/robosystems/middleware/mcp/README.md
+++ b/robosystems/middleware/mcp/README.md
@@ -124,14 +124,20 @@ Require `roboledger` in `schema_extensions`. These read LadybugDB (OLAP).
 |------|-------------|------------|
 | `get-example-queries` | Query patterns tailored to this graph's schema | — |
 | `resolve-element` | Map a concept ("revenue") to XBRL element qnames | manifest `has_semantic_enrichment=True` |
-| `financial-statement-analysis` | Graph-backed statement read with auto-resolve and dedup | — |
+| `financial-statement-analysis` | Graph-backed statement read with auto-resolve, dedup, and a period cap at the columns a filing presents | — |
 | `live-financial-statement` | Statement from the tenant's live OLTP ledger via CoA→GAAP mapping | tenant graphs only |
 | `build-fact-grid` | Cross-company comparison over canonical concepts | `FACT_GRID_ENABLED` |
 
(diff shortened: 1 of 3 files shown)


jfrench9:
@claude please review this PR

jfrench9:
Decision: fix now, in c682d4cd. All three points addressed.

1. **`limit` starving the cap** — real. The query now always takes its 1,000-row ceiling (`QUERY_ROW_CEILING`) regardless of the caller's `limit`, and `limit` applies after dedup and the cap. A fetch that hits the ceiling is flagged `rows_truncated` with a tip, since the oldest kept period may then be incomplete. Tests: the query receives the ceiling when `limit=5`; a 1,000-row fetch is flagged; a smaller one is not.
2. **`name` == local part of `qname`** — the assumption was wrong, not just undocumented. It held on every SEC row but a tenant graph's rs-gaap elements carry a readable label there (`rs-gaap:NonoperatingIncomeExpense` → "Nonoperating Income (Expense)"). `compact_fact` now keeps `name` when it differs from the local part and drops it only when redundant; docstring says which graph has which shape. Test added for the tenant row.
3. **`periods: 0`** — now floored to 1 like any other out-of-range value; only an absent argument takes the default. Test added.

`just test-code` clean, 249 tests in the tool's suites green.

@claude please re-review on the new commit.

PR #1378, opened 2026-09-11
feat(ledger): native accounting cutover: provider guard, sever, chart templates

jfrench9:
## Summary

Native and synced ledgers never mix, and a bank feed needs a chart of accounts. This PR is the prerequisite for the in-core Mercury bank-feed connector: a provider-compatibility guard, a `sever` disposition that cuts a QuickBooks tenant over to native books while keeping the chart QuickBooks created, and a one-time `initialize-chart-of-accounts` operation that gives a fresh company a chart from a shipped template. No database migration. Design record: `local/RoboSystems/specs/ledger/native-accounting-cutover.md`.

## Changes

**Provider guard** (`operations/connection_service.py`, `routers/graphs/connections/management.py`, `routers/graphs/connections/oauth.py`)
- `assert_provider_compatible(graph_id, provider, session)` raises `ProviderConflictError` with a stable code: a bank feed (`mercury`) while a QuickBooks connection is live → `QUICKBOOKS_ACTIVE`; a bank feed on a graph with no active `chart_of_accounts` taxonomy → `CHART_REQUIRED`; QuickBooks over natively-kept books (posted line items on elements it did not create, or a live bank feed) → `NATIVE_BOOKS_PRESENT`. Other providers pass without touching the graph. A missing extensions schema reads as no books.
- Wired into `create_connection` (after the provider-enabled check, before the duplicate short-circuit) and OAuth init (the callback can revive a soft-deleted row, so the guard runs before any authorize URL). Both surface as HTTP 409 with the code in the error body.
- Graph-side predicates live in `operations/roboledger/reads/books.py`.

**Sever — the cutover** (`models/core/connection/connection.py`, `operations/connection_service.py`, `operations/roboledger/commands/connections.py`, `security/audit_logger.py`)
- `ConnectionStatus.SEVERED` (string column, no migration). `deleteConnection` gains `disposition=disconnect|sever` (default `disconnect`, today's behaviour). `sever` is QuickBooks-only (400 otherwise) and runs first: `sever_synced_chart` stamps the connection's elements `source='native'` and nulls `external_source` / `connection_id` / `external_id`, which takes them out of the loader's upsert key so a later sync can neither overwrite nor re-adopt them; qnames are kept. Then `write_policy` drops to `native`, the row is marked `severed`, provider cleanup and the soft-delete follow. Recorded as a `connection_severed` security-audit event.
- `find_soft_deleted_for_realm` excludes severed rows, so a later re-OAuth mints a fresh connection (which the guard then refuses over native books) instead of reviving the severed one.
- Plain disconnect now also resets `write_policy` to `native`; `Connection.restore` re-applies the provider default on revival. Reviewers: this is a behaviour change on the existing disconnect path, deliberate — `native` describes a graph with no active authoritative external GL.

**Chart of accounts from a template** (`operations/taxonomy_block/chart_templates/`, `operations/roboledger/commands/chart_of_accounts.py`, `models/api/extensions/chart_of_accounts.py`, `routers/extensions/roboledger/operations.py`, `graphql/`)
- Three shipped templates — `saas` (20 accounts), `services` (27), `product` (27) — each an `ACCOUNTS` list plus form-aware rs-gaap mappings (`_forms.py` swaps the equity rows by entity legal form). `saas` and `services` are the SaaS and RoboLedger demo charts promoted into core, and those demos now import them back so there is one copy; `product` is the coffee-roaster chart with the roaster-specific names generalized.
- `initialize_chart_of_accounts` refuses (409) when an active chart exists, builds the Taxonomy Block envelope (elements `coa:<code>`, a `coa_mapping` structure) through the existing declarative CoA handler, then resolves the template's rs-gaap targets against the library and creates the mapping associations in the same session. Library misses are reported in `unresolved`, never fatal. `entity_type` defaults to the graph's primary entity, then corporation.
- Registered through the operation registrar as `initialize-chart-of-accounts` (REST op + MCP tool from one spec; `mark_stale_reason="chart_of_accounts_initialized"`). The template catalogue is the new `chartTemplates` GraphQL field on the ledger query.
- A comment on the loader's `_ensure_mapping_structure` records why it may still assume an empty-or-own chart.

**Tests**
- New: `tests/operations/taxonomy_block/test_chart_templates.py` (registry consistency; every mapping target verified against the rs-gaap framework source; demos-are-the-templates, including the structural pin between `product` and the coffee demo), `tests/operations/roboledger/commands/test_chart_of_accounts.py`, `tests/operations/roboledger/commands/test_connections.py`.
- Extended: guard matrix + `sever_connection` + write-policy reset in `tests/operations/test_connection_service.py`; 409 and sever-disposition cases in the management and OAuth router suites (their autouse fixtures now patch the guard, since it opens an extensions session); severed-row exclusion and `restore` write-policy in the Connection model tests.

## Breaking Changes

None. Additive surface only — a new operation, a new optional query parameter on `deleteConnection`, a new GraphQL field, a new `ConnectionStatus` value and three new 409 codes — so this rides a client **minor** on the next SDK regen (generated tier; nothing the integration template imports changes). The Python client's GraphQL drift gate will pick up `chartTemplates` on its next `just refresh-schema`.

## Testing

- `just test` — 14,314 passed, 42 skipped.
- `just test-code` — ruff, format, basedpyright, cf-lint all passed.
- Not exercised end-to-end against a live graph in this session; the sever and initialize paths are covered by unit tests with mocked sessions, and the template tests read the rs-gaap framework source directly.

## Certification

https://claude.ai/code/session_01JNVVqyL2inzXtxnWZp68Rb

### DIFF
--- a/examples/roboledger_demo/data.py
+++ b/examples/roboledger_demo/data.py
@@ -16,6 +16,10 @@ from __future__ import annotations
 
 from datetime import date
 
+from robosystems.operations.taxonomy_block.chart_templates.services import (
+  ACCOUNTS as ACCOUNTS,
+)
+
 # ---------------------------------------------------------------------------
 # Evergreen reference dates
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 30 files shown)


jfrench9:
@claude please review this PR

PR #1380, opened 2026-09-12
feat(ledger): retire chart accounts — is_active patch and a posting guard

jfrench9:
## Summary

A chart of accounts changes constantly, but an account that carries facts or line items is never removed. This adds the lane natively kept books were missing: retire an account with `is_active=false` on an `update-taxonomy-block` element patch, and refuse new postings to a retired account. Spec: `local/RoboSystems/specs/taxonomy/chart-templates-as-data.md` §5.3.

## Changes

**Patch field** (`models/api/taxonomy_block.py`, `operations/taxonomy_block/update_apply.py`)
- `ElementUpdatePatch.is_active: bool | None` (additive, optional). Applied in place beside the other scalar fields; omitted leaves the column alone, so a rename never clobbers the flag the QuickBooks loader lands from QuickBooks' Active. Library immutability and the unknown-qname check already cover element patches generically, so the validator is unchanged.

**Posting guard** (`operations/roboledger/commands/_guards.py`, `commands/journal_entries.py`)
- `assert_accounts_postable(session, element_ids, *, source=None)` and `InactiveAccountError(ValueError)`, which names each retired account by code, name and id and says how to reactivate. `ValueError` already maps to 422 on every write operation, so no registrar change.
- Called in `create_journal_entry` after line normalization with `source=body.source`, and in `update_journal_entry` on replacement lines as an authored edit. Both event-handler paths (`journal_entry_recorded`, nested and flat) funnel into `create_journal_entry`, so manual entries and approved events share the one guard.
- Exemption: a `source` in `SEVERABLE_SOURCES` (QuickBooks) passes through, **and only for the replay shape** — `create_journal_entry` forwards the source to the guard only when `status='posted'`, which is what the loader stamps on every replayed entry; a draft that merely names a synced source is authored and is checked. QuickBooks retires accounts after they carry activity and a full rebuild replays that history verbatim; blocking it would break re-syncs. Bank feeds and native postings are checked. **Accepted limitation** (from the review): a tenant with a live QuickBooks connection can still hand-author a `posted` event under `source='quickbooks'` and post to its own retired account. That is its own graph and its own books; the guard is a safety on authored work, not a boundary, and the exemption is documented on the guard.

**Reads**: no change needed. `get_account_tree` already hides inactive accounts unless `include_inactive=True`, and `list_accounts` filters on `is_active`; both response models already carry the flag.

**Tool descriptions** (`routers/extensions/roboledger/operations.py`)
- `update-taxonomy-block` now says how to retire an account and that removal and whole-chart delete need no activity. `update-journal-entry` notes that a line on a retired account is refused. Both reach the MCP tools through the registrar.

**Tests**
- New `tests/operations/roboledger/commands/test_guards_accounts.py`: passes on no retired rows, names every retired account, is a `ValueError`, the QuickBooks exemption is case-insensitive, other sources are checked, blank ids skip the query, duplicates are queried once.
- `test_journal_entries.py`: a module-wide autouse stub for the guard (every test there runs on a mocked session), plus three tests pinning what create and update pass to it and that a refusal writes nothing.
- `test_update.py`: `TestApplyIsActive` covers retire, reactivate, and omitted-leaves-alone.

**Not in this PR**: schedule-generated entries (`schedules/service.py` writes `LineItem` rows directly to elements chosen at schedule creation); an app editor for retiring accounts.

## Breaking Changes

None. Additive optional request field on `update-taxonomy-block` (generated tier, client minor on the next regen; the integration template's emit path is untouched). New behaviour: an authored posting to an account that was already `is_active=false` now returns 422 instead of succeeding, which is the intended semantic of the flag.

## Testing

- `just test-code`: ruff, format, basedpyright, cf-lint all passed.
- `uv run pytest` over `tests/operations/roboledger/commands`, `tests/operations/taxonomy_block`, `tests/operations/event_block`, `tests/routers/extensions/roboledger`, `tests/middleware/mcp` with `-m "not slow"`: 1,348 passed against the local stack, including the DB-backed state-transition suite that posts real entries through the guard.
- Not run: the full `just test` in this session. Not exercised end-to-end against a live graph.

## Certification

### DIFF
--- a/robosystems/models/api/taxonomy_block.py
+++ b/robosystems/models/api/taxonomy_block.py
@@ -423,6 +423,15 @@ class ElementUpdatePatch(BaseModel):
   code: str | None = None
   parent_ref: str | None = None
   metadata: dict[str, Any] | None = None
+  is_active: bool | None = Field(
+    None,
+    description=(
+      "Retire (`false`) or reactivate (`true`) a chart account. Retiring keeps "
+      "its history and hides it from account pickers and the chart tree, and "
+      "new line items on it are refused — the way to take an account with "
+      "activity out of use, since removal needs no facts and no line items."
+    ),
+  )
 
 
 class StructureUpdatePatch(BaseModel):
--- a/robosystems/operations/roboledger/commands/_guards.py
+++ b/robosystems/operations/roboledger/commands/_guards.py
@@ -8,13 +8,16 @@ opens) — they require DB access.
 
 from __future__ import annotations
 
+from collections.abc import Iterable
 from datetime import date
 from typing import Any
 
-from sqlalchemy import text
+from sqlalchemy import select, text
 from sqlalchemy.orm import Session
 
+from robosystems.models.extensions import Element
 from robosystems.operations.locking import acquire_shared_period_fence
+from robosystems.operations.roboledger.commands.connections import SEVERABLE_SOURCES
 
 _LIBRARY_SEEDER = "library-seeder"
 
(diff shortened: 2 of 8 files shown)


jfrench9:
@claude please review this PR

PR #1381, opened 2026-09-12
feat(adapters): Mercury bank feed — the provider, the adapter, one Dagster asset

jfrench9:
## Summary

The Mercury bank feed — the bank as a first-class source, and the first bank-feed provider. A bank feed is native accounting: it captures every posted Mercury transaction into the ledger inbox with a Tier-0 account suggestion attached, and authors no chart, no elements and no GL rows. The provider guard that shipped in v1.11.19 already refuses it beside a live QuickBooks connection or on a graph with no chart of accounts; this PR is the provider behind that guard.

Off by default (`CONNECTION_MERCURY_ENABLED=false`): it needs Mercury's partner OAuth client, and the production client sits behind Mercury's compliance review.

## Changes

- **Provider** (`operations/providers/mercury_provider.py`): `MercuryOAuthProvider` for the core `OAuthHandler` (sandbox/production hosts, `read offline_access`, `scope` re-sent on every refresh, revocation); `create` / `sync` / `cleanup`. Two credential modes on one provider: `oauth` (hosted default) and `api_key` — a personal read-only token, gated by `MERCURY_API_KEY_CONNECTIONS_ENABLED`, which hosted production never turns on.
- **Adapter** (`adapters/mercury/`): `MercuryClient` over a `TokenSource` — `ConnectionTokenSource` refreshes ahead of expiry and on 401, persists the rotated refresh token at once, and flips `needs_reauth` only on `invalid_grant` / `invalid_scope` / `invalid_client`; retries 429/5xx on `Retry-After`. `accounts.py` links or creates one chart account per bank account through the TaxonomyBlock envelope (a link lives in `metadata.bank_feed`, never as an element source or a `mercury:` qname). `transform.py` turns transactions into captured-event payloads (internal transfers paired into one event; `tier0.py` hints from `glAllocations` → custom category → `mercuryCategory`). `load.py` writes agents scoped to the connection and events through `create_event_block_in_session`; a re-run reports existing rows and refreshes hints on still-captured ones.
- **One Dagster asset** `mercury_feed` in job `mercury_sync` (`dagster/max_retries: 3`): every failure recorded on the connection via `record_sync_result`, the sync lock always released, the fiscal calendar bootstrapped at the first posted month on a fresh company.
- **Routers**: `mercury_config` on create; OAuth init for both providers; a Mercury callback path that parks the connect-time sync config into the token bundle, records the consent to the security audit log, and dispatches the first sync; the options catalog entry advertises `api_key` only where the deployment allows it.
[...]

PR #226, opened 2026-01-21
Refactor user graph assets and improve Dagster UI observability

jfrench9:
## Summary
This PR refactors user graph asset definitions and enhances observability in the Dagster UI by introducing a new materialized source asset. The changes also include datetime handling improvements and removal of deprecated code.

## Key Accomplishments
- **Enhanced Observability**: Introduced `user_graph_materialized_source` asset to provide better visibility and monitoring capabilities in the Dagster UI
- **Asset Architecture Improvement**: Refactored user graph asset definitions for better maintainability and performance
- **Code Cleanup**: Removed deprecated `staged_files` asset definition, reducing technical debt
- **Datetime Handling**: Updated datetime processing logic for improved reliability
- **Demo Updates**: Enhanced element mapping demonstration with updated report generation functionality

## Breaking Changes
- **Removed Asset**: The `staged_files` asset has been completely removed. Any downstream dependencies on this asset will need to be updated to use alternative data sources
- **Asset Structure**: Changes to user graph asset definitions may affect existing asset dependency chains

## Testing Notes
- Verify that all existing user graph workflows continue to function correctly
- Confirm that the new materialized source asset appears properly in the Dagster UI
- Test datetime handling in graph processing pipelines
- Validate that removal of staged_files asset doesn't break any dependent processes
- Check element mapping demo functionality with the updated report generation

## Infrastructure Considerations
- Monitor asset materialization performance after deployment
- Ensure adequate resources are allocated for the new materialized source asset
- Review asset refresh schedules to accommodate the new observability features
- Consider any impacts on existing backup and recovery processes due to asset structure changes

---
🤖 Generated with [Claude Code](https://claude.ai/code)

**Branch Info:**
- Source: `bugfix/user-graph-material-asset`
- Target: `main`
- Type: bugfix

Co-Authored-By: Claude <noreply@anthropic.com>

PR #1382, opened 2026-09-12
feat(ledger): the bank-feed classify loop — classify, then commit

jfrench9:
## Summary

The second half of the Mercury bank feed (#1381): the loop that turns a captured bank line into a posted entry. The feed captures every posted transaction into the inbox with an account suggestion; until now nothing posted it, and approving one in the inbox moved it to `committed` with no GL rows, because no handler existed for the bank event types and the dispatch falls through silently. This PR makes classify-then-commit real, over MCP first (the spec's §6 doctrine) and for the inbox UI's approve button.

## Changes

- **Transitions** (`operations/event_block/commands.py`, `models/api/event_block.py`): `captured → classified` is open to callers. It records the account choice — `metadata.classified_element_id`, a `classified_allocations` split, or `accept_suggestion: true` — without posting; the later commit fires the handler. The request model's `transition_to` literal and the `update-event-block` operation description follow.
- **Bank-feed handlers** (`python_handlers/bank_feed.py`, registered for `bank_transaction`, `bank_fee`, `external_transfer`, `internal_transfer`): build the balanced entry from the classification (money in: DR bank / CR contra; money out: the reverse; a split posts one contra line per allocation and must add up; an internal transfer posts from its two bank legs) and delegate the GL write to `journal_entry_recorded` as a local-lane draft (`publish_to_source=False`, never written back anywhere) that close posts. The closed-period fence, balance check and the postability guard are the journal handler's. An unclassified commit falls back to the tenant's DSL rules — the deterministic floor for a counterparty that never varies — and is refused (422, `BankEventNotClassifiedError`) when none matches or several tie. `preview-event-block` shows the planned entry and names the missing choice.
- **MCP instructions** (`middleware/mcp/tools/instructions.py`): an INBOX / BANK FEED block, present only when the graph exposes the write tools: find the captured lines, recall the counterparty, classify, commit, remember the decision; the `recall`/`remember` and preview lines track their tools.
- Reviewers: `bank_feed.py::plan_lines` (the posting shape) and `_apply_dsl_floor` (what an unclassified commit does) carry the doctrine.

## Breaking Changes

None. Generated-tier addition: `classified` joins `UpdateEventBlockRequest.transition_to`. Both client SDKs regenerate once more on their open Mercury branches so 1.17.0 carries the provider and this together.

## Testing

- `just test-code` green (ruff, format, basedpyright, cf-lint).
- `just test`: see the follow-up comment once the full run finishes; the event-block, MCP-instructions, model and extension-router suites pass locally (1,313 tests).
- New: `tests/operations/event_block/python_handlers/test_bank_feed.py` (metadata precedence, every posting shape, the refusals, the DSL fallback, preview), transition tests in `test_commands_update.py`, registry and instruction tests.
- Not run: a live classify-and-commit against the Mercury sandbox on a local graph; owed with the whole-loop run recorded in the spec.

## Certification

### DIFF
--- a/robosystems/middleware/mcp/tools/instructions.py
+++ b/robosystems/middleware/mcp/tools/instructions.py
@@ -119,6 +119,44 @@ def build_instructions(
       )
     )
 
+  # The inbox loop for bank-feed lines: classify, then commit.
+  if has("update-event-block") and has("list-event-blocks"):
+    inbox_lines = [
+      "INBOX / BANK FEED",
+      (
+        "- A bank feed (Mercury) lands every posted transaction `captured` "
+        "with `metadata.suggested_element_id` + `suggested_account_name`; "
+        "nothing posts until it is classified. Find them → "
+        "`list-event-blocks(source='mercury', status='captured')`."
+      ),
+      (
+        "- Classify one → `update-event-block(event_id, "
+        "transition_to='classified', metadata_patch={classified_element_id, "
+        "classified_by: 'claude', basis})`. A split: `classified_allocations: "
+        "[{element_id, amount}]` summing to the amount. To take the "
+        "suggestion as-is: `metadata_patch={accept_suggestion: true}`."
+      ),
+      (
+        "- Post it → `transition_to='committed'` (a person, or you when "
+        "asked): the handler writes DR/CR against the linked bank account as "
+        "a draft that close posts. An unclassified bank line is refused at "
+        "commit — never force it. Internal transfers are pre-classified "
+        "(both bank legs known) and commit as they are."
+      ),
+    ]
+    if has("recall") and has("remember"):
+      inbox_lines.append(
+        "- `recall` the counterparty before deciding, and `remember` each "
+        "decision ('Stripe payouts → Subscription revenue') so the next line "
+        "from that counterparty classifies itself."
+      )
+    if has("preview-event-block"):
+      inbox_lines.append(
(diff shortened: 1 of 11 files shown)


jfrench9:
@claude please review this PR

jfrench9:
`just test` on b1b12fe7: 14,506 passed, 42 skipped (the PR body deferred this line to the full run). Both client SDKs regenerated once more on their open branches (python #211, typescript #226) so 1.17.0 carries the provider and this transition together.

jfrench9:
Addressed:

- **Issue 1** — you're right that the flag on the transient journal metadata reached nothing. The handler now stamps `metadata.publish_to_source = False` on the persisted event before posting, on both the classified path and the DSL-floor path (`_pin_to_local_lane`, idempotent), so `writeback_source_clause` reads the pin from the row. The tests now assert `event.metadata_` rather than the mocked call arguments, and the rule-posted path is covered too.
- **Suggestion 2** — left as a local helper: `commands.py` imports the handler registry, so importing `_resolve_agent_type` from it into a handler module would be a cycle. If a third caller appears it can move to `engine.py`.

jfrench9:
Suggestion 2 taken after all: the counterparty lookup now lives in `engine.py` as `resolve_agent_type` (both `commands.py` and the bank-feed handler import the engine already, so there is no cycle). Event-block suite green.

PR #1384, opened 2026-09-12
fix(taxonomy): keep a mapped chart of accounts editable

jfrench9:
## Summary

Every `update-taxonomy-block` on a chart of accounts that had ever been mapped failed with one `phantom_to_ref` per mapping arc — a template-initialized chart from the first edit, a QuickBooks chart after the mapping operator ran. Found by the first local run of the Mercury feed, whose link-or-create step adds a chart account through the envelope and was refused by the template chart. Reproduced with a plain account add on the same chart; it is a Stage 0 defect that shipped in v1.11.19, not a Mercury one.

## Changes

- `operations/taxonomy_block/update_validator.py`: the projection of a block's existing associations into the virtual create request now leaves out arcs whose endpoint lies outside an **unparented** block. Those arcs (a chart's mapping arcs into the rs-gaap library) are created by the mapping commands, not by an envelope, and the create validator can only resolve a foreign endpoint through a parent taxonomy. Parented blocks (reporting extensions) keep projecting theirs, resolved through the parent as before.
- `tests/operations/taxonomy_block/test_update_validator_coa_traits.py`: a real-Postgres fixture with a chart mapped into a library concept; an account add and a rename both validate clean.

## Breaking Changes

None.

## Testing

- `just test-code` green; `tests/operations/taxonomy_block` 254 passed (real Postgres).
- On the local stack with the fix mounted: a plain `elements_to_update` on the template chart returned 200 where it had returned 422, and the Mercury link-or-create step then created 10 chart accounts through the envelope.

## Certification

### DIFF
--- a/robosystems/operations/taxonomy_block/update_validator.py
+++ b/robosystems/operations/taxonomy_block/update_validator.py
@@ -204,11 +204,27 @@ def validate_update_envelope(
     )
   virtual_structures.extend(payload.structures_to_add)
 
+  # Arcs that leave the taxonomy — a chart's mapping arcs into the rs-gaap
+  # library, or any block's arc to a concept it does not own — are created
+  # by the mapping commands, not by an envelope, and the create validator
+  # can only resolve a foreign endpoint through a *parent* taxonomy
+  # (``_load_library_qnames``). Projecting them for an unparented block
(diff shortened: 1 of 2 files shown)


jfrench9:
@claude please review this PR

PR #1385, opened 2026-09-13
fix(ledger): a classified bank line is one the commit can post

jfrench9:
## Summary

Found in the second local end-to-end run of the Mercury bank feed, walking the inbox loop over MCP. A bank line could be marked `classified` with a choice the commit could not post (`accept_suggestion: true` on a suggestion the chart never matched), and the refusal that followed at commit told the caller to set `accept_suggestion` — the step they had just taken. The classify transition now runs the same plan the commit runs and refuses with the actual reason; the preview names its handler; the per-graph MCP instructions say what a split's amounts are and what `preview-event-block` needs.

## Changes

- `operations/event_block/python_handlers/types.py` — `EventBlockPythonHandler.validate_classification`, an optional hook a handler declares to veto `captured → classified` when the metadata records no choice it could post. Handlers without it (all but the bank feed) behave as before.
- `operations/event_block/python_handlers/bank_feed.py` — `unclassified_reason` names the case the caller is in (accepted but unresolved suggestion / suggestion available / nothing suggested) and is used by the commit-time refusal and the preview; `validate_classification` runs `plan_lines` on classify; `BankFeedMetadata` gains `suggested_account_name` so the refusal can name the suggestion.
- `operations/event_block/commands.py` — `update_event_block` calls the hook after the metadata patch on a `classified` transition (refusals surface as 422 like a commit's); `_python_preview_to_response` puts the handler's display name in `handler_metadata.handler`, since `matched_handler` is the DSL row shape and stays empty for a Python handler.
- `middleware/mcp/tools/instructions.py` — the INBOX / BANK FEED block: split amounts are positive cents summing to the line's absolute amount; `accept_suggestion` applies only when `suggested_element_id` is set; classifying validates; `preview-event-block` takes the line's fields plus the classification.
- Tests: the new reason branches, the classify hook (resolvable choice passes, name-only suggestion refused with the reason, a split that does not add up, an internal transfer as-is), the command wiring (refusal with reason, handlers without the hook untouched), the preview's handler name.

## Breaking Changes

None. No API shape changes — `handler_metadata` gains a key, and a `classified` transition that used to succeed silently on an unpostable choice now returns the same 422 the commit would have. Generated SDK tier unaffected.

## Testing

- `uv run pytest` on the five touched test files: 145 passed.
- `ruff check` + `ruff format` clean; `basedpyright` on the three changed modules: 0 errors.
- Verified live on the local stack against the Mercury sandbox graph over MCP: classify with `accept_suggestion` on a name-only suggestion → refused naming 'Office Supplies' as unmatched; commit of an unclassified line → refused with the choice to make; preview → `handler_metadata.handler = "Bank Transaction"`.

## Certification

https://claude.ai/code/session_01EmuU6KfAcqtfL8pfhRJWW7

### DIFF
--- a/robosystems/middleware/mcp/tools/instructions.py
+++ b/robosystems/middleware/mcp/tools/instructions.py
@@ -133,8 +133,12 @@ def build_instructions(
         "- Classify one → `update-event-block(event_id, "
         "transition_to='classified', metadata_patch={classified_element_id, "
         "classified_by: 'claude', basis})`. A split: `classified_allocations: "
-        "[{element_id, amount}]` summing to the amount. To take the "
-        "suggestion as-is: `metadata_patch={accept_suggestion: true}`."
+        "[{element_id, amount}]` — positive cents summing to the line's "
+        "absolute amount. To take the suggestion as-is: "
+        "`metadata_patch={accept_suggestion: true}` (only when "
+        "`suggested_element_id` is set; a name-only suggestion needs a "
+        "choice). Classifying validates the choice — one that resolves no "
+        "account is refused there, with the reason."
       ),
       (
         "- Post it → `transition_to='committed'` (a person, or you when "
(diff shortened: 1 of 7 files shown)


jfrench9:
@claude please review this PR

PR #57, opened 2025-11-04
Refactor authentication responses to include token metadata

jfrench9:
## Summary
This refactor enhances the authentication system by standardizing JWT token responses across all authentication endpoints and improving the session refresh mechanism. The changes provide clients with essential token metadata needed for better token lifecycle management.

## Key Changes
- **Enhanced Authentication Response Model**: Extended the base authentication response to include token expiry timestamp and refresh threshold information
- **Standardized Token Metadata**: All authentication endpoints (login, registration, email verification, password reset, SSO) now return consistent token expiration and refresh timing data
- **Improved Session Refresh Logic**: Refactored JWT token creation process in the session refresh endpoint to be more maintainable and consistent with other authentication flows

## Key Accomplishments
- ✅ Unified authentication response format across 6 different auth endpoints
- ✅ Improved client-side token management capabilities by exposing expiry and refresh metadata
- ✅ Streamlined session refresh implementation for better code maintainability
- ✅ Enhanced authentication system consistency without breaking existing functionality

## Breaking Changes
None. This refactor maintains backward compatibility while adding new optional fields to authentication responses.

## Testing Notes
- Verify all authentication endpoints return the new token metadata fields
- Test session refresh functionality continues to work as expected
- Confirm existing client integrations remain unaffected by the additional response fields
- Validate token expiry and refresh threshold values are accurately calculated and returned

## Infrastructure Considerations
This change affects all authentication-related API responses. Monitor authentication endpoint performance and ensure logging systems can handle the additional metadata fields in responses.

---
🤖 Generated with [Claude Code](https://claude.ai/code)

**Branch Info:**
- Source: `refactor/improved-refresh`
- Target: `main`
- Type: refactor

Co-Authored-By: Claude <noreply@anthropic.com>

PR #1386, opened 2026-09-13
feat(views): disclosures and information-block read a report's sections whole

jfrench9:
## What

The two shaped tools `xbrlkit serve` runs over a loaded filing — `disclosures` (the map of a report's sections) and `information_block` (one section read whole: rows in presentation order, the same rows by the section's own axes, calculation footing, text blocks) — served over a report the platform holds.

**A whole filing is a file; the graph answers what crosses filings.** A section read whole wants the whole report in memory, and the platform has two ways to get one that never touch the graph:

- **Shared repository (`sec`):** the published holon the pipeline writes per filing to the public data bucket, read by xbrlkit's holon reader in-cluster, its text-block fragments read beside it. A filing processed before the artifacts existed answers 404 "not published yet" until the repository is reprocessed. Never a whole-report walk of the corpus-scale graph, which was measured at 30–60 s of replica IO per filing.
- **Tenant graph:** the report bundle the Tavi and holon exports already build from the extensions database (`build_report_bundle` → `bundle_to_xbrl_model`), live. Two gaps in that bridge came out of running the block tools over it and are fixed here: the link's `block_type` and the element's standard label were dropped on the way into the model.

The model is cached per report in `MCP_CACHE` (compressed; 6 h for a published filing, 5 min for a tenant's live report; a cache outage costs a read, not the call). Reading a holon into the model and freezing or thawing the cached copy are CPU work of a few hundred milliseconds on a 10-K, so they run off the event loop like the fetch. Report resolution is the `financial-statement-analysis` contract: `ticker` on a shared repo, `report_id` on a tenant; a tenant subgraph is told to read on its parent.

Surfaces: view operations `disclosures` and `information-block` under `/extensions/roboledger/{graph_id}/operations/` (envelope; 400 for a selector error, 404 for a report, block or artifact not found), the MCP tools of the same names beside `financial-statement-analysis` on shared repos and tenants, read-only, in the generated instructions' REPORTING line and the authored `sec` instructions. The block rules stay in the xbrlkit library, imported by the API process; the local `xbrlkit serve` is not in any path.

## Deployment wiring

The API and worker task definitions carried neither the public data bucket's name nor a grant on it — only the Dagster stack did, because only the pipeline wrote there. The bucket is CloudFront-OAC-only, so a read from inside the cluster needs the role's own grant. `api.yaml` and `worker.yaml` gain a `PublicDataBucketArn` parameter, a read-only `s3:GetObject` / `s3:ListBucket` statement, and `PUBLIC_DATA_BUCKET` in the environment; `deploy-api.yml`, `deploy-dagster.yml` (the worker stack), `prod.yml` and `staging.yml` pass the S3 stack's existing output through. Without this every `sec` call in prod would have answered "not published yet".

## Measured on the local stack

| call | result |
|---|---|
| cold `disclosures`, 3M FY2024 10-K from LocalStack's holon (12 MB) | 0.7 s |
| warm map, family, or any block | 0.06–0.09 s |
| cold `disclosures` on a demo tenant's published FY2025 report | 0.28 s |

Both substrates were exercised end to end over REST and the remote MCP surface: the map, a family, a block with its footing (`block_type` carried on the tenant), the 404 for a filing without artifacts, the 404s for an unknown ticker or block, the 400 for a missing selector, and a non-numeric cap argument answered as an error rather than a traceback.

## Tests

The view tests let xbrlkit's own tools run unmocked over both substrates — a fake bucket and a fake coordinates lookup for the published filing, a bundle fixture for the tenant — plus the seams: resolution (including the tenant-subgraph refusal), cache TTLs and outage, an unreadable fragment, a filing without artifacts. Router, tool, manager, instructions, read-only-set and description-contract tests extended. `cfn-lint` and `actionlint` pass on the templates and workflows.

## Dependency

Moves xbrlkit to 0.15.0, which carries #57: Calculations 1.1 read as the calculation linkbase, `xbrldt:targetRole` written and read back, the property-graph projection keeping a type's namespace, and the CDN-first `load_filing` for the local server. The processor writes holons through this release, so the corpus reprocess that gives `sec` its coverage must run after this deploys, not before. `MODEL_CACHE_VERSION` moves to 3 so cached models are re-read with the newer reader.

https://claude.ai/code/session_01GD7d1Rjg5PPV8swJUDFdyr

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -39,7 +39,7 @@ Extensions (roboledger, roboinvestor) live under `/extensions` and are
 
 - **Typed reads** → `POST /extensions/{graph_id}/graphql` (Strawberry + GraphiQL in dev)
 - **Command writes** → `POST /extensions/{roboledger|roboinvestor}/{graph_id}/operations/{operation_name}`
-- **Analytical view operations** → `POST /extensions/{domain}/{graph_id}/operations/{view_name}` — graph-backed operations that query LadybugDB rather than the extensions OLTP database. Read-only, same envelope contract as command writes. `build-fact-grid` (pivot tables over the XBRL hypercube), `financial-statement-analysis`, and the pair `disclosures` (the map of a report's sections) / `information-block` (one section read whole — xbrlkit's own tools run over the report's slice read out of the graph); gated independently of the OLTP domain flags so deployments without the corresponding tenants can still mount them
+- **Analytical view operations** → `POST /extensions/{domain}/{graph_id}/operations/{view_name}` — graph-backed operations that query LadybugDB rather than the extensions OLTP database. Read-only, same envelope contract as command writes. `build-fact-grid` (pivot tables over the XBRL hypercube), `financial-statement-analysis`, and the pair `disclosures` (the map of a report's sections) / `information-block` (one section read whole — xbrlkit's own tools run over the report held whole: the published holon on `sec`, the report bundle on a tenant; the graph is not in that path); gated independently of the OLTP domain flags so deployments without the corresponding tenants can still mount them
 
 All three surfaces take `graph_id` as a URL path parameter — auth + per-graph
 access are validated by FastAPI dependencies before the handler runs.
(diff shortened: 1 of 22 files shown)


jfrench9:
@claude please review this PR

jfrench9:
@claude can you review

PR #1387, opened 2026-09-13
feat(sec): the corpus index says which filers have a page to render

jfrench9:
## Summary

The public company pages on roboinvestor.ai render a filing from its Tavi model or its holon; a filing carrying only the document as filed has nothing to project. Which filers those are has been unanswerable from the corpus index, so a consumer deciding which filers to list — or to put in a sitemap — would have to read all ~6,800 per-filer catalogs to find out. Each index row now carries `renderable` and `latest_renderable`.

This matters now because the corpus reprocess is about to fill those artifacts in filer by filer: a 40-filer sample of the live corpus today finds none with a single renderable filing, so the flag also has to stay current as the reprocess lands rather than being computed once.

## Changes

- **`robosystems/adapters/sec/pipeline/catalog.py`**
  - `renderable()` and `renderable_summary()`: a filing counts when it has a `holon` or `tavi` representation. `RENDERABLE_KINDS` is deliberately stricter than the existing `latest`, which counts *any* representation including a document-only filing — the two answer different questions, and the docstrings now say which.
  - `latest_renderable` names the filing a page would actually show. That is not `latest`: the newest filing of any kind is often a quarter whose artifacts are not written, and using its date as a page's last-modified would claim a change that did not happen.
  - `read_prior_renderable()`: the index is built for every listed filer while manifests are read only for the filers a run rewrites, so an untouched filer's renderability is carried forward from the index the previous run wrote — one GET, against a manifest read per filing of every filer otherwise. A `full_rebuild` touches everything and carries nothing forward. A missing, unreadable, or pre-flag index is not an error: those rows read as not-renderable and correct themselves on the next run that touches the filer, so the rollout needs no backfill.
  - `index_row()` takes the summary as an optional third argument and defaults to not-renderable without one; `build_company()` folds the same summary into the per-filer document.
  - `CATALOG_VERSION` 1 → 2. The asset now logs the renderable count and reports it as `filers_renderable` metadata, so the reprocess can be watched through it.
- **`tests/adapters/sec/pipeline/test_catalog.py`** — two new classes covering the predicate, the newest-renderable-is-not-newest distinction, the absent-summary default, and the carry-forward path including a missing index and an index written before the flag existed. The existing `build_company` test additionally asserts the new keys.

Additive JSON fields only; no existing key changes shape or meaning.

## Breaking Changes

None. `renderable` and `latest_renderable` are additive fields on a public CDN artifact, and no API or SDK surface changes — the catalog is a file on the public data bucket, not an endpoint. Consumers that ignore the fields are unaffected.

## Testing

- `just test` on `tests/adapters/sec/pipeline/test_catalog.py` — 17 passed.
- `just format` and `just lint` — all checks passed, 1889 files unchanged.
- `just typecheck adapters/sec/pipeline/catalog.py` — 0 errors, 0 warnings, 0 notes.
- Not exercised: a live materialization of the asset. The carry-forward path is covered by unit tests against a mocked S3 client rather than a real index read.

## Certification


jfrench9:
@claude please review this PR

PR #1389, opened 2026-09-14
feat(admin): sweep search orphans by when they were indexed

jfrench9:
## Why

A text re-index leaves orphans behind. Document ids are content-independent — `sha256(graph:source:accession:section_id)` plus the part number — so a section that merges with another, splits into a different number of parts, or crosses the `part_size` boundary stops regenerating the id it used to have. The old document stays in the index forever, holding stale text.

Sweeping them needs the date we last *wrote* a document. `search delete --before` could not do it: it filters `filing_date`, which is a property of the filing, not of the index.

This is the prerequisite for the orphan sweep in the SEC text re-index procedure, which runs after the current corpus rebuild.

## What

`--indexed-before`, filtering `indexed_at` — the field `bulk_index` stamps on every write.

The two date filters stay separate rather than overloaded. `--before` still trims by `filing_date` for retention; they compose. The confirmation prompt now names which field it is about to use, because conflating them deletes the wrong half of a corpus and the old prompt said only "filing_date" regardless.

No deploy is involved: the bastion script is built from the local checkout, base64-encoded and shipped over SSM.

## Tests

The script that runs on the bastion is a **string** built by `str.format`, so neither ruff nor basedpyright ever looks at it — a brace-escaping slip or a wrong field name would first surface in production, mid-delete. `tests/admin/test_search_script.py` renders the template exactly as the dispatcher does and holds the invariants the toolchain cannot:

- it parses as valid Python for every action
- every format parameter lands as a module-level literal
- `--indexed-before` filters `indexed_at` and `--before` filters `filing_date`, neither bleeding into the other
- all four narrowings compose in order
- **a delete always carries the `graph_id` term**

That last one is the load-bearing case. `OPENSEARCH_INDEX` is a single shared `documents` index holding every tenant's uploaded documents alongside SEC, isolated only by that filter, so losing it is data loss rather than a bug — the test asserts it holds even when every optional filter is absent.

Both invariants were mutation-checked: swapping `indexed_at` for `filing_date`, and dropping the `graph_id` term, each fail exactly the test that claims to guard them.

## Risk

Additive. Existing invocations are unchanged — the new filter is only applied when the option is passed.

https://claude.ai/code/session_01GD7d1Rjg5PPV8swJUDFdyr


jfrench9:
@claude please review this PR

jfrench9:
Both notes addressed in 8d4e1d66.

**Action names** — fixed. The list had the CLI subcommand names rather than the values the template branches on. Your read of why it was harmless is right: `ast.parse` covers the whole rendered script regardless of which value `action` holds. Corrected anyway, since a later test asserting runtime behaviour off that list would have exercised nothing and passed.

**Parameter substitution** — folded in rather than deferred. Not because it crosses a privilege boundary; it doesn't, the caller already holds prod SSM. Two reasons it was worth doing here: the mitigation already existed for one field, with a comment naming the concern, and was simply never generalised — so this is finishing an existing decision, not opening a new one. And this branch added the fifth field to the ungeneralised half while a runbook now tells an operator to paste a captured timestamp into exactly that argument, so the sharp edge became something the documented procedure walks into.

Every scalar now goes through `json.dumps` on the way in and the template stops supplying its own quotes, which makes the encoding load-bearing rather than belt-and-braces — the template no longer parses without it. `size` stays an int, and query text keeps travelling base64 in the env as before.

Tests extended with values that would otherwise end their own literal, asserting both that the script still parses and that the value arrives unchanged — escaping that mangles the input would be its own defect. Mutation-checked: reverting the helper to raw substitution fails the render test.

15 tests in the file, `tests/admin/` 85 green, `just test-code` clean.

PR #1391, opened 2026-09-14
fix(serialization): carry a tenant concept's label into the holon

jfrench9:
## Summary

A tenant-authored concept came out labelled in the XBRL export of a report and unlabelled in the holon of the same report. Three encoders read `BundleElement.label` and disagreed about what a concept is called: the XBRL label linkbase fell back to `name`, the JSON-LD encoder and the XbrlModel bridge did not. The label was never missing from the data — `create-taxonomy-block` takes the display name and derives the QName from it, so the wording sits on `Element.name`, while `label` is sourced from a standard `ElementLabel` row or the description and only the library seeding path writes those rows. `cadence:RevenueRecognitionPolicyTextBlock` reached the bundle as `name="Revenue Recognition Policy"`, `label=None`.

The fallback moves to `concept_label()` beside `BundleElement` — the model every encoder reads — and all three call it, so they cannot drift apart again.

## Changes

**`serialization/bundle.py`**

- New `concept_label(concept)`, the single label rule: the authored label (standard label-linkbase entry, else description), else `name`, skipping a `name` that merely echoes the QName local part (`"Assets"` for `rs-gaap:Assets`) since that adds nothing over the element declaration. Lifted from the XBRL serializer, where it already lived and was already correct.

**`serialization/xbrl/xbrl_21.py`**

- `_concept_label` is now an alias for the shared function; its body was identical apart from calling the module-local `_local_name`. The emitted `report-lab.xml` is unchanged. `_local_name` stays — it still supplies the NCName `name` attribute on `<xs:element>` declarations.

**`serialization/rdf/jsonld.py`**

- `skos:prefLabel` comes from `concept_label` rather than the raw `label` field.

**`serialization/model.py`** — the change that actually fixes the holon, and worth a close look

- `serialize_to_holon_jsonld` does not use the JSON-LD encoder: it re-expresses the bundle as an `XbrlModel` through `bundle_to_xbrl_model` and hands it to xbrlkit's writer. So `_concept` is a third reader of `element.label`, and the one the holon flows through — fixing the flat encoder alone left the holon unlabelled, which a round-trip caught.
- It also cannot be recovered downstream: `_concept` sets xbrlkit's `Concept.name` to the QName local part, so xbrlkit's own `pref_label or name` fallback resolves to `RevenueRecognitionPolicyTextBlock`. The label has to arrive as `pref_label`, which now comes from `concept_label` — for the standard-label entry too, so the holon and the Tavi flavor both carry it.

**`tests/operations/serialization/test_jsonld_encoder.py`**

- Five cases in `TestConcepts`: a tenant concept is labelled from its name; a name echoing the QName local part is not emitted as a label; both encoders agree on the same concept; and the bridge carries the label, asserting `pref_label` and the standard-label entry plus the negative case. The existing fixture could not have caught this — its unlabelled concept is `rs-gaap:AssetsCurrent` with `name="AssetsCurrent"`, which the echo rule correctly skips.

No migration and no backfill: the fix reads a field every existing report already carries, so previously-exported reports come out labelled on their next export.

## Breaking Changes

None.

No change to the GraphQL schema, the operations envelope, or any REST shape, so no SDK coordination is needed — neither the stable nor the generated tier is touched.

Two output notes for reviewers, both additive rather than breaking:

- Holon, flat JSON-LD and TAVI now emit `skos:prefLabel` for tenant-authored concepts where they previously emitted nothing. Consumers that read the QName local part as a display fallback will start seeing the tenant's own wording, which is the intent.
- The XBRL 2.1 label linkbase is unchanged — that arm already had this fallback and now reaches it by delegation.

## Testing

- `just test-code` — ruff check, ruff format, basedpyright: all clean, 0 errors.
- `just test` — **14,600 passed, 42 skipped**. (Run as the two halves rather than `just test-all` in one invocation.)
- End-to-end round trip, since the bridge defect was invisible to the unit tests and to reading the call graph: built a bundle carrying a native text block with `name="Revenue Recognition Policy"` and no authored label, serialized it with `serialize_to_holon_jsonld`, confirmed `skos:prefLabel` on the emitted `rs:Element`, then loaded the written file into xbrlkit and ran `search_text`. Its hit routing reported `"section": "Revenue Recognition Policy"` where the same call previously reported `"RevenueRecognitionPolicyTextBlock"`.

## Certification


jfrench9:
@claude please review this PR

Issue #1393, opened 2026-09-14
Committed demo sample artifacts have no freshness gate — SHACL conformance passes while content rots


jfrench9:
### What

Add a gate over the committed demo artifacts in `examples/*/sample_output/` that fails when they no longer match what the encoders produce. Today the only check over them is SHACL shape conformance, which passes while their *content* silently rots.

### Why

`tests/operations/serialization/test_sample_bundles_shacl.py` says in its own docstring:

> The demos' `sample_output/*.jsonld` are committed reference artifacts. … if an encoder change (or a hand-edit of a sample) drifts the shape, this fails loudly.

It validates SHACL conformance, so shape drift does fail. Content drift does not, and the gap is not theoretical — two separate drifts were sitting in the tree undetected:

**1. Element labels vanished from every committed holon.** Comparing the committed SaaS sample against a current export of the same demo:

```
committed (regenerated 2026-09-09, 665c91be)
  prefLabel on: rs:Structure 6, rs:InformationBlock 6, rs:Entity 1   → 0 elements
current export
  prefLabel on: rs:Structure 6, rs:InformationBlock 6, rs:Entity 1,
                rs:Element 38
```

All five committed holons carry zero element labels. A reader of those files sees QNames where a name is expected, and the SHACL suite is green.

**2. Tenant-authored concepts lost their labels in the flat JSON-LD.** Fixed in #1391, but the samples predate it:

```
saas-startup-demo.jsonld     5 native concepts missing prefLabel
coffee-roaster-demo.jsonld   8 native concepts missing prefLabel
roboledger / seattle-method / world-online   unaffected — no native concepts
```

Both classes are the kind of thing these artifacts exist to demonstrate. They are referenced as reference output, so a stale one misinforms whoever reads it — and #1391's own defect was a label that reached one representation and not another, exactly what a fresh sample set would have shown.

### Scope - Affected

- [ ] `tests/operations/serialization/test_sample_bundles_shacl.py` — either extend it or add a sibling test module
- [ ] `examples/*/sample_output/` — the artifact sets under gate (`.jsonld`, `.holon.jsonld`, `.tavi.json`, `.zip` × 5 demos)
- [ ] The test-module docstring, which currently overstates what it checks

### Scope - Out of Scope

- Regenerating the samples themselves — tracked separately, and a prerequisite rather than part of this.
- Standing the local stack up in CI. See Notes: a byte-diff regen gate is the wrong shape here.

### Acceptance Criteria

[...]

PR #1394, opened 2026-09-14
chore(examples): regenerate the demo sample artifacts

jfrench9:
## Summary

Regenerate the committed demo reference artifacts in `examples/*/sample_output/`. Two drifts had accumulated there, neither of them visible to the only gate over the set — `test_sample_bundles_shacl.py` validates SHACL *shape*, so content rot passes green. Closes nothing on its own; #1393 tracks gating the content so this does not recur.

## Changes

Regenerated all five demos against merged `main` (`197618d6`) and copied like-for-like into `sample_output/`.

**The larger drift predates #1391.** Every committed holon carried `skos:prefLabel` on structures, information blocks and the entity, but **none on elements** — a reader of those files saw QNames where a name belongs:

| demo | holon elements labelled, before → after |
| --- | --- |
| coffee-roaster | 0/146 → 146/146 |
| roboledger | 0/127 → 127/127 |
| saas-startup | 0/140 → 140/140 |
| seattle-method | 0/134 → 134/134 |
| world-online | 0/128 → 128/128 |

The set was last regenerated 2026-09-09 (`665c91be`), so this has been the state of the committed holons since then.

**The smaller drift is #1391**, and reaches only the two demos that have tenant-authored concepts, in the flat JSON-LD:

| demo | flat JSON-LD elements labelled, before → after |
| --- | --- |
| coffee-roaster | 93/101 → 101/101 — the 8 `driftline:` concepts |
| saas-startup | 93/98 → 98/98 — the 5 `cadence:` concepts |
| roboledger / seattle-method / world-online | 93/93 → 93/93, unaffected (no extension concepts) |

**What was copied.** Every filename already committed had a counterpart in the demo's `output/`, so each was replaced in place; nothing was added or removed. The leftover `*-preclose` / `*-postclose` artifact sets sitting in the gitignored `output/` directories from earlier runs were deliberately **not** added.

## Breaking Changes

None. Reference artifacts only — no code, no API surface, no SDK implication.

## Testing

- `just test-code` — ruff check, ruff format, basedpyright: clean, 0 errors.
- `uv run pytest tests/operations/serialization` — **214 passed**, including `test_sample_bundles_shacl.py` over the regenerated files, so the new set is still SHACL-conformant.
- Arelle, per the regenerated validation reports: **0 load errors and 0 validation errors** on all five XBRL packages.
- Three SHACL validation reports came back byte-identical, which is the expected result for a conformance report over an unchanged shape.

Reviewers should know the diff is noisier than the content change: these artifacts embed a run's own graph id and report id (`output/.gitignore` says so), so a regeneration rewrites those strings throughout regardless of what actually changed. That non-reproducibility is also why #1393 argues for an invariant gate rather than a regenerate-and-diff one.

## Certification


jfrench9:
@claude please review this PR

PR #1395, opened 2026-09-14
fix(serialization): a disclosure's calculation arcs are bundled once

jfrench9:
## What

A tenant-authored disclosure note footed to exactly **twice** its reported total in every renderer that checks footing:

```
Inventory Components / rs-gaap:InventoryNetOf…      expected 44,000      actual 22,000
Disaggregation of Revenue / …ExcludingAssessedTax   expected 385,599.60  actual 192,799.80
```

The graph holds four calculation arcs for that note. The bundle held eight — byte-identical pairs, same `from`, `to`, `order`, `weight`, `arcrole`, on the same link.

## Why

`_build_bundle` loads every association on the rendered structures. For a tenant disclosure that includes its **own** calculation arcs, because `create-taxonomy-block` writes presentation *and* calculation onto the single structure the note renders from. `_source_calculation_arcs` then ran a second, structure-blind query (`association_type == "calculation"` with both endpoints declared) and re-hosted each arc under the structure carrying both endpoints — which for that note is itself. Its `seen` set only deduplicated within its own output, never against the list it was handed.

A library statement never hit this: its calculation arcs live on separate `rs-gaap-calculation` structures the direct load never touches. That is precisely what the call-site comment asserted of *all* calculation arcs, and why the tenant case stayed invisible.

## The fix

The helper is now handed those arcs in order to **exclude** them: `seen` is seeded with every `(structure_id, from, to)` already present. Exclusion is per structure, so an arc legitimately re-hosted onto a *different* ELR is still a distinct arc there and is kept.

The parameter is renamed `presentation_associations` → `loaded_associations`. The old name asserted the list could not contain calculation arcs — the assumption that hid this — and the call-site comment is corrected to match.

## Verification

Not only unit tests. Regenerated both affected demos against the fix:

| | calc arcs | note's arcs | subtotals footed | mismatches |
|---|---|---|---|---|
| coffee, before | 76 (72 distinct) | 8 | 34 | **2** |
| coffee, after | 72 (72 distinct) | 4 | 34 | 0 |
| saas, before | 72 (70 distinct) | 4 | 34 | **2** |
| saas, after | 70 (70 distinct) | 2 | 34 | 0 |

Footing verified clean through all three projections — dataset-form holon, flat JSON-LD, and Tavi. The note's four arcs now match the four in the graph exactly.

The SHACL reports capture the change precisely: coffee `177 → 173` associations, saas `170 → 168`; fact and element counts untouched. The XBRL 2.1 validation reports are unchanged — that arm strips disclosures via `_strip_disclosure_content`, so it was never affected (and would have had Arelle reject the duplicates if it weren't).

## Tests

Three cases added to `test_calc_sourcing.py`:

- a disclosure's own arcs are not sourced again (fails without the fix, reproducing the duplicate)
- the same parent→child **is** still sourced onto a different structure — exclusion is per ELR, not per arc
- a library statement's off-network arcs are still sourced, or the calculation linkbase ships empty

The `_pres` stub now carries `association_type`, mirroring the row it stands for.

`just test-code` clean; 1,312 tests pass across `tests/operations/serialization/` and `tests/operations/roboledger/`.

## Samples

Refreshed for the two affected demos only. The other three carry no tenant-authored numeric disclosure and were verified byte-unaffected, so they are left alone rather than churned with new run ids.

https://claude.ai/code/session_01GbYwabmPG1vskYNWTVWHvd


jfrench9:
@claude please review this PR

Issue #1397, opened 2026-09-14
The flat bundle and the holon name the same things differently


jfrench9:
### What

`ref/serialization.md` states the invariant that the flat JSON-LD bundle and the holon "are two forms of one Dataset and name the same things the same way". It is violated in two places, by the same mechanism.

**Concepts.** Expanded to absolute IRIs, from the committed coffee-roaster sample:

```
library concept
  flat   https://robosystems.ai/taxonomy/rs-gaap/v1/Assets
  holon  https://robosystems.ai/taxonomy/rs-gaap/v1/Assets          ✓ same node

tenant concept
  flat   https://robosystems.ai/concept/driftline:InventoryRawMaterials
  holon  https://robosystems.ai/taxonomy/driftline/InventoryRawMaterials   ✗ two nodes
```

Five concepts in the SaaS demo, eight in the coffee one. The flat form's IRI also carries a colon inside a path segment, which no consumer will read as a QName.

**Structures.** The same report's equity statement, from the two forms:

```
flat   structureName / prefLabel = "rs-gaap — Statement of Changes in Equity — Roll Forward (Total)"
holon  structureName / prefLabel = "0004 - Statement - Statement of Changes in Equity"
```

The holon's writer sets both to the composed sort key, so the **form** name — *Classified*, *Multi-step*, *Indirect*, *Roll Forward* — is absent from the holon entirely. For a ledger report that is not decoration: it says which of several presentations of the same accounts this is. It is recoverable only from the role-URI slug (`…/Equity-rollforward`). There is no `rs:roleDefinition` or documentation term beside it. This is finding 3 of `specs/xbrlkit/three-renderings-disagree.md`, folded in here.

### Why

`rdf/jsonld.py::_concept_uri` resolves a prefix through a static `_PREFIX_NS` map and falls through when it misses:

```python
def _concept_uri(qname: str) -> URIRef:
  if ":" not in qname:
    return URIRef(f"https://robosystems.ai/concept/{qname}")
  prefix, local = qname.split(":", 1)
  ns = _PREFIX_NS.get(prefix)
  if ns is None:
    return URIRef(f"https://robosystems.ai/concept/{qname}")   # ← every tenant
  return URIRef(str(ns) + local)
```

[...]

PR #1398, opened 2026-09-14
test(serialization): gate the demo artifacts on content, not just shape

### LINKED ISSUE #1393: Committed demo sample artifacts have no freshness gate — SHACL conformance passes while content rots

jfrench9:
Closes #1393.

## What

Four invariants over the committed artifacts in `examples/*/sample_output/`, as pure Python over checked-in JSON — no stack, no live API, runs in about a second.

SHACL conformance (`test_sample_bundles_shacl.py`) checks their *shape*, which is why two content drifts sat in the tree unseen: every committed holon carried zero element labels, and tenant concepts had lost theirs in the flat JSON-LD.

| invariant | the drift it catches |
|---|---|
| every `rs:Element` carries a `prefLabel` | labels vanishing (#1391 / #1394) |
| the two forms agree on the label of every shared concept | a label reaching one representation and not the other |
| every arc endpoint is declared as an element | the holon partition that seeded its scene from facts alone |
| no calculation arc appears twice on one role | the disclosure double-source (#1395) |

## Verified against the artifacts as they actually were

The issue's own acceptance criterion. Extracted each historical tree and ran the checks:

```
665c91be  (pre-#1394)
  ✗ labelled       coffee.holon 146 · roboledger.holon 127 · saas.holon 140
                   seattle.holon 134 · world-online.holon 128  unlabelled elements
  ✗ labelled       coffee.jsonld 8 · saas.jsonld 5             (the flat-bundle drift)
  ✗ duplicate arcs 4 artifacts
  → 7 of 10 artifacts violate the label invariant

a29f509a  (pre-#1395)
  ✗ duplicate arcs coffee ×4, saas ×2, both forms
  → 4 artifacts

HEAD
  no invariant violated
```

Both drifts the issue names, plus this session's own defect, all caught. Every message names the artifact and the invariant, so the fix reads "regenerate that demo".

## It found a live defect on its first run

Now **#1397**. Expanded to absolute IRIs:

```
library concept   flat  …/taxonomy/rs-gaap/v1/Assets
                  holon …/taxonomy/rs-gaap/v1/Assets                      ✓ same node
tenant concept    flat  …/concept/driftline:InventoryRawMaterials
                  holon …/taxonomy/driftline/InventoryRawMaterials        ✗ two nodes
```

`rdf/jsonld.py::_concept_uri` resolves a prefix through a static `_PREFIX_NS` map and falls through for anything it lacks — which is every tenant namespace, since those are minted per tenant. So the flat-vs-holon identity invariant `ref/serialization.md` states fails for every concept a tenant authored.

That is the **third instance of one defect class** (after #1391 and report-components#54): a static prefix table cannot know a tenant namespace. The first two were label and display; this one is identity.

Recorded as `xfail(strict=True)` on exactly the two demos that author concepts — computed from the artifact, not hardcoded — so fixing #1397 flips the test rather than leaving a green lie. The other three demos assert it normally and pass.

## Two smaller things in scope

**The XBRL validation report now states its scope.** It said "✅ Valid XBRL 2.1 · Facts loaded by Arelle: 69" beside a holon with 86 facts and six sections, with nothing explaining the gap — `_strip_disclosure_content` excludes tenant notes because that emitter's fixed framework prefixes cannot declare an extension concept. One sentence in the generator, and the five committed reports re-run from their committed zips (`validate_arelle` reads a zip from disk, so no stack). The diff in each is that sentence and nothing else — fact counts and error counts are byte-identical, which also confirms the committed reports were faithful.

**The SHACL module's docstring** said a drift "fails loudly". It now says shape only, and points at the companion.

## Verification

256 passed, 2 xfailed in `tests/operations/serialization/`. `just test-code` clean.

https://claude.ai/code/session_01GbYwabmPG1vskYNWTVWHvd


jfrench9:
@claude please review this PR

PR #1400, opened 2026-09-14
feat(alarms): report free tenant slots per tier, and fix the capacity alarm that never matched

jfrench9:
Two related changes to graph fleet capacity signalling. The second is the one that matters; the first is the bug that made it necessary.

## 1. `DatabaseCapacityAlarm` watched a stream nothing writes to

It declared no `Dimensions`, while `allocation_manager.py` `_publish_allocation_metrics` publishes `DatabaseUtilizationPercent` under `[NodeType=writer]`. CloudWatch matches on the exact dimension set, so it resolved to nothing and `TreatMissingData: notBreaching` read that as `OK`. `StateUpdatedTimestamp` was still its creation transition of **2025-12-23**. Same defect class as `AllocationFailureAlarm` (#1177).

`cfn-lint` cannot see this, so the check is `get-metric-data` against the live namespace — 20-day window, `Period: 300`:

| query | datapoints |
|---|---|
| **as fixed** — `[NodeType=writer]` | **17** (14 above the 85 threshold, max 100.0) |
| as deployed — no dimensions | 0 |
| `[NodeType=reader]` / `[Environment=prod]` / typo'd name / wrong namespace | 0 |

Also `EvaluationPeriods: 2 → 1`. The metric publishes only on allocate/deallocate, so requiring two *adjacent* event-bearing windows was a second, independent reason it could not fire. `TreatMissingData` stays `notBreaching` deliberately — reasoning in the template comment, but briefly: the destination is a monitored inbox, and a latched alarm sends no further mail.

## 2. `TenantSlotsFree` — the signal that fires *before* a refusal

A per-tier count of placeable slots, published every 5 minutes by the existing `instance_metrics_collection_job`, alarmed at `< 2` on `ladybug-standard`.

**Why a count per tier rather than repairing the fleet percentages.** `ClusterUsedCapacityPercent` sums every tier, and shared-repository instances contribute 10 slots each (`TIER_CAPACITY_MAP`). On 2026-08-31 it read **30.77%** while Standard was **100% full and refusing signups**, because 20 of its 26 "slots" were SEC replica slots no tenant can be placed on. That is not a denominator bug — a tier with headroom will always mask a tier that is full. `TENANT_TIERS` excludes `ladybug-shared`.

Occupancy comes from the graph registry under the **same status rule the allocator places against**, now extracted to a shared `OCCUPYING_DATABASE_STATUSES` constant instead of a set literal inside `_count_allocated_graphs`. A capacity *reading* on a different rule than the capacity *decision* disagrees with reality in whichever direction nobody is looking.

The alarm treats missing data as **breaching**: the emitter is scheduled, so no data means capacity is *unknown*, not fine. Every other capacity signal here failed by rendering silence as health. Large and xlarge publish but are not alarmed — they are request-only at desired 0, so 0 free slots is their correct resting state.

## Incidental

The graph-registry active-count scan is folded into the same pass that builds the occupancy map, and paginated — the previous `Select="COUNT"` form read `Count` off the first page only, silently undercounting past 1 MB.

## Tests

`TestAlarmDimensionsMatchEmitters` reads dimension names out of the emitter source with `ast` and checks every alarm in `graph-infra.yaml` and `graph-ladybug.yaml` against them. Verified it fails on the pre-fix state:

```
graph-infra.yaml DatabaseCapacityAlarm: alarm watches DatabaseUtilizationPercent on
[] (no dimensions), but the emitter publishes it only on [['NodeType']]
```

Plus seven `TestTenantSlotsFree` cases: per-tier dimensioning, full fleet reports 0, shared instances excluded, only occupying statuses consume a slot, and a registry scan failure publishes nothing rather than asserting unverified headroom.

One existing test changed meaning: `test_no_instances_no_metrics` asserted an empty fleet publishes nothing. It now asserts it publishes `TenantSlotsFree: 0` for each tenant tier — an empty fleet is one nothing can be placed on, which is exactly the state worth alarming on.

## Test plan

- [x] `just test-all` — 14,653 passed, 42 skipped, 2 xfailed; ruff/format/basedpyright/cf-lint clean
- [x] `get-metric-data` validation above, against the live namespace
- [x] Dimension guard confirmed failing on the pre-fix template
- [ ] After deploy: confirm `TenantSlotsFree` publishes for all three tiers and the Standard alarm transitions

Note: `graph-infra.yaml` is now 47,641 bytes against the 51,200 inline ceiling — ~3.5 KB headroom.

https://claude.ai/code/session_01KLwrRn9WdYxUNj6tx2XSFe


jfrench9:
@claude please review this PR

PR #1401, opened 2026-09-15
fix(ledger): reversal netting and schedule auto-reversal integrity

jfrench9:
## Summary

Reversing a journal entry netted to minus the entry instead of to zero, and an
`auto_reverse` schedule's reversing entry was deleted by the next period's run.
Both produce wrong books silently — every entry involved stays internally
balanced, so the trial balance foots and nothing raises a flag.

Three of the four P0 findings from the application-layer data-integrity review
(`specs/appsec-data-integrity-review-2026-09-14.md` and its validation
companion `specs/appsec-data-integrity-disposition-2026-09-14.md` in the
design vault). F11, the `(source_structure_id, period)` unique index, is not in
this PR: it needs a migration and duplicate cleanup ahead of the constraint.

## Changes

**Ledger reads — one owner for the landed-entry predicate**

- New `operations/roboledger/entry_status.py`. `reverse_journal_entry` posts a
  full offsetting entry *and* flips the original to `status='reversed'`, but
  every balance read filtered `status = 'posted'` — so the original dropped out
  of the sums while its offset, an ordinary posted row, stayed in. Reversing a
  $5,000 accrual reported the expense at -$5,000 and cash at +$5,000.
- The invariant was restated in twelve places and owned by none. It is now
  defined once and rendered three ways from that definition: a frozenset for
  Python membership, an expanding bind param for the `text()` reads, and a
  projection helper for the materializer.
- Applied across `reads/{trial_balance,account_rollups,taxonomies,closing_book}`
  and `reports/{fact_grid,rollforward_filters}`.
- `event_block/commands.py` imports it instead of restating it — it was already
  asking the same question, correctly, for its retraction fence.

**Graph projection and the AI-facing guidance**

- `extensions/materialize.py`: `Entry.is_live` and `LineItem.is_live` now
  project the landed predicate. **Takes effect on a graph's next
  materialization** — existing graphs keep the old flag until rebuilt.
- `schemas/extensions/roboledger.py`: the `is_live` property comments described
  the old rule and are corrected.
- The defect had been written into the instructions the AI follows. The MCP tool
  guidance (`middleware/mcp/tools/constants.py`, `example_queries_tool.py`) and
  the analyst operator prompt told callers `is_live ⇔ status = 'posted'` and
  that reversed entries would "inflate every balance" — steering the operator to
  compute the wrong answer. Corrected to describe the pair that nets to zero.

**Schedules**

- `schedules/service.py`: the reconcile that asks "does this period already have
  a closing entry?" keyed on structure + posting-date window with no type
  filter. An auto-reversal carries the same `source_structure_id` and posts on
  the first day of the *next* period, so next month's run found last month's
  reversal, judged it stale (its DR/CR are flipped by construction) and deleted
  it. `get_period_close_status` in the same file had always filtered the type —
  the asymmetry between the two queries was the defect.
- `event_block/promotion.py`: the stranded-obligation sweep shared that key and
  is fixed with it, so a genuinely stranded obligation no longer reads as
  already drafted.
- `_delete_draft_entry` now enforces the precondition its docstring claimed.
  "Must be status='draft'" was prose while every statement ran unconditionally,
  including a cascade into `WHERE reversal_of = :eid` — which targets a posted
  reversing entry by construction. Status is re-read under `FOR UPDATE`, a
  landed row raises, and every delete is bounded to drafts so the cascade cannot
  reach a posted row even if the guard were bypassed.
- The caller's guard refused only `posted`, letting `reversed` through into the
  regenerate path; it now refuses anything that landed. The truncate and rebuild
  guards move to the same predicate — both already said "posted (non-draft)" in
  their comments while testing `= 'posted'`.

**Worth a close look:** whether `status IN ('posted','reversed')` is right for
every one of the six read modules, and whether `is_live` widening is the
intended semantics for existing saved Cypher.

## Breaking Changes

None to the SDK contract — no GraphQL schema change, no operations-envelope
change, no REST shape change, so neither client needs a major.

One behavior change worth flagging for consumers rather than integrators:
`Entry.is_live` / `LineItem.is_live` in the materialized graph now mean
"landed" rather than "posted", so hand-written Cypher and saved queries that
filter on them will include a reversed original alongside its reversing entry —
which is the point, since summing only one half of that pair is what produced
the wrong balance. It applies to a graph at its next materialization.

## Testing

- `just test` — **14,664 passed**, 42 skipped, 2 xfailed, 0 failures.
- `just test-code` — ruff, format and basedpyright all clean.
- New `tests/operations/roboledger/reads/test_reversal_balance_db.py` asserts the
  **balance**, not the status, against a real Postgres schema. The existing
  suites checked that the original becomes `reversed` and that an offsetting
  entry is written — both true, and both ratifying the defect. I verified this
  test fails on the old predicate (`-5000.0`, with `total_debits ==
  total_credits`) and passes on the new one, rather than assuming it would.
- New `tests/operations/roboledger/test_entry_status_coverage.py` is structural:
  it fails on a *new* hardcoded predicate anywhere in the balance-read modules,
  which is the only way this stays owned. It caught a stale docstring on its
  first run.
- New cases in `tests/operations/roboledger/schedules/test_service.py` cover the
  reversing-entry exclusion on both queries and the delete guard's refusal of
  `posted` and `reversed`.

Not covered here: the data repair. Exposure counts (reversal pairs, schedules
missing reversals) are unknown, and the code fix has to land before any backfill
can be computed against it.

## Certification


jfrench9:
@claude please review this PR

PR #1402, opened 2026-09-15
fix(schedules): enforce one closing entry per period, lock the schedule row

jfrench9:
## Summary

Closes the last P0 item from the data-integrity review (F11) and the two the
companion rated P0 alongside it (F21.09, F21.04). Follows #1401, which fixed
F1/F2/F3.

The theme is the one the review kept finding: an invariant stated in application
code and owned by nothing. A schedule posts one closing entry per period, a
schedule row must not be read-then-written by two writers at once, and an entry's
`triggered_by_event_id` records the event that created it — all three were true
by convention and false under concurrency or correction.

## Changes

**F11 — one primary closing entry per (schedule, period)** · migration `0036`

- New partial unique index `uq_entries_one_primary_per_schedule_period` on
  `(source_structure_id, posting_date) WHERE source_structure_id IS NOT NULL AND
  reversal_of IS NULL`.
- The application check was a `SELECT ... LIMIT 1` with no lock. It holds today
  only because one caller reaches it under the obligation row lock; a second
  writer on a *different* event row (an operator or MCP close co-pilot firing
  `schedule_entry_due` during the Dagster sweep) passes the same check and
  inserts the twin. Close posts both, and the reconcile can't repair it —
  `ORDER BY created_at DESC LIMIT 1` only ever sees one.
- `posting_date` stands for the period: obligations are minted with
  `posting_date = period_end` and the handler passes it through, so both racers
  derive the same date. Two real columns, so no expression index and no "fiscal
  periods are calendar months" frozen into a migration.
- `reversal_of IS NULL` excludes the auto-reversal, keyed on the reversal *link*
  rather than entry type — `entry_type` is caller-authored and may itself be
  `"reversing"` (the regression caught in #1401's review).
- `create_manual_closing_entry` writes `source_structure_id=None`, so manual
  adjustments can never collide with scheduled ones.

**F21.09 — lock the schedule row**

- `create_closing_entry` used an unlocked `session.get(Structure, ...)`;
  `_load_schedule_or_404` (behind update / terminate / delete / rebuild) used a
  plain `scalar_one_or_none`. Both read the entry template and originator event
  id and write back decisions derived from them.
- `information_block/forecast.py` and `rollforward.py` already take `lock_by_id`
  on their `Structure` for exactly this shape; schedules were the outlier.
- Bounded wait — `create_closing_entry` runs on a request path as well as the
  sweep, and `RowLockedError` is retryable. The lock turns the loser into a clean
  conflict; the index makes the bad row impossible for any path that forgets it.

**F21.04 — creation provenance survives a reversal**

- The reversal handler re-pointed the *original* entry's
  `triggered_by_event_id` at the reversal event. `Entry` has one such column, so
  this overwrote the creating event rather than adding a link.
- Three readers treat it as creation provenance: `_assert_retractable` (the
  creating event became retractable while its posted entry stood), the QuickBooks
  full-rebuild wipe (a re-pointed original survives its own wipe as an orphan),
  and the graph projection (a March entry attributed to a September event).
- The reverse link needs no column — the reversing entry already carries
  `reversal_of` plus its own event id.

**Worth a close look:** the migration's tenant fan-out, and whether
`posting_date` is the right stand-in for "period" on every path that writes a
schedule-derived entry.

## Breaking Changes

None to the SDK contract — no GraphQL schema, operations-envelope or REST shape
change, so neither client needs a major.

Two operational notes:

- **This PR carries a migration** (`0036`, extensions DB). It creates the index in
  `public` and in every tenant schema via `for_each_tenant_schema`. Alembic
  autogenerate produced 25 spurious operations against the schema-per-graph
  layout — including dropping `fk_entries_triggered_by_event` — so the migration
  is hand-corrected to the single index, per the repo's "review the generated
  file" rule.
- **A concurrent duplicate now fails instead of silently duplicating.** That is
  the point, but it is a behavior change: the losing writer gets a conflict where
  it previously got a second entry that close would post.

## Testing

- `just test` — **14,673 passed**, 42 skipped, 2 xfailed, 0 failures.
- `just test-code` — ruff, format and basedpyright clean.
- Migration applied locally, verified present in `public` + 5 tenant schemas,
  then **downgrade/re-upgrade round-tripped** (6 → 0 → 6 schemas, head back at
  `0036`).
- **Zero conflicting rows in production** at exactly this key, checked across all
  five tenant schemas before writing the migration — so it deploys with no
  cleanup step.
- New DB-backed cases in `test_reconcile_key_db.py` assert the constraint rejects
  a second primary entry, permits the auto-reversal beside its original, permits
  two reversals sharing a date for different originals, and leaves manual entries
  alone. The fixture builds the schema from the model, so these also assert the
  model declaration and the migration describe the same constraint.
- `test_journal_entry_reversed.py` had a test named `..._links_both_entries`
  asserting two UPDATEs — it encoded the defect. It now asserts one.

## Certification


jfrench9:
@claude please review this PR

PR #1403, opened 2026-09-15
Index the foreign private issuers: xbrlkit 0.16.2 and the 20-F

jfrench9:
The robosystems half of `specs/adapters/filing-section-coverage.md` gap 1. The xbrlkit half is [RoboFinSystems/xbrlkit#64](https://github.com/RoboFinSystems/xbrlkit/pull/64), released as 0.16.2.

## What changes

**`xbrlkit>=0.16.0` → `>=0.16.2`**, relocked. The floor moves rather than resting on the existing `<0.17.0` range because the config change below *depends* on the new behaviour: on 0.16.0, `form_types` including 20-F reads a 20-F through the **10-K** map, where Item 4 is mine safety rather than the business. The pin is what stops that combination from resolving.

**`form_types` gains `20-F` on both text index assets** — `SECNarrativeIndexConfig` and `SECiXBRLIndexConfig`. Until now both defaulted to `["10-K", "10-Q"]`, so every foreign private issuer's annual report was absent from search: **3,011 20-F reports on the shared graph**, beside 420 40-Fs.

**40-F stays out**, deliberately. It is an MJDS wrapper that files its substance as the Canadian AIF and MD&A exhibits, so 0.16.2 extracts no sections from one by design — indexing the form would run the pass for nothing. A test holds that, with the reason.

Verified the resolved package is what the config assumes:

```
xbrlkit 0.16.2
20-F map: ['3', '4', '5', '11', '15', '16K']
10-K map: ['1', '1A', '1C', '2', '3', '5', '7', '7A', '9A', '11', '13']
```

`text_index.py:564` passes the filing's own form straight to the extractor, so a 20-F routes to its own map with no further change. No job config or sensor overrides `form_types` on either index asset — checked.

## This does not change the index

These defaults decide what the **next** pass covers; existing documents stay as they are until a re-index runs. That pass is `runbooks/sec-text-reindex.md`, it is due on its own account (`last_run: 2026-09-05`), and it delivers three things at once:

- **gap 0** — the boundary fix, merged and released in 0.16.x but never delivered; prod still opens every 3M `item_7` on table-of-contents rows, and `section=item_3, form_type=10-Q` returns the index's 10,000-hit ceiling with every hit mislabelled Market Risk
- **gap 1** — this, plus the 10-K items #64 adds (3, 5, 9A, and Part III where it is inline)
- **gap 2** — the 50,000-character truncation, replaced by parts

It is a text-axis run and independent of the graph rebuild. Per §2 of the runbook, `skip_embeddings` must be settled before the first quarter runs: **already decided 2026-09-13, SEC is BM25-only**, so there is nothing to decide inside the run.

## One thing to know before the pass

#64 skips an item that points at its content rather than carrying it — Part III boilerplate, and bare cross-references like 3M's Item 3 → Note 17. **That rule is not limited to the new items**, so it also drops pre-existing ones: Pfizer's `item_7a` is 42 words of "incorporated by reference to … MD&A" and is indexed today. After the re-index it will not be. That is the intended behaviour and the content is indexed elsewhere (the note as a tagged text block, the MD&A as its own section), but it is a removal from the corpus, not only an addition.

14,678 passed, 0 type errors.


jfrench9:
@claude please review this PR

PR #1405, opened 2026-09-16
feat(operators): port the AI client to Bedrock Converse and register models by profile

jfrench9:
## Summary

Ports the AI Operator client from Bedrock `InvokeModel` with an Anthropic-shaped body to the Bedrock **Converse** API, and replaces the hard-coded Claude enum with a **model registry** keyed by profile (`economy` / `balanced` / `quality`). Every model in the Bedrock catalog now runs through one request shape, so which model an operator uses is a registry row rather than a code path. The token meter resolves its rate through the same registry and **raises on an unregistered model** instead of silently billing at Sonnet rates. Steps 1–3 of `specs/ai-operators/llm-provider-abstraction.md`; no customer-facing surface changes.

## Changes

**Rate card and meter**
- `config/billing/ai.py`: adds `anthropic_claude_5_opus` (5.5 / 27.5 credits per 1K, cache 0.55 / 6.875) and `openai_gpt_5_6_luna` (0.22 / 1.32, cache 0.022 / 0.275), read from the Bedrock agreement rate cards and the model card. The docstring now points at the registry as the owner of model → key.
- `operations/graph/credit_service.py`: `consume_ai_tokens` resolves the pricing key via `OperatorConfig.pricing_key_for(model)`. The hand-maintained map (and its silent default to Sonnet 4 rates, which contradicted the rate card's own docstring) is gone; an unknown model raises `ValueError`, which the tracked client latches as an unbilled call.

**Model registry** (`config/operators.py`, exports in `config/__init__.py`)
- `ModelSpec` rows carry the wire id, `pricing_key`, `cache_points`, `accepts_sampling_params`, `additional_request_fields`, `max_output_tokens`, `cross_region`. Behaviour is data, not string matching: `model_accepts_sampling_params` and `BEDROCK_MODELS` are removed.
- `ModelProfile` + `PROFILE_MODELS`: `economy` → `us.openai.gpt-5.6-luna`, `balanced` → `us.anthropic.claude-sonnet-5` (the default, unchanged), `quality` → `us.anthropic.claude-opus-5`.
- `resolve_model()` is most-specific-wins: explicit per-call model or profile → `OPERATOR_MODEL_OVERRIDES` (still empty) → `DEFAULT_MODEL_CONFIG.default_profile`. An unregistered name raises. `validate_configuration()` now checks every row against the rate card and every profile against the registry; `config/validation.py`'s startup summary reads the new fields.

**Converse client** (`operations/operators/ai_client.py`)
- `invoke_model` + `anthropic_version` body → `client.converse(...)`: `inferenceConfig`, `system` with a trailing `cachePoint`, `toolConfig` built from MCP-shaped tool definitions, `additionalModelRequestFields` from the registry (Claude 5: thinking disabled, as before). `cache_control` → `cachePoint`; the trailing-turn breakpoint keeps its semantics. Usage is read from `inputTokens` / `outputTokens` / `cacheReadInputTokens` / `cacheWriteInputTokens`.
- New typed `AIProviderError` for entitlement, credential and request-shape failures (`AccessDenied`, `ResourceNotFound`, `Validation`, expired/unrecognized credentials); throttling and 5xx keep their botocore type. `AIResponse.tool_calls` and the `tool_result_block` / `text_block` helpers keep the provider shape in this one file.
- `create_message` no longer takes `tool_choice`: Converse cannot forbid tool use, and nothing else used it. Reviewers: this is the one behavioural change worth a close look, in `tool_loop.py` below.

**Tool loop and access** (`tool_loop.py`, `tool_access.py`, `tracked_ai.py`, `operator_context.py`)
- The loop builds Converse `toolUse` / `toolResult` blocks (`status: "error"` replaces `is_error`) and replays the assistant turn verbatim, reasoning blocks included.
- The wrap-up turn keeps `toolConfig` (Converse rejects a transcript carrying tool blocks without it), forbids further calls in the nudge text, and treats a stray tool call as terminal: answer text if any, else the existing fallback sentence. Previously this was enforced by `tool_choice: none`.
- `get_tool_schemas()` returns MCP definitions unchanged (`inputSchema`); the client wraps them into `toolSpec`.

**Mapping operator** (`implementations/mapping/operator.py`)
- The per-batch `except Exception` re-raises `AIProviderError` first, so a provider outage fails the operation instead of reporting a successful run with every batch "skipped".

**Docs**: `config/README.md` and `operations/operators/README.md` describe the registry, profiles and resolution order.

## Breaking Changes

None for the published SDKs: no OpenAPI, GraphQL, or operations-envelope change; `OperatorRequest` is untouched. Internal only: `AIClient.create_message` loses `tool_choice`; `config.model_accepts_sampling_params` and `OperatorConfig.BEDROCK_MODELS` are removed; `DEFAULT_MODEL_CONFIG.default_model` is now `default_profile`. All callers are in this repo and updated here.

## Testing

- `just test-code` (ruff, format, basedpyright, cf-lint): passed.
- Targeted suites — `tests/config`, `tests/operations/operators`, `tests/operations/test_ai_client.py`, `tests/operations/test_credit_flow.py`, `tests/middleware/mcp/test_registrar.py`: 850 passed. The full `just test-all` was not run in this session.
- Live on the local stack (analyst operator, standard mode, demo ledger, default profile): the cold run wrote the ~34.6K-token prefix and each later call read it back; a warm run with a different question three minutes later read ~33.9K cached tokens on its first call. Transaction metadata carries the wire id and all four token classes at the Sonnet 5 rates. The error-feedback path fired once (a Cypher binder error came back as an error-status tool result and the model corrected).
- Raw Converse probes against the account: Sonnet 5, Opus 5 and GPT-5.6 Luna each complete a tool-use round trip; Luna rejects `temperature` and explicit `cachePoint` blocks but reports implicit cache reads in usage, which is why its registry row has `cache_points=False`.

## Certification


jfrench9:
@claude please review this PR

PR #1406, opened 2026-09-16
fix(serialization): the holon stops calling every tenant report an SEC filing

jfrench9:
## Summary

A tenant report's holon was asserting it was an SEC as-filed filing. xbrlkit's graph writer defaults an unset `reporting_style` to `"sec-as-filed"`, and the bundle→model bridge never set the field — so the unset value did not omit the claim, it published a wrong one. The two downloadable flavors of one report disagreed: the flat JSON-LD carried the real style while the holon beside it said `sec-as-filed`.

## Changes

**`robosystems/operations/serialization/model.py`** — `_filing` now passes `bundle.reporting_style` into `FilingMeta`, the same value the flat encoder has always stamped, so the two projections agree by construction. The docstring records why the field cannot be left to default.

**How it got here, for reviewers.** The holon half of the serialization waist (#1373) moved the encoder from platform code to xbrlkit without `bundle_to_xbrl_model` filling this field — which the preceding xbrlkit change had added to `FilingMeta` for exactly this purpose. Before that move the platform's own encoder wrote no `reportingStyle` at all, so the value went from *absent* to *wrong*.

**Tests — three layers, because no single one would have caught it:**

- `test_model_bridge` — the bridge sets the field.
- `test_holon_alignment` — the two projections agree in the emitted output. This is the file whose premise is that the flat form and the holon name one report the same way, and the style is one claim about that report.
- `test_sample_artifact_invariants` — the committed demo artifacts obey the same rule. This file exists (#1393) for precisely this class of drift going unnoticed in checked-in output.

**Demo artifacts** — all five demo artifact sets regenerated, so they demonstrate current behavior rather than pinning the old. The diff is large but mechanical: each set's node-type profile and label set is byte-for-byte identical to the previous commit, so the churn is report-id IRIs plus this fix. SHACL conforms and Arelle validates on every regenerated set.

**Residual, not addressed here:** derived holons are cached per report generation, so a holon already materialized for an existing generation keeps the old value until that generation is re-materialized. New generations are correct from this commit. The SEC public artifacts are unaffected — they are written from xbrlkit's parse of an actual filing, where `sec-as-filed` is true.

## Breaking Changes

None.

No SDK impact. This changes the *content* of one downloadable artifact, not any REST shape, GraphQL schema, operations envelope, or generated symbol — neither the stable nor the generated tier of the client contracts is touched. The affected predicate (`rs:reportingStyle`) is already published and keeps its name and cardinality; only the value becomes correct.

## Testing

- `just test` — 14698 passed, 42 skipped, 2 xfailed. The two xfails are the pre-existing concept-IRI pair (#1397), unchanged by this branch.
- Verified end to end against the running local stack rather than in unit tests alone: republished a report on a live demo graph and re-downloaded every flavor — the holon and the flat JSON-LD now carry the same style.
- Confirmed the tests fail without the fix: reverting the one-line change turns `test_holon_alignment` red.

## Certification


jfrench9:
@claude please review this PR

PR #1408, opened 2026-09-16
fix(admin): a running SSM command is not a failed one

jfrench9:
## What

`SSMExecutor.execute()` reported a command that was **still running** as one that had **failed**, with `Command failed with exit code -1`.

## Why it happens

`aws ssm wait command-executed` is botocore's `CommandExecuted` waiter — `delay: 5, maxAttempts: 20` — so it gives up after **~100 seconds** regardless of the executor's `timeout` (default 300s), and neither number is configurable. `execute()` then fetched the invocation and read `ResponseCode`, which an **in-flight** invocation does not carry, defaulted it to `-1`, and raised.

The docstring already stated the intent — *"A wait timeout is not fatal"* — but `data.get("ResponseCode", -1)` quietly made it fatal anyway.

## Why it's worth fixing rather than living with

The same executor runs `search delete` against the shared OpenSearch `documents` index. A delete over ~1.4M documents reliably outlives a 100s wait **while succeeding**. Reporting that as a failure invites the natural response — re-running it — on a destructive, non-idempotent command.

Hit live on 2026-09-16 by `search force-merge`: failed at exactly 100s while `aws ssm list-commands` showed `InProgress`, and the merge then ran on for another 20 minutes.

## What changed

- Poll `get-command-invocation` for the remainder of the real budget instead of treating the waiter's return as the verdict. A command that finishes at 150s now succeeds normally.
- When it genuinely has not finished, raise **`SSMCommandStillRunning`** — a `RuntimeError` subclass, so existing `except RuntimeError` handlers are unaffected — carrying `command_id`, `instance_id`, `status`, and the exact CLI call to check it by hand.
- Terminal failures are unchanged: still a plain `RuntimeError`, now with the SSM status alongside the exit code.
- The "may have timed out (waited 300s)" message was wrong — it had waited 100. Reworded.

## Tests

`tests/admin/test_ssm_executor.py`, 6 cases: in-flight raises the new type rather than a generic failure; the message names the command and warns off re-running; it subclasses `RuntimeError`; a command finishing after the waiter gave up still succeeds; terminal failure still raises plainly; the fast path doesn't sleep.

Mutation-checked — emptying `_IN_FLIGHT_STATUSES` (restoring the old behaviour) fails the first test, and restoring it passes all six.


jfrench9:
@claude please review this PR

PR #1409, opened 2026-09-16
chore(deploy): make the writer-tier AZ a guarded deploy parameter

jfrench9:
## Summary

Every LadybugDB writer tier is pinned to one availability zone (EBS volume affinity), and that zone was a hardcoded literal in `deploy-graph-ladybug.yml`, recomputed on every release. Relocating a tier out of band — after an AZ outage, or when the default zone has no instance capacity — was undone by the next input-less dispatch from `create-release` / `graph-maintenance`, and a `VPCZoneIdentifier` change replaces every instance in the ASG, each relaunching on an empty volume because the volume registry is AZ-scoped. This PR makes the zone a deploy parameter with two levers and a guard, so a relocation survives later releases and a release can never relocate the fleet by accident. With nothing set, behaviour is unchanged (`<region>c`).

## Changes

- **`.github/workflows/deploy-graph-ladybug.yml`** — new `workflow_call` inputs `writer_az` (default `""` → `<aws_region>c`) and `writer_az_change_allowed` (default `"false"`). The *Select Availability Zone* step now:
  - validates the zone against the region before any AWS call;
  - errors, instead of silently falling back to the first subnet, when an **explicitly requested** zone has no private subnet — the fallback is kept for the default zone only;
  - reads the existing stack's `SubnetIds` parameter and **fails the deploy when it differs from the resolved subnet unless `writer_az_change_allowed` is `true`**; when a move is allowed it emits a `::warning::` naming the follow-ups (set the sustain variable, repoint graph-registry rows, restore);
  - writes `tier | AZ | subnet | create/update` to the step summary — `selected_az` was previously computed and read by nothing.
  - The dead `use_single_subnet` output is dropped. The deploy step is untouched; it already consumed `selected_subnet`.
- **`.github/workflows/deploy-graph.yml`** — declares the two inputs and threads them to the writer matrix job (every tier, including `ladybug-shared`). The replica job is untouched (multi-AZ by design).
- **`.github/workflows/prod.yml`, `staging.yml`** — new `workflow_dispatch` input `graph_writer_az`. Resolution: `writer_az: inputs.graph_writer_az || vars.GRAPH_WRITER_AZ_<ENV> || ''`; `writer_az_change_allowed` is `true` only when the dispatch input is set. `create-release.yml` / `graph-maintenance.yml` are unchanged — their input-less dispatches are the sustain path and cannot move a tier by construction.
- **`bin/setup/gha.sh`** — comment only: `GRAPH_WRITER_AZ_*` is deliberately not seeded (the `SHARED_REPLICA_IMAGE_TAG_*` precedent), so `just setup-gha` cannot move the tiers back after a recovery.

Reviewers: the guard block in the AZ step is the part worth reading closely — it must refuse the variable-driven mismatch and permit the input-driven one, and both paths must resolve to `<region>c` when nothing is set. One value applies to all writer tiers; per-tier placement is a follow-up.

## Breaking Changes

None. No API, GraphQL, operations-envelope, or SDK surface is touched; the change is confined to deploy workflows and a setup-script comment. With no input and no variable, every deploy produces the same stack parameters as before.

## Testing

- `just lint-actions` (actionlint) — passed. The shellcheck-integrated `just lint-actions-shell` reports only `SC2086` info items on `>> $GITHUB_OUTPUT`, the same style the surrounding steps use; it is not part of the gate.
- Pre-commit hooks passed on the commit.
- `just test-all` not run — the change is workflow YAML and a shell comment; no Python is touched.
- Not yet exercised against AWS. Regression check at the default: the next staging deploy should log `Preferred AZ for <tier> tier: us-east-1c (explicit: false)` for each tier and `No updates are to be performed` on each writer stack. Proving the levers on staging (writers at desired 0): dispatch with `graph_writer_az=us-east-1d` → warning, stacks move; dispatch with no input → the guard fails each writer tier; set `GRAPH_WRITER_AZ_STAGING=us-east-1d` → passes; dispatch back with `us-east-1c` and clear the variable.

## Certification


jfrench9:
@claude please review this PR

PR #1410, opened 2026-09-16
feat(search): fp16 vector quantization via a rebuildable index

jfrench9:
## What

The prod OpenSearch node holds ~1.43M SEC vectors at float32, half its k-NN budget and growing with every filing. fp16 scalar quantization halves that. A `knn_vector` encoder is a mapping property and cannot change in place, so the index has to be recreated — and rather than a versioned index and a swap, it is **deleted and rebuilt from source**: SEC content through the existing text-index pipeline, user graphs' documents from PostgreSQL, which `models/core/document/document.py` already declares as their source of truth.

This PR ships the three pieces that needs. It does **not** run the rebuild; that is an ops procedure, documented separately, with a prod pre-flight that deletes nothing (below).

### 1. Mapping — `operations/search/client.py`
`INDEX_MAPPING` gains the Faiss SQ encoder, `fp16`, `clip: false`. bge-small vectors are normalized to [-1, 1], far inside fp16's range, so an out-of-range value is a bug to reject rather than round away. New indexes inherit it from `create_index_if_not_exists()`.

### 2. `rebuild_documents_job` — new `dagster/jobs/documents.py`
Manually triggered. Walks every `Document` row through **`DocumentService.resync_document`** — the row→index primitive extracted from `create_document` / `update_document`, which now share it. `graph_id` run config for one graph; every graph that has rows when unset (`Document.graph_ids`). A failing row is logged, counted and skipped, never fatal: it is still in PostgreSQL, so the recovery is a re-run for its graph, and one bad document does not strand every other tenant's search.

It is **idempotent on a live index** — `upload_document` deletes by document prefix and rewrites the same stable ids — which is what makes the pre-flight below possible.

### 3. `search recreate-index` — `admin/commands/search.py`
DELETE the index and PUT the local checkout's `INDEX_MAPPING` **in the same bastion script**. `action.auto_create_index` is on, so an index left missing between two commands would be recreated by the next writer with a dynamic mapping and no `knn_vector` field. Dry-run first with a per-graph count of what is about to go; confirmation is the environment name typed back (`--force` to skip); the output reads the encoder back from the domain so the result is checked, not assumed. The mapping comes from the local checkout — the reason this needs no deploy, and the reason it must be run from the merged branch.

### Also
`README.md` names [xbrlkit](https://xbrlkit.com) as the extraction library behind the SEC repository, and drops `XBRL text blocks` from the search bullet — that source type left the index some time ago.

## Verification

- 4,804 unit tests pass; `just test-code` clean (ruff, format, basedpyright, cf-lint).
- Three mutations caught: dropping `clip` from the mapping, removing the PUT from the recreate script, and making a failing row abort the job.
- **Live e2e against local OpenSearch 2.19.5 — prod's version.** Deleted the local `documents` index, restarted; the first `index-document` call recreated it with `{"encoder": {"name": "sq", "parameters": {"type": "fp16", "clip": false}}}`. Then wiped a demo graph's sections from the index and launched `rebuild_documents_job` for it: **10 documents / 50 sections** restored from PostgreSQL — including nine documents that had only ever existed in the deleted index. Embeddings present.

## Operational note

There is no staging OpenSearch, so the rebuild runs in prod with a pre-flight that deletes nothing: deploy, then run `rebuild_documents_job` against the **live** index and reconcile per-graph counts with PostgreSQL — if that fails, nothing has been lost. Only then: a `recall@k` baseline on the fp32 index the same day, record per-graph counts, `recreate-index`, `rebuild_documents_job`, then the SEC lanes newest-quarter-first. SEC search is partial for the 3–4 days the quarters take to land; user-graph search is back within the first hour.


jfrench9:
@claude please review this PR

PR #1412, opened 2026-09-16
fix(admin): recreate-index proves both verbs on a probe index before the delete

jfrench9:
## What happened

The first live `search recreate-index` on prod (v1.12.5, the fp16 rebuild) was refused with **HTTP 403 on the DELETE**. `cloudformation/bastion.yaml` grants the bastion role `es:ESHttpGet / Post / Head` only. The index was untouched.

It was untouched *only because* the delete was refused: the role lacks `es:ESHttpPut` as well, so the create on the next line would have failed after the delete — leaving no index and no mapping, for the next writer to auto-create over with a dynamic mapping. That is precisely the hazard the DELETE-and-PUT-in-one-script design was built against, and the design never checked that its executing principal could perform either verb.

## Fix

- **`cloudformation/bastion.yaml`** — the bastion role gains `es:ESHttpPut` and `es:ESHttpDelete`, scoped to the same domain resource. The Dagster task role already carries both; the API and worker roles are unchanged.
- **`admin/commands/search.py`** — before touching `documents`, the script PUTs the mapping to `documents-recreate-probe` and DELETEs it. A verb the role lacks, or a mapping the engine rejects, now fails on the probe with the live index still standing — never after it is gone.
- **Tests** — the full call order (probe DELETE/PUT/DELETE → live DELETE/PUT → GET mapping), the probe carrying the real mapping, and a 403 on either verb raising before any call reaches `/documents`.

## Verification

`cfn-lint` clean; 23 admin-script tests pass; ruff, format, basedpyright clean. The bastion stack ships through `prod.yml` (`deploy-bastion.yml` is `workflow_call`-only), so this needs a release and prod deploy before `recreate-index` is retried.


jfrench9:
@claude please review this PR

PR #1414, opened 2026-09-16
chore(xbrlkit): bump to 0.17.1 and page information-block like xbrlkit serve

jfrench9:
## Summary

Bumps xbrlkit from 0.16.3 to 0.17.1 and brings the platform's uses in line with it. The hosted `information-block` runs xbrlkit's own `information_block` over a report the platform holds, so the REST op and the MCP tool now take the same `offset` paging `xbrlkit serve` gained in 0.17.0. The view also imports from `xbrlkit.serve`'s declared surface rather than its internal modules.

## Changes

**Dependency**
- `pyproject.toml`: `xbrlkit>=0.17.1,<0.18.0` (was `>=0.16.2,<0.17.0`). `uv.lock` moves xbrlkit 0.16.3 → 0.17.1 and nothing else.
- Between 0.16.3 and 0.17.1 only xbrlkit's `serve/` and `view.py` changed: loading a taxonomy package, paging `statement` / `information_block`, and `resolve_element` ranking. The SEC adapter's parse, serialize and EDGAR paths are unchanged. The cached-model format did not change either, so `MODEL_CACHE_VERSION` stays at 3.

**Declared surface** (`operations/roboledger/views/information_blocks.py`)
- Imports `LoadedFiling`, `ToolError`, `disclosures`, `information_block`, `MAX_BLOCK_ROWS` and `MAX_BLOCK_MEMBERS_CAP` from `xbrlkit.serve`, the names xbrlkit pins in its `test_serve_surface.py` for this host. It no longer imports from `xbrlkit.serve.session` or `xbrlkit.serve.tools`.

**`information-block` paging**
- `query_information_block` takes `offset` and passes it to xbrlkit.
- REST: `InformationBlockRequest.offset` (optional, ≥ 0). `InformationBlockResponse` declares `offset`, `ancestors` and `next_offset`; `extra="allow"` already carried them through.
- MCP: an `offset` property on the `information-block` tool, validated as a whole number (a negative value reads the first page), plus one line each in PARAMETERS and RETURNS.
- Behavior: a block longer than `max_rows` is `truncated` with `next_offset`. Passing it back returns the next page, and a later page carries `ancestors` and leaves the section's axes, calculation and text to the first page.

## Breaking Changes

None. The request gains an optional field and the response gains three optional fields, so this is additive in the SDKs' generated tier and rides a client minor on the next regeneration. Calls without `offset` behave as before.

## Testing

`just test-all`: 14,751 passed, 42 skipped, 2 xfailed; ruff, format and basedpyright clean.

- New: `test_a_block_longer_than_max_rows_pages` runs xbrlkit 0.17.1 unmocked over the published-holon fixture. The pages join into the unpaged rows, the later page carries `ancestors`, and the calculation comes only with the first page.
- New: `test_offset_defaults_to_the_first_page_and_must_be_a_number` covers the MCP tool.
- Updated: the router and MCP passthrough tests assert `offset` reaches `query_information_block`, and the caps test reads `xbrlkit.serve`'s constants.


jfrench9:
@claude please review this PR

PR #1415, opened 2026-09-17
feat(plaid): the Plaid bank feed on a shared bank-feed lane

jfrench9:
## Summary

Adds Plaid as a second bank feed. Everything Mercury built for bank feeds is source-agnostic, so this PR first moves those shared pieces out of the Mercury adapter into `adapters/bank_feed/`, then builds the Plaid provider and pipeline on top of them. The feed is gated by `CONNECTION_PLAID_ENABLED`, which defaults to off.

## Changes

**1. The shared bank-feed code (`refactor`, no behavior change)**
- `adapters/bank_feed/` now holds the pieces both feeds use:
  - `hints.py`: the suggestion vocabulary.
  - `chart.py`: the pure `BankAccount` and `ChartIndex`.
  - `accounts.py`: link-or-create, now taking a `provider`, plus `build_chart_index`. A chart account another connection feeds is never claimed by name: the same bank account connected through two feeds gets a second chart account instead of the two feeds overwriting each other's link.
  - `load.py`: counterparties, capture through the event-block kernel, the hint refresh.
  - `sync.py`: a shared `BankFeedSyncConfig` and the sync bookkeeping (failure record, lock release, `last_sync`, fiscal-calendar bootstrap, staleness mark).
- `operations/providers/bank_feed.py` now holds the consent record and the disconnect purge, stamped with the provider.
- Mercury keeps its client, its category tables and its transform. `BankAccount.mercury_id` became `account_id`, and an `institution` field was added.
- The Mercury tests for the moved code moved to `tests/adapters/bank_feed/` and `tests/operations/providers/test_bank_feed.py`.

**2. Suggestions resolve on template charts (`feat`, affects Mercury too)**
- `AccountHint` gained `aliases`: the names the same account has on the shipped saas, services and product templates. `ChartIndex.resolve_hint` tries the hint's own name first, then its aliases.
- On a fresh services-template chart, none of a first sync's 129 suggestions resolved; after this change, 117 did. Where a template has no fitting account, the suggestion stays name-only, as before.

**3. The Plaid feed (`feat`)**
- **Client** (`adapters/plaid/client/api.py`): an httpx client with no new dependency. It covers Link tokens (new Item or update mode), the public-token exchange, accounts, the `/transactions/sync` cursor loop, Item removal, and sandbox public tokens.
  - A mutation during pagination restarts from the original cursor.
  - 429 and 5xx responses are retried inside the call.
  - A broken login or a removed Item is recognized as needing the customer back in Link.
- **Connect flow**, reusing the OAuth endpoints (`routers/graphs/connections/oauth.py`):
  - `oauth/init` returns a `link_token` in place of an `auth_url`, and the state token lives 30 minutes for Plaid.
  - `oauth/callback/plaid` takes Link's `public_token` as `code`.
  - A connection that already holds an Item gets update mode. An Item Plaid no longer knows falls back to a fresh Link.
  - A bank login already connected to the graph (same institution and a shared account, matched by mask and subtype, or by name and subtype when the institution sends no mask) is refused with `409 DUPLICATE_BANK_CONNECTION`. The new Item is removed at Plaid, best-effort (a removal error doesn't change the 409), and the pending row withdrawn.
- **Provider** (`operations/providers/plaid_provider.py`): one connection per Item, so a graph can hold several. A second create returns the connection still waiting on Link. The Item's access token is stored in `ConnectionCredentials`. Disconnect removes the Item at Plaid, then runs `purge_bank_feed`.
- **Pipeline** (`adapters/plaid/pipeline/`), the `plaid_feed` asset in job `plaid_sync`:
  - The cursor advances only after the load commits. A full rebuild, or an explicit `since_date`, replays the Item's history.
  - Only cash and card accounts are booked.
  - Pending transactions are deferred until they post.
  - A transfer out and a transfer in of the same amount, on two of the Item's accounts within 3 days, become one `internal_transfer`. That includes a leg that posts on a later sync, merged with its waiting partner inside a savepoint.
  - Removed lines: unposted events are deleted; posted ones get `payload_drift` with `source_removed`, so they surface as reconciling items.
  - Modified lines: an unposted event takes the new amount and date in place; a classified one keeps its hints. A posted event whose amount or date changed becomes a reconciling item. Its stashed payload carries the entry the line should now have, rebuilt from its classification, so the plan shows the difference rather than a reversal, plus a `source_amount` so a resolved change isn't raised again.
  - Suggestions come from Plaid's personal-finance category, v2 (`tier0.py`).
- **Retries:** the job sets `dagster/retry_on_asset_or_op_failure: false`, so a crashed run worker is retried but a failed body (a dead login) is not. `Failure(allow_retries=False)` does not stop run-level retries; this was seen on a live run.
- **Wiring:**
  - `plaid` added to `ProviderType`, the provider registry, `BANK_FEED_PROVIDERS` (so the chart-required and no-QuickBooks guard applies), the options catalogue (`auth_type: link`) and the Dagster definitions.
  - The purge's scrub keys now include Plaid's payload keys.
  - The MCP inbox instructions now name both feeds.
  - Config: `PLAID_CLIENT_ID` / `PLAID_SECRET` / `PLAID_ENVIRONMENT` in the secrets mapping and validation, plus `.env.example`, `bin/setup/aws.sh` and the setup README.

**Worth a close look:** `adapters/plaid/pipeline/load.py`: removals, `reconcile_existing` and `_flag_changed` (the reconciling-item payloads), and cross-sync transfer merges.

**Not in this PR:**
- The app's Link widget.
- Redirect handling for banks that run their own OAuth inside Link.
- Plaid webhooks.
- Matching transfers across two different connections: each side still lands as its own external transfer.
- Detecting one bank account connected through two feeds: it now lands in two chart accounts, visibly, rather than being merged.
- A line posted through a tenant rule rather than a classification: an amount change on it plans as a reversal, because there's no classification to rebuild the entry from.

## Breaking Changes

None.

SDK notes (generated tier, client minor):
- `OAuthInitResponse.auth_url` changes from required to optional. It is still set for QuickBooks and Mercury.
- Added: `OAuthInitResponse.link_token`, `CreateConnectionRequest.plaid_config`, the `PlaidConnectionConfig` model, and `plaid` in `ProviderType`.
- None of these touch the SDK facades or the integration template's emit path.
- The TypeScript client needs a regeneration before the app can build the Link flow.

## Testing

- `just test-all` on the final tree: 14,883 passed, 42 skipped; lint, format and basedpyright clean. The pre-commit gate passed on each commit.
- New unit tests: the Plaid client, category table, transform (sign flip, cents rounding, transfer pairing), load (removals, refresh versus source change, merge and its rollback), asset (cursor, window, retry tags, failure paths), provider, Link routes and models. Plus shared-lane tests for aliases and provider-scoped account links.
- End to end on the local stack against the Plaid sandbox:
  - The chart-required guard.
  - Create, init, a sandbox public token, then the callback.
  - First sync: 153 captured, 2 chart accounts created, pending lines deferred, loan-account lines excluded.
  - A full-rebuild replay: no duplicates.
  - An unclassified commit refused; accept-suggestion then commit posting a balanced draft.
  - An incremental sync after `/sandbox/transactions/create`.
  - Three Items on one graph, and the duplicate-login refusal.
  - Disconnect: unposted lines deleted, the posted line kept with Plaid keys scrubbed, links cleared, the Item removed.
  - `/sandbox/item/reset_login` giving `needs_reauth` plus an update-mode Link token, with no Dagster retry after the tag change.
- Reconciling items, checked on the local ledger against the real planner and resolver inside a transaction that was rolled back:
  - A $134 → $150 change on a committed Plaid line planned as Travel & Entertainment +1,600 and checking −1,600. `catch_up` resolved it, and the same change on the next sync was not raised again.
  - A removal planned as the full reversal, and `catch_up` resolved it.
  - Existing planner behavior, not changed here: the default disposition is `restate` even while the line's entry is still a draft, and restate then refuses; `catch_up` works.
- Not verifiable in the sandbox: a real transfer pair (the sandbox legs don't match), a bank removing a posted line, and Link's UI.

## Certification


jfrench9:
@claude please review this PR

PR #1416, opened 2026-09-17
feat(mcp): add /v1/mcp/roboledger, a RoboLedger-scoped OAuth MCP endpoint

jfrench9:
## Summary

Adds `POST /v1/mcp/roboledger`, a graph-agnostic, OAuth-only MCP endpoint for RoboLedger graphs. A ChatGPT plugin listing freezes one tool list per URL, and OpenAI allows one MCP URL per plugin in an organization, so a RoboLedger plugin that lists the accounting workflow needs a URL of its own. `/v1/mcp` and the per-graph route are unchanged apart from the challenge fix below.

## Changes

**A third OAuth protected resource** (`operations/oauth_server/resources.py`)
- `ResourceTarget` gains a `product`; `roboledger_target()` and `route_target(graph_id, product)` pick the resource a route serves. `resolve_resource` accepts `/v1/mcp/roboledger` (trailing slash tolerated; query strings and unknown sub-paths still rejected).
- RFC 9728 document at `/.well-known/oauth-protected-resource/v1/mcp/roboledger` (`resource_name: "RoboLedger MCP"`).
- Tokens are audience-bound as before, so a `/v1/mcp` or per-graph token is refused here and a RoboLedger token is refused on the other two.

**Consent gate** (`operations/oauth_server/authorization.py`)
- `record_decision` refuses a grant on a product resource unless `graph_serves_product(graph_id, product)`: not a subgraph, not a shared repository (by id, flag or `graph_type`), and provisioned for the product. The extension alone is not the test, because the SEC manifest declares `roboledger` and subgraphs inherit their parent's extensions.
- Runs after the existing access check, so the refusal never describes a graph the user cannot read.
- The pending-authorization payload now carries `product`, so the consent page can filter its graph picker. That endpoint is schema-excluded.

**Bearer validation** (`middleware/auth/dependencies.py`)
- `get_oauth_roboledger_mcp_principal` shares one implementation with `get_oauth_mcp_principal`; the challenge and audience helpers take the route's `product`.
- **Fix, reviewers please look:** on the graph-agnostic route, a 403 for a revoked graph membership named the grant graph's *per-graph* metadata document in `WWW-Authenticate` instead of the route's own. The challenge now names the route's resource (`route_graph_id`), while the access check still runs against the grant's graph.

**Transport and tool profile** (`routers/graphs/mcp/remote.py`, `middleware/mcp/tools/manager.py`)
- `dispatch_jsonrpc` takes `excluded_tools`: withheld from `tools/list` and from the list `initialize` builds its instructions from, and refused on `tools/call` before the authorization gauntlet (a client can call a name it never listed). Default is empty, so existing routes behave as before.
- `ROBOLEDGER_ROUTE_TOOL_EXCLUSIONS` (12): the three subgraph-only tools (`write-graph-cypher`, `add-node-table`, `add-relationship-table`, which on a parent graph only answer `subgraph_required` and point at a connector a chat user cannot add), workspace administration (`create-subgraph`, `delete-subgraph`, `list-subgraphs`), `create-backup`, `materialize`, `set-write-policy`, `rebuild-schedule`, `backfill-plan-history`, `delete-taxonomy-block`. An exclusion list rather than an allowlist, so new ledger tools appear on this route by default.
- New `roboledger_router`, mounted at `/v1/mcp/roboledger` beside the agnostic router.

**Rate limits** (`config/rate_limits.py`, `middleware/rate_limits/subscription_rate_limits.py`)
- Both path matchers compared against `/v1/mcp` exactly; the new path is added so it takes the MCP bucket rather than falling to the generic limiter.

**Docs**: README MCP section; "two resources" docstrings in `resources.py`, `oauth_server/__init__.py`, `OAuthGrant` and `env.py`.

## Breaking Changes

None. Every OAuth and MCP transport endpoint touched here is excluded from the OpenAPI schema, so there is nothing for the SDKs to regenerate. `/v1/mcp` and `/v1/graphs/{graph_id}/mcp` keep their full tool lists.

## Testing

- `just test-all`: 14,788 passed, 42 skipped, 2 xfailed; lint, format and typecheck clean.
- New tests cover: resource resolution and discovery for the new URL; the consent gate (non-RoboLedger graph refused, access checked before product, RoboLedger graph granted, agnostic resource unaffected) and `graph_serves_product` (extension, repository flag and type, shared-repo and subgraph ids never reaching the DB, missing graph); cross-audience refusal in both directions and the 403 challenge on both agnostic routes; tools/list, initialize and tools/call filtering; a typo guard that every excluded name is a real tool on a RoboLedger graph at full flag surface; the route's 401 challenge, discovery document and pending payload through the app; rate-limit categorization.
- In-process smoke: unauthenticated `POST /v1/mcp/roboledger` returns 401 with `resource_metadata` naming the new document, and that document's `resource` is the new URL.
- Not yet run: a live OAuth walk against a deployed build (developer-mode connector on the new URL). The consent page's graph filter is a separate change in `robosystems-app`; until it ships, choosing a non-RoboLedger graph on this resource is refused by the server with the page's generic access message.

## Certification


jfrench9:
@claude please review this PR

PR #1417, opened 2026-09-17
fix(plaid): restate the bank's line, wait for the history, keep failed rows

jfrench9:
## Summary

Post-merge review of the Plaid bank feed (#1415) found five defects on the reconciling-item contract and the first-sync path. This PR fixes them: a restated bank line now posts what the bank currently says, a retracted line can only be reversed, a flag never carries the previous flag's entry forward, the first sync waits for Plaid's historical pull before the fiscal calendar opens, an empty first pull fails instead of reading as synced, and a row that fails to capture is retried on the next sync instead of being lost behind an advanced cursor.

## Changes

**Bank-feed handler** (`operations/event_block/python_handlers/bank_feed.py`)
- `BankFeedMetadata` types `source_amount`, `source_posted_date` and `source_removed`, the keys a feed writes when it flags a posted line.
- `dispatch` posts the bank's amount on the bank's date when those keys are present (a restate regenerates from the accepted metadata, which previously fell back to the event's own columns and re-posted the old line), refuses a retracted line, and refuses to re-run a tenant rule when the bank's amount or date changed. `dispatch_preview` refuses exactly the same cases, so preview and commit agree.

**Reconciling items** (`operations/roboledger/commands/reconciling_items.py`, API model description)
- Restate is blocked when the accepted payload carries no entry to restate to: the source retracted the line, or the line was posted through a rule the feed could not re-plan. The blocker text says to catch up.
- `default_disposition` is never a disposition the plan already knows is blocked.

**Plaid load** (`adapters/plaid/pipeline/load.py`)
- A new flag strips the previous flag's `source_*` keys and entry (`line_items`, `posting_date`, `memo`) before building the accepted payload. A resolution leaves them in the live metadata, and a removal flagged over them made the planner net against an entry the bank no longer sends.

**Plaid asset** (`adapters/plaid/pipeline/assets.py`, `client/api.py`)
- `TransactionsSync` exposes `pull_pending` and `history_complete`. While Plaid's pull is still running (`NOT_READY`, then `INITIAL_UPDATE_COMPLETE` with only the most recent ~30 days in), the Item's first sync waits up to 10 minutes; a later run, which already has a cursor, rechecks for at most a minute, so a scheduled sync never holds a worker long.
- A run that still has nothing fails, so the connection never stamps `last_sync` with zero data.
- The first run to see `HISTORICAL_UPDATE_COMPLETE` keeps pulling until a cursor returns nothing, and the credential bundle records `history_complete_at` so later runs never pay for it. Against the sandbox the flag flipped on a page of 16 transactions and 201 more answered the next cursor a beat later; without the drain the calendar would have opened on the 16.
- A run with a partial history captures it and stores the cursor, but the fiscal calendar bootstraps only once the history is complete. Before this, the calendar closed every month before the first 30 days and the closed-period gate then refused the history when it arrived.
- When any row fails to capture, the cursor is not advanced and the run fails naming the row; the next sync replays the window and the rows that did capture are found as existing. `/transactions/sync` never resends a window on its own.

**Tests**
- Handler: source overrides, retraction, rule-posted refusal, and the unchanged default path.
- Reconciling items (database-backed): a Plaid line restated at the bank's amount and date; a retracted line refused restate, defaulted to catch-up, and reversed.
- Plaid load: flags after a resolution carry no stale entry; a second change replans from scratch.
- Plaid asset: the wait loop and its short recheck on later runs, the empty first pull, the partial history, the once-only drain after the history flag, failed captures.
- Handler preview: the same refusals as dispatch.

## Breaking Changes

None. The reconciling-item API shape is unchanged; only the `default_disposition` description was reworded. No SDK tier is touched.

## Testing

`just test-all`: 14,940 passed, 42 skipped, 2 xfailed; ruff check, ruff format and basedpyright clean. The database-backed reconciling-item suite (19 cases, including the two new bank-line cases) was run locally against `robosystems_test`. The Plaid sandbox was probed directly for the status sequence on a new Item: `NOT_READY` at 0 s, `INITIAL_UPDATE_COMPLETE` at 4 s, `HISTORICAL_UPDATE_COMPLETE` at 7 s with 16 rows, then 201 rows on the next cursor.

## Certification


jfrench9:
@claude please review this PR

jfrench9:
Both review points addressed in 379e5366 and the follow-up commit: preview now refuses exactly what dispatch refuses (retracted line, rule-posted line with a changed bank record), and the historical wait keeps its ten-minute bound on the first sync only — a run that already holds a cursor rechecks for at most a minute. Also added since the review: the first run to see HISTORICAL_UPDATE_COMPLETE drains further cursors until one returns nothing and stamps history_complete_at, because against the sandbox the flag flipped on a page of 16 rows and 201 more answered the next cursor.

PR #1418, opened 2026-09-17
fix(plaid): re-key a replaced Item's history; pin the default window at connect

jfrench9:
## Summary

Findings 4 and 7 from the Plaid post-merge review. Plaid's account and transaction ids are Item-scoped, so a re-Link (a dead Item replaced in place, or a reconnect after a disconnect) brought the same bank transactions back under new ids and a replay captured the whole history again beside the posted originals. A replay now re-keys the events the feed already holds instead. Separately, a connection created without `since_date` stored none, Link sent no `days_requested`, and Plaid pulled its default 90 days while the model promised 1 January of last year; the default is now pinned into the connection's sync config at connect, so Link and the sync agree.

## Changes

**Plaid load** (`adapters/plaid/pipeline/load.py`)
- `rekey_replaced_events`, run only on a replay (no cursor). Two shapes may be claimed: an event this same connection captured under its previous Item (the dead Item replaced in place), and an event with no Item at all (a disconnect purged its source keys). An event that still carries another connection's Item is another feed's line and is never touched, whatever else matches. Within that, an event no payload in the batch identifies is matched on the chart account, the posting date and amount the bank last reported (a resolved change counts), and the bank's description where both sides have one; pairs also on their `from` account. Two lines alike on everything but description never cross. The match takes the new identity keys; its classification, books and trail are untouched, and every old identity key (transaction, accounts, Item, connection, legs) is kept on `metadata.rekeyed_from`.
- The main loop then reconciles the re-keyed event as existing. `events_rekeyed` is reported.

**Plaid asset** (`adapters/plaid/pipeline/assets.py`)
- Passes `rekey_replaced=cursor is None` and reports the count.

**Plaid provider** (`operations/providers/plaid_provider.py`)
- `_sync_config` materializes the default window when none is given, so Link asks Plaid for the documented history and the date is pinned rather than drifting a year every January. A row from before the change falls back to the same default when Link is minted.

**Shared lane** (`adapters/bank_feed/window.py`, `sync.py`)
- `default_backfill_start` moves to a Dagster-free module so the provider can import it; `sync.py` re-exports it.

**Purge** (`operations/roboledger/commands/connections.py`)
- `rekeyed_from` joins the payload keys scrubbed on disconnect.

**Tests**
- Load: the replay re-keys and reconciles as existing; an incremental sync never re-keys; same-Item events, events a payload already identifies, and another connection's live line are never targets; descriptions decide between same-day same-amount lines and differing descriptions never cross; a purged line matches without a description; a posted line matches on the date the bank last reported; a captured line matches on its own columns; a pair re-keys its legs; nothing to match costs no query.
- Asset: the replay flag follows the cursor.
- Provider: the default window is pinned at create; a stored window and a missing one both set `days_requested`.

## Breaking Changes

None.

## Testing

`just test-all`: 14,956 passed, 42 skipped, 2 xfailed; ruff check, ruff format and basedpyright clean.

End to end against the Plaid sandbox on the local stack: connect, first sync (58 captured), classify and commit two lines, disconnect (the 58 unposted deleted, the two posted kept with their source keys scrubbed), reconnect as a new Item. The replay reported `events_rekeyed: 2`, `events_existing: 2`, `events_captured: 56`. The two posted lines now carry the new Item's transaction ids and the new connection, with the old ids on their trail, and no line was captured twice.

## Certification


jfrench9:
@claude please review this PR

jfrench9:
Both suggestions taken in the follow-up commit. (1) The cross-connection case now has a hard guard rather than relying on the description: a replay may claim an event this same connection captured under its previous Item, or an event with no Item at all after a purge; an event that still carries another connection's Item is never a candidate, whatever else matches. Test added for the live-other-connection line matching on account, date and amount. (2) The trail keeps every identity key the event held, old account ids included. On the question: yes, in the in-place replacement the old events still carry the current connection_id, and that is now exactly the signal that admits them, with the missing item_id admitting the purged ones. On (3), noted in the query's docstring; a replay is rare and the date index carries the scan, so no index migration for now.

PR #1419, opened 2026-09-17
fix(plaid): reconcile legs inside pairs, settle against a posted counterpart, release survivors

jfrench9:
## Summary

The transfer set from the Plaid post-merge review, findings 6, 8, 10, 11 and 12. A modified leg of a captured pair now reconciles the pair. A leg whose other side is already classified or posted settles against it instead of becoming a second live line for the same movement. A pair keeps each leg's own date, so a leg that outlives the pair keeps its posting period. A posted pair that loses one leg releases the survivor as its own line. Cash and check deposits and ATM withdrawals are no longer transfer candidates.

## Changes

**Transform** (`adapters/plaid/pipeline/transform.py`)
- `TRANSFER_IN_DEPOSIT` and `TRANSFER_OUT_WITHDRAWAL` are not transfer candidates: nothing the feed can see is their other side, and a customer's check deposit is revenue. They capture as ordinary bank transactions and never pair. Everything else transfer-shaped, and the card payment, is unchanged.
- A pair records `from_date` and `to_date`.

**Load** (`adapters/plaid/pipeline/load.py`)
- `reconcile_pairs`: a `modified` leg already inside a pair reaches the pair. While the legs still match, an unposted pair takes the new amount and dates in place and a posted one becomes a reconciling item whose stashed payload carries the transfer it should now be, with `source_legs` so the change is not raised again. When the legs no longer match, the pair dissolves: an unposted pair is deleted, a posted one is flagged with no entry so its catch-up reverses it and its `legs` are emptied; both legs come back as single-leg payloads through the main loop, where one may pair anew.
- `find_waiting_leg` now also sees classified and posted counterparts, a still-captured one first. `settle_with_counterpart` decides: merge while the other side is unposted and either unclassified or classified to this leg's bank account, since the merged pair posts the same entry; capture and void this leg, through the kernel's own transition, when the other side is posted to this leg's bank account, because the movement is booked and a live second line would book it again the moment someone classified it; otherwise capture this leg on its own naming its counterpart, for the operator.
- A posted pair that loses one leg is flagged as before, and now also releases the surviving leg as its own captured line on its own date. The pair's `legs` narrow to the retracted leg, in the stashed payload and on the row, so a later sync finds the survivor on its own id.
- `survivor_payload` uses the leg's own date where the pair kept it, the pair's date for pairs captured before it did.
- New report counts: `pairs_dissolved`, `legs_voided`.

**Tests**
- Transform: the candidate table, a deposit and a withdrawal that never pair, the per-leg dates on a pair.
- Load: the pair reconciliation in each state, the released survivor and its date, the counterpart settlement in each state, the candidate ranking.
- Database-backed (`tests/adapters/plaid/test_load_db.py`): a leg classified to the other bank account merges into a pair through the real kernel, and a leg posted to the other bank account voids the arriving leg through the real transition, with the reason on the row.

## Breaking Changes

None.

## Testing

Full suite left to CI: the local stack was mid-rebuild when this was opened. Locally, ruff check, ruff format and basedpyright are clean, and the Plaid suite passes with the two database-backed cases against `robosystems_test` (124 cases). The first CI run failed only on a fixture registration in the new database-backed module, fixed in the follow-up commit.

## Certification


jfrench9:
@claude please review this PR

jfrench9:
Thanks — nothing to change from this pass. On the observations: (1) the in-place annotation is deliberate and settle_with_counterpart is reached once per payload from the main loop only; (2) leaving bank_description alone on an in-place pair update mirrors what reconcile_existing does for singles (only a captured line refreshes its hint keys), so it is scope, not an oversight. On the question: a teller-mediated move between two accounts at different institutions cannot pair regardless, because pairing is scoped to one connection, and one between two accounts at the same institution arrives as an account transfer, not a deposit and a withdrawal. If a bank ever tags both sides that way they land as two ordinary lines for the operator, which is the smaller harm next to a customer's check deposit vanishing into a false transfer. CI is green on the follow-up commit; the first run failed only on the fixture registration in the new database-backed module.

PR #1420, opened 2026-09-17
fix(plaid): scrub nested payloads on purge, lock the callbacks' first sync, sweep the small findings

jfrench9:
## Summary

Findings 9 and 23 from the Plaid post-merge review, plus the cheap remainder of its P3 list. Disconnect now scrubs the feed's payload keys wherever they sit in an event's metadata, not only at the top level. The QuickBooks, Mercury and Plaid connect flows start their first sync through the same per-connection lock the sync endpoint takes. The sweep fixes categories and counterparty types that followed money direction instead of nature, the institution-name match, a same-window add-and-remove, agents outside the window, Plaid's older not-ready error, an unretried exchange, a newer window on a pending row, and the order of the credential store.

## Changes

**Purge** (`operations/roboledger/commands/connections.py`)
- `scrub_payload_keys` walks nested dicts and lists, so `drift_payload`, `reconciliation_history` and any trail lose the feed's keys too; a resolution after the purge can no longer copy them back into live metadata. The key set gains the newer identity keys: `source_legs`, `released_legs`, `source_removed_transaction_ids`, `from_date`, `to_date`.

**Connect flows** (`operations/connection_service.py`, `routers/graphs/connections/oauth.py`, `operations/providers/plaid_provider.py`)
- `dispatch_first_sync` wraps the locked dispatcher for the connect flows: the run id when a sync started, `None` when one was already running, anything else propagated. The QuickBooks and Mercury callbacks and Plaid's Link completion use it instead of the registry directly; the QuickBooks path targets the revived row where one was revived.
- Plaid's Link completion stores the credential first, turns the row connected second, and removes the replaced Item at Plaid only after both. A failed credential write leaves a pending row and the old Item intact.
- A second create for a Plaid row still waiting on Link carries a newer explicit `since_date` onto that row (`refresh_pending_window`).

**Transform** (`adapters/plaid/pipeline/transform.py`)
- `event_category` follows the line's nature: income is `sales`, an unattributed deposit or withdrawal is `treasury`, everything else is `purchase`, refunds included. A tenant rule keyed on `sales` no longer sees a vendor refund.
- A counterparty is a customer when the bank called any of its lines income, never by the sign of one window; a vendor first seen through a refund stays a vendor.
- The institution name matches whole words, so "US Bank" is not dropped from a "Business Checking" account.

**Load and asset** (`adapters/plaid/pipeline/load.py`, `assets.py`, `client/api.py`)
- A transaction in both `added` and `removed` of one window is never captured; agents are made only for lines inside the window.
- `PRODUCT_NOT_READY` on the wire waits like `NOT_READY` on the page.
- The public-token exchange is never retried on a transport error; a public token exchanges once.

**Tests**
- New: the locked first-sync entry in each outcome; nested scrub; nature-based categories and types; the whole-word institution match; a same-window retraction; agents inside the window; `PRODUCT_NOT_READY`; the unretried exchange; the pending-window refresh; the credential store before the Item removal.
- Updated: the QuickBooks, Mercury and Plaid callback tests stub the locked entry instead of the registry; two QuickBooks tests now stub the soft-deleted-realm lookup they had left to the mock database.

## Breaking Changes

None.

## Testing

Full suite left to CI: the local stack was rebuilt mid-session and its Postgres hit its client limit during the gate. Locally, ruff check, ruff format and basedpyright are clean, and the adapter, provider, connection-router and purge suites pass.

## Certification


jfrench9:
@claude please review this PR

PR #1423, opened 2026-09-17
fix(search): stable hybrid ranking and grouped filings in results

jfrench9:
## Summary

Two search-quality fixes found while testing semantic search against the SEC corpus.

1. **Hybrid ranking now holds still across page sizes.** Each hybrid sub-query's candidate pool followed the requested page (`k = size + offset`, and the BM25 leg's default depth). The min_max normalization runs over that pool, so the same document scored differently at `size` 3, 5 and 10, a smaller page was not a prefix of a larger one, and keyword-only matches tied at a flat 0.4 took slots in every page. A hybrid search with `offset > 0` also failed outright: OpenSearch 2.19 rejects a hybrid query that sets `from` without `pagination_depth`.
2. **A filer's successive filings of the same section can fold into one hit.** Cross-company searches returned the same disclosure from one filer's consecutive 10-Qs in consecutive slots, so a page of ten often held a handful of companies. The MCP `search-documents` tool now groups them; the REST request opts in.

## Changes

**Hybrid candidate pool** (`operations/search/client.py`)
- New `HYBRID_CANDIDATE_DEPTH = 100`, sent as the hybrid query's `pagination_depth` and as the KNN `k`, independent of `size` and `offset`. 100 is the k-NN default `ef_search`, so the KNN traversal does the same work as before.
- Worth a close look: this changes the ranking of every semantic search, not only paginated ones. Scores for strong semantic matches rise relative to title-only keyword ties, because normalization now runs over a 100-candidate pool instead of a pool the size of the page.

**Result grouping** (`operations/search/service.py`, `models/api/search.py`)
- `group_successive_filings` folds hits sharing `(entity_cik, section_id, part)`. The group sits where its best-ranked member ranked; its lead is that member, or the newest filing among members tied with it (identical boilerplate scores identically). `also_in_filings` on the lead counts the others.
- `part` is in the key so different parts of one long section stay separate.
- A grouped search over-fetches `(offset + size) × 4` hits from offset 0, capped at the candidate pool, then slices the page from the grouped list. Over-fetching leaves the ranking unchanged only because of change 1.
- An `entity` filter turns grouping off, since one filer's filing history is then the point. Hits without a filer (uploaded documents) are never folded.
- OpenSearch 2.19 accepts `collapse` on a hybrid query and does not apply it (checked on 2.19.5), which is why this runs in the service rather than in the query.

**API and MCP surface**
- `SearchRequest.group: bool = False`. The REST default leaves existing callers, including the app's offset-paginated search page, unchanged.
- `SearchHit.also_in_filings: int | None`.
- `search-documents` (MCP) sends `group=True` and its description explains the folding.
- The `entity` descriptions (MCP tool and `SearchRequest`) now say that only a CIK selects exactly one filer: a company name is a loose word match across filers, and so is a ticker that is also a word (`ON` also matches filers with "On" in their names). The matching itself is unchanged.
- `SearchResponse.total` and the `search-documents` description state that `total` is the raw match count before grouping (OpenSearch stops counting at 10,000), not the number of distinct results a grouped search can page through.

## Breaking Changes

None. The request and response each gain an optional field with a behavior-preserving default (a generated-tier SDK regen opportunity). Hybrid search with `offset > 0` goes from an error to working results. Semantic result ordering changes as described above.

## Testing

- `uv run pytest tests/operations/search tests/middleware/mcp/tools/test_search_tools.py tests/middleware/mcp/test_tool_description_contract.py tests/routers/graphs/test_search.py` — 307 passed (after the review follow-up commit).
- New tests pin the fixed pool and the offset case (client), and the grouping behavior: fold and count, tie-break to the newest, a better score leading over a newer filing, parts kept apart, uploaded documents never folded, the over-fetch window and its cap, the entity opt-out, the REST default, and `total` staying the raw count (service + MCP). Each was checked by breaking the code it guards and confirming a test fails.
- `just test-code` — clean.
- Against local OpenSearch 2.19.5 with throwaway indices (deleted afterwards):
  - The same documents and query through the old and new client: the old client's pages disagreed across sizes and `offset=3` returned a 400; the new client returned identical scores at every size and `offset=3` returned exactly items 3–5 of the size-10 list.
  - A grouped search through `SearchService` in keyword and semantic modes folded three identical filings from one filer into the newest with `also_in_filings: 2`, and returned all three with the entity filter set.
- `just test-all` (the full unit suite) was not run in-session.
- Not yet verified on the production corpus; the plan is to re-run repeat-heavy semantic queries through the MCP tool after deploy, and to watch search latency, since a grouped MCP search now fetches up to 100 hits with highlights.

## Certification


jfrench9:
@claude please review this PR

PR #1424, opened 2026-09-17
fix(reports): carry the sender's filing status onto a shared copy

jfrench9:
## What

A report shared into another graph landed there as `draft` regardless of where the sender was in the filing lifecycle. The snapshot `share_report` builds never carried `filing_status`, so the recipient's copy took the column default while `generation_status` was set explicitly.

Nothing downstream could correct it: `_assert_report_mutable_by` closes every lifecycle transition to a shared-in copy, so the status was permanent. A fund reading filed statements from a portfolio company saw them labelled **Draft** in the viewer, with no path to anything else.

Reproduced locally — sender `filing_status=filed` (`filed_at` stamped), recipient's copy `filing_status=draft`.

## Changes

- `share_report` carries `filing_status` and `filed_at` in the snapshot; `_share_to_target` sets both on the copy.
- `filed_by` deliberately stays behind — it is the sender's platform user id, which means nothing in the recipient's graph and is printed verbatim by the report viewer.
- The status is **copied, not forced**. Sharing from `draft` is legal (only `generation_status` is gated), and telling a draft from final statements is the whole point of surfacing the field.
- Deleting a shared-in copy is now exempt from the filed/archived immutability lock. That lock protects an *author's* audit trail; a received copy is a delivery, and without the exemption the carried status would shut the recipient's only per-report exit for exactly the reports most likely to be shared. Block-and-purge already deletes such copies in raw SQL without consulting the guard, so the two exits now agree.

## Knock-on, intended

A received report that arrives `filed` now counts as protected in `taxonomy_block/immutability.py`, so a recipient's library curation can no longer cascade-delete its fact sets. That reads as an improvement: the copy cannot be regenerated locally, and the recipient still has the explicit exits (delete the report, or block the sender).

## Tests

Four new DB-backed cases in `test_share_controls_db.py`, each verified to fail against the unfixed code:

- the copy reports the sender's lifecycle position, with `filed_at` carried and `filed_by` withheld
- a draft shared report still arrives as a draft
- a recipient admin can delete a filed copy
- a filed report the graph authored keeps its lock

Full unit suite green (10,678 passed). Two unrelated local failures are environmental — the dev Postgres runs `max_connections=20` with the stack holding 17, so DB-backed tests error on connections under `-n auto`; each passes in isolation.


jfrench9:
@claude please review this PR

PR #1427, opened 2026-09-17
feat(docs): retire /docs, keep the Swagger console

jfrench9:
## Summary

Splits this host's two documentation pages by what each is actually for.

`/` keeps **Swagger UI** in every environment. Its try-it panel is a tool, and it is the only way to run a call against a deployed API from a browser. It is marked `noindex`: a tool, not a document.

`/docs` **retires**. It served ReDoc — a read-only renderer of the same specification, which the per-operation pages at `robosystems.ai/docs/api` do far better, and which no crawler could ever read because it renders in the browser from a ~950 KB file. It now answers `301`, so every README, CONTRIBUTING file and outside backlink keeps working and its link equity moves to the published pages.

**Deploy after robosystems-app [#395](https://github.com/RoboFinSystems/robosystems-app/pull/395)**, which serves the pages `/docs` redirects to. Merging ahead of it is fine; deploying ahead of it would point `/docs` at a 404.

## Changes

**`main.py`**
- `/docs` answers `301` to `{ROBOSYSTEMS_URL}/docs/api`, on GET **and** HEAD — FastAPI does not imply HEAD from GET, and link checkers probe with it. `ROBOSYSTEMS_URL` is per-environment, so staging redirects to the staging app and local development to `localhost:3000`.
- `/` and the `/static` mount are unchanged and stay in every environment.
- `X-Robots-Tag: noindex` on `/`, `/static/*` and `/openapi.json`. Nothing on this origin is the reference any more, so none of it should compete with the published pages in an index.
- `csp_variant_for_path` drops its flag: the `docs` variant now applies to `/` and `/static` only. `/docs` joins the strict policy — a redirect needs nothing relaxed — and ReDoc's `worker-src blob:` is gone from the variant.
- `/openapi.json` is otherwise untouched and stays served: both SDK generators read it, as does the isolation harness.

**`robosystems/utils/docs_template.py`** — `generate_robosystems_redoc` deleted, now that nothing calls it. `generate_redoc_docs` stays: the Graph API microservice serves its own ReDoc from it and mounts the same vendored bundle, so neither the generator nor `static/vendor/redoc-2.5.3` is removed.

**Tests** — `tests/test_main_docs_cutover.py` covers the redirect (GET and HEAD), the console and the policy it needs, and the `noindex` set. `tests/test_main_csp.py` pins that `/docs` is now strict and that the blob worker is gone; its ReDoc supply-chain guard follows the generator to its surviving caller. `tests/test_main_security_baseline.py` no longer follows redirects, so its probes assert on the response this origin returns.

**Text** — `CLAUDE.md`, `README.md` (which now lists the console alongside the reference and the raw spec), and `bin/tools/tunnels.sh`, which printed a `/docs` URL that now redirects.

## Breaking Changes

None for any SDK: `/openapi.json` is byte-identical and no operation changed, so neither client tier is touched. Both HTML routes were `include_in_schema=False` and have never been in the spec or generated.

`/docs` stops serving HTML and starts redirecting. Every in-repo pointer is updated; outside links ride the 301.

The ALB target group checks `/v1/status`, so nothing here affects health checks.

## Deploy Notes

Ordering, as above: app first. No new environment variables, no infrastructure change, no migration.

One consequence to record rather than bury: **the relaxed `style-src` on this origin does not go away.** Swagger sets inline style attributes at runtime, so `/` and `/static/*` keep the `docs` variant, and the accepted residual against this origin stays accepted. That is the deliberate price of keeping the console, and the surface narrows — `/docs` moves to the strict policy and the blob `worker-src` is dropped. The next periodic pentest's target list should name both `robosystems.ai/docs/api` and this console.

## Testing

`just test-code` — clean.

Full unit suite: **14,971 passed** at `-n 2`.

Verified against a running instance rather than only the test client, which is what caught the HEAD gap: `/` serves Swagger with the policy it needs and no third-party origin; `/docs` returns `301` on GET and HEAD with the strict policy; `/`, `/static/*` and `/openapi.json` carry `noindex` while `/v1/status` does not; `security.txt` and `/v1/status` are unaffected.

At the default `-n auto` the suite errors in setup with `psycopg2.errors.OutOfMemory` across unrelated modules. That is the local dev Postgres, not this change: it runs `max_connections=20` and `shared_buffers=50` blocks against ~13 workers, and adding any test module that uses the shared `client` fixture exceeds it. CI's `postgres:14` service uses upstream defaults and passes the full suite — worth a separate look at the local compose settings.

## Certification

https://claude.ai/code/session_011KBrhQHzwrB39C4c4Vw9Pq


jfrench9:
@claude please review this PR

PR #1428, opened 2026-09-18
chore(api): drop the parentheticals from three operation summaries

jfrench9:
## Summary

Three operation summaries carried a parenthetical that explained rather than named. A summary is the API reference page's `<h1>` and its nav label, so it wants to be a name — and in all three cases the parenthetical was already the opening clause of the operation's own `description`, so nothing is lost by dropping it.

| Before | After | Its description already said |
|---|---|---|
| `Execute Event Block (publish to source-of-truth system)` | `Execute Event Block` | "publish the captured GL plan to the source-of-truth system **(QuickBooks)**" — and it names QuickBooks, which the summary did not |
| `Terminate Schedule Early (No Entry)` | `Terminate Schedule Early` | "End a schedule early at a month-end cutoff **without booking any entry**" |
| `Auto-Map Elements via AI (async)` | `Auto-Map Elements via AI` | "**Dispatches to the background worker** — returns a `pending` envelope immediately" |

The first was **55 characters, 16 more than any other operation in the spec**, and wrapped to two lines in the reference's nav. The longest summary is now 39.

`(async)` was inconsistent besides: `autoSelectOperator`, `executeSpecificOperator` and `ingestFile` are equally asynchronous, say so in their descriptions, and carry no marker.

## Changes

`robosystems/routers/extensions/roboledger/operations.py` — three `summary=` values. No description, behaviour, schema or error-map change.

## Breaking Changes

None.

**No URL changes.** The reference's slug is derived from `operationId`, not the summary, so `/docs/api/extensions-roboledger/{execute-event-block,terminate-schedule,auto-map-elements}` are unchanged — no redirects, no sitemap churn.

**No client contract moves.** `summary` is descriptive metadata; it reaches the generated SDKs only as a docstring. Generated tier, cosmetic.

## Deploy Notes

The rendered reference reads the live `/openapi.json` and revalidates hourly, so this appears on `robosystems.ai/docs/api` about an hour after the API deploy, with no app deploy needed.

## Testing

`just test-code` — clean. `tests/routers/extensions` + `tests/middleware/test_extensions.py` — 195 pass. Nothing pinned these strings.

Verified against the regenerated spec from a running server rather than the source: no summary in the whole 200-operation spec carries a parenthetical any more, the three `operationId`s (and therefore the docs URLs) are unchanged, and the longest summary is 39 characters.

I did **not** render the docs page to confirm the nav no longer wraps — a `next dev` server is already running in the app repo and Next refuses a second one in the same directory, and I was not going to kill it or disturb its build output.

## Certification

https://claude.ai/code/session_011KBrhQHzwrB39C4c4Vw9Pq


jfrench9:
@claude please review this PR

PR #1430, opened 2026-09-18
feat(docs): RoboLedger guides for planning, mapping, schedules and the app

jfrench9:
## Summary

The RoboLedger product docs covered connecting, the close and reports, and said almost nothing about planning and forecasting, mapping, schedules or the app itself. This adds seven pages, deepens the existing eight, and adds screenshots. Two small publisher changes make that possible: the product lane can now ship images, and the sidebar groups pages by section.

Product docs publish on merge to `main`, so merging this puts the pages live on `roboledger.ai/docs`.

## Changes

**Publisher (`robosystems/scripts/publish_docs.py`)**

- A product page can show a file from its site's `images/` folder. The build rewrites `alt` to the content CDN and adds the file to the output, the same way the wiki lane already does. An image that does not exist, or sits outside `images/`, fails the build. A file no page shows is reported as a warning and not shipped. Dotfiles are skipped.
- A page's `section` front matter groups it in the sidebar. Sections appear in the order of their first page. A site that names no sections produces exactly the single untitled list it did before, so `docs/product/robosystems/` is unchanged. The apps already render titled sections for the technical docs, so nothing changes there.
- `build_product` takes `asset_base` as an optional fourth argument, defaulting to the public CDN, so the existing caller in `tests/scripts/test_product_docs.py` is untouched.
- The publish workflow already syncs non-markdown files, so it needs no change.

**New pages (`docs/product/roboledger/`)**

- `plan-and-forecast.md`: scenarios, the four drivers, numbers set directly on a line, per-line growth, what carries forward, the three-statement calculation and its monthly checks, the Plan page, re-running after a close, and what forecasting does not do.
- `ask-about-your-books.md`: what Claude can read, the difference between live, closed-month and report figures, ratios, and how to ask so an answer can be checked.
- `compare-with-public-companies.md`: adding the SEC filings graph as a second connection, and making a comparison fair.
- `map-your-chart-of-accounts.md`: what the first sync maps, reviewing it, and why mapping comes before the close.
- `schedules.md`: what a schedule is, setting one up with Claude or in the app, ending one early, and how schedules carry into forecasts.
- `changes-after-sync.md`: transactions edited in QuickBooks after they synced, and the three ways to settle one.
- `the-roboledger-app.md`: a tour of the app by sidebar section.

**Existing pages**

- `month-end-close.md`, `reports-and-sharing.md` and `troubleshooting.md` go deeper (the app's side of the close, what happens after it, reading and regenerating a report, seven new troubleshooting entries).
[...]

PR #1431, opened 2026-09-18
fix(docs): fail the build on a product image with a rooted path

jfrench9:
## Summary

Follow-up to #1430, from its automated review. A product page image written with a rooted path, like ``, skipped the image check and shipped a link nothing serves. It now fails the build like any other image that isn't a file under the site's `images/` folder.

## Changes

- `robosystems/scripts/publish_docs.py`: `rewrite_product_links` still passes a site path through untouched when it's a link (`[pricing](/pricing)`), but no longer when it's an image. A rooted image falls through to the existing check and is reported as `image must be a file under images/ that exists`.
- `tests/scripts/test_publish_docs.py`: the missing-image test also covers a rooted image, and checks that a rooted link still passes.

None of the published pages use a rooted image path, so the published output is unchanged.

## Breaking Changes

None.

## Testing

- `uv run pytest tests/scripts/test_publish_docs.py tests/scripts/test_product_docs.py`: 36 passed.
- `just docs-build`: 54 pages, 63 files, no warnings or errors.
- The pre-commit hook ran ruff, format and basedpyright clean.
- The full `just test-all` unit suite was not run.

## Certification


jfrench9:
@claude please review this PR

PR #1433, opened 2026-09-18
chore(infra): widen the shared replica Spot pool below on-demand price

jfrench9:
## Summary

Adds five instance types to the shared replica fleet's Spot pool. The existing five alternatives sit in heavily reclaimed Spot pools, and the fleet has been seeing repeated interruptions and failed Spot launches (`UnfulfillableCapacity`). The on-demand base is set separately by the `SHARED_REPLICAS_OD_BASE_*` variables. This PR only widens the pool that Spot and scale-out capacity draw from.

## Changes

- `cloudformation/graph-ladybug-replicas.yaml`: changes to the `MixedInstancesPolicy` overrides.
  - Added `r8gd.large`, `i4g.large`, `r9g.large`, `r8gn.large` and `r6gd.large` after the existing five.
    - All five have 16 GB (the floor the admission-control memory guard needs), are Graviton and non-burstable, and have 2 vCPUs.
    - `r8gd` and `i4g` are in the Spot Instance Advisor's lowest interruption band (<5%).
    - `r9g` (Graviton5) is too new to have Advisor data and isn't offered in one of the fleet's zones.
    - `r8gn` (10–15%) and `r6gd` (>20%) sit in busier bands. They were added because they raise the one zone that still scored low at full fleet size.
  - `!Ref InstanceType` (r7g.large) stays first. On-demand capacity is allocated by list order, so the on-demand base keeps landing on r7g.large.
  - Pinned `OnDemandAllocationStrategy: prioritized`. It is already the default for an overrides list and the live group already uses it, so nothing changes today. The pin keeps the on-demand base on r7g.large if the group ever moves to attribute-based selection, where the default becomes `lowest-price`.
  - The comment now records the rule that decides membership: every entry's Spot price must stay below r7g.large on-demand.
    - `price-capacity-optimized` ranks pools by capacity first and price second. It will launch an expensive type whenever it judges that pool the deepest.
    - That rules out the c-family 2xlarge types, the xlarge types with large local disks (`m7gd`, `m8gd`, `im4gn`, `m6gd`), and `r7gd.large`, whose Spot price sits at its on-demand price in every zone. `t4g` stays out as burstable, and the 1-vCPU 16 GB types (`x2gd.medium`, `x8g.medium`) stay out because the replica is sized on 2 vCPUs.
- **Instance store is unused.** `ladybug-replica.sh` downloads the database to `/mnt/ladybug-data` on the root gp3 volume and never mounts a local disk, so instance-store size doesn't matter.
- **Deploy behavior to watch.** The running instance types stay in the list, so the group has no reason to replace them. If CloudFormation treats the overrides edit as a launch-template change, `AutoScalingRollingUpdate` replaces replicas one at a time with one kept in service (about 15–20 minutes of cold boot each).

## Breaking Changes

None. Infrastructure only; no API, GraphQL or SDK surface changes.

## Testing

- `just cf-lint graph-ladybug-replicas` (cfn-lint + `validate-template`): passed.
- The commit hook's checks passed.
- `just test-all` not run. The change is a CloudFormation template with no Python changes.
- Checked Spot support, zone offerings and current Spot prices for each new type (`describe-instance-types`, `describe-instance-type-offerings`, `describe-spot-price-history`). All five currently price between $0.042 and $0.098/hr.
- Spot placement scores (`get-spot-placement-scores`, single-zone, 6 instances):
  - Current five: 2, 4, 6, 8 and 9 across the fleet's zones.
  - With the first three added: 7 to 9 in every zone.
  - With all five added: 9 in every zone, at both 6 and 10 instances (the fleet maximum). At 10 instances, the first three alone left one zone at 4.
  - `r8gb.large`, `r9gd.large` and `m8g.xlarge` also pass the price rule but added no measurable score, so they're left out.
  - At 2 instances, both lists score 8–9 today. Earlier the same day the current list scored 3 and 5 in two zones, so scores move.
- `r9g.large` hasn't run on this stack yet. A launch failure would show as an unhealthy instance that the group replaces. The on-demand base (r7g.large) is unaffected.

## Certification


jfrench9:
@claude please review this PR

PR #1434, opened 2026-09-18
chore(ci): file prod container scans under main

jfrench9:
## Summary

Prod deploys from a version tag, so the Trivy container scans in `build.yml` were filed under `refs/tags/v*`. Code scanning only shows the default branch, so the container alerts on `main` never got updated by a normal release. They only updated on a one-off deploy from `main`.

That's why the `soupsieve` alerts are still open after v1.12.8 shipped the fix. v1.12.8's scan dropped both findings, but it was filed under `refs/tags/v1.12.8`, and the latest scan on `main` is still from the v1.12.7 one-off. This PR files prod's container scans under `main`, so those alerts describe what is deployed.

## Changes

- **`.github/workflows/build.yml`**
  - New optional input `sarif_ref` (default empty).
  - Both container-scan uploads (`container-scan-api`, `container-scan-lambda`) now pass `ref: ${{ inputs.sarif_ref || github.ref }}` and `sha: ${{ github.sha }}`. When `sarif_ref` is empty, this behaves as before.
- **`.github/workflows/prod.yml`**: passes `sarif_ref: refs/heads/main`.
- **Staging**: unchanged. It keeps its own ref, since feature branches are tested there and shouldn't update the alerts on `main`.
- **Dependency scan**: unchanged. `test-ci.yml` already runs `dependency-scan` on `main`.

For review: `upload-sarif` describes `sha` as the latest commit on the ref. On a tag deploy it will be the release commit, which is on `main` but not the latest commit there.

## Breaking Changes

None. CI only; the API and SDKs are unaffected.

## Testing

- `actionlint` on `build.yml`, `prod.yml` and `staging.yml`: clean.
- Pre-commit hooks (ruff, format, basedpyright): passed.
- `just test-all`: not run. The change is workflow-only.
- **Not yet checked end to end.** That needs the next prod deploy from a tag. After it, confirm:
  - the `container-scan-api` and `container-scan-lambda` scans are filed under `refs/heads/main` (`gh api repos/RoboFinSystems/robosystems/code-scanning/analyses`);
  - the two open `soupsieve` alerts close.

## Certification


jfrench9:
@claude please review this PR

PR #1435, opened 2026-09-18
chore(dev): pin local Postgres connection and memory settings

jfrench9:
## Summary

The local Postgres keeps running out of connections and shared memory during test runs, which has been forcing full teardowns of the dev stack. The cause is that the local data directory was initialized with initdb's minimum settings, `max_connections = 20` and `shared_buffers = 400kB`, instead of the Postgres 16 defaults. The running stack's idle pool holds 11 of those 20 connections, so any parallel test run hits `too many clients already` and `out of shared memory`. This PR pins the settings in `compose.yaml` so they no longer depend on what initdb picked.

## Changes

- **`compose.yaml`, `pg` service:**
  - adds `command: postgres -c max_connections=100 -c shared_buffers=128MB -c max_locks_per_transaction=256`
  - adds `shm_size: 256mb`
  - Command-line settings override `postgresql.conf`, so an existing data directory picks them up when the container is recreated. No wipe is needed.
  - `shared_buffers` stays well inside the service's existing 1g memory limit.
- Local development only. Nothing here reaches staging or prod, which run RDS.

## Breaking Changes

None

## Testing

- `docker compose --profile robosystems config -q`: the file validates.
- Ran a disposable `postgres:16` container with the same command, `--shm-size=256m` and `--memory=1g`:
  - `pg_settings` reports `max_connections=100`, `shared_buffers=16384×8kB` (128MB) and `max_locks_per_transaction=256`
  - `/dev/shm` is 256M
- The existing `robosystems-pg` container has not been recreated in-session.
- `just test-code` passed in the pre-commit hook. The unit tests were not rerun, because nothing in them changed.

## Certification


jfrench9:
@claude please review this PR

PR #1436, opened 2026-09-18
feat(operators): self-hosted OpenAI-compatible model provider, off by default

jfrench9:
## Summary

Adds an optional second model provider, so the open-source stack can run the AI operators on open weights through any OpenAI-compatible Chat Completions server (Ollama, LM Studio, vLLM, NVIDIA NIM), with no proprietary inference service. Bedrock stays the platform path. The new provider is **off by default**: hosted deployments leave it off, and a deployment turns it on only when it wants its own model.

## Changes

**Registry and gating** (`config/operators.py`, `config/env.py`, `config/billing/ai.py`)
- `OPENAI_COMPAT_ENABLED` turns the provider on. It is read from the env var first, then the `features/` SSM flag, and defaults to off.
- When off, the `openai-compat` registry row and its `openai_compat` rate-card key are never created. Nothing can resolve to the model, and the meter refuses it.
- When on, `OPENAI_COMPAT_BASE_URL`, `OPENAI_COMPAT_MODEL` and `OPENAI_COMPAT_API_KEY` are read through `get_secret_value`: the env var first, then the deployment secret in staging/prod. When off, none of the three is read at all.
- The config fails at import in three cases:
  - the base URL or model is missing while the provider is on;
  - `OPENAI_COMPAT_MODEL` collides with a registered short name, a profile name or a Bedrock wire id, because resolution would then run and bill it as a different model;
  - an `OPERATOR_PROFILE_*` value is unknown, or names the self-hosted model while the provider is off.
- `ModelSpec` gains a `provider` field (`bedrock` | `openai_compat`).
- `OPERATOR_PROFILE_ECONOMY` / `_BALANCED` / `_QUALITY` point a tier at any registered model for one deployment. This is deployment-scoped and never set by customers. With all three unset, the platform mapping is unchanged.
- Self-hosted rates come from `OPENAI_COMPAT_CREDITS_PER_1K_INPUT` / `_OUTPUT` and default to 0: the rate card is a cost passthrough, and a self-hosted GPU has no per-token vendor cost. Cache reads and writes bill at the input rate, since no caching discount is assumed. Malformed, infinite or negative rates raise.
- `BedrockModel` is renamed `OperatorModel`, because the registry is no longer Bedrock-only. The enum is internal and not part of the API or the SDKs.

**Client** (`operations/operators/ai_client.py`, new `operations/operators/openai_compat.py`)
- `AIClient.create_message` dispatches on `spec.provider`. The Bedrock path is unchanged.
- Converse stays the canonical transcript, and `openai_compat.py` translates at the edge:
  - outbound, `toolUse` blocks become `tool_calls`, and each `toolResult` becomes a `tool` message, with error status written as an `Error:` prefix because Chat Completions has no error flag;
  - reasoning blocks and cache points are dropped;
  - inbound, the reply is turned back into Converse blocks;
  - any turn that carries tool calls is reported as `tool_use`, even when the server says `stop`;
  - unparseable tool arguments are kept under a marker key, so the tool rejects the call and the model can correct it;
  - inline `<think>` blocks are stripped;
  - `cached_tokens` is reported as cache reads.
- The tool loop, the operators and the meter are unchanged.
- The transport is `httpx`, which is already a dependency, with one client per call so no connection pool outlives its event loop.
- Every non-2xx response and every transport failure raises `AIProviderError`, the same operator-surfaced failure Bedrock refusals already use.

**Docs**
- `.env.example` has a commented self-hosted block.
- `config/README.md` has a new paragraph describing the provider.

Worth a close look: the import-time validation in `build_model_registry` / `build_profile_models`, and the user-turn ordering in `_user_messages`. Tool messages must come directly after the assistant turn that called them; the loop's wrap-up nudge follows as trailing text.

## Breaking Changes

None. There is no change to the REST, GraphQL or operations surface, so neither SDK tier is touched.

## Testing

- **New tests:**
  - `tests/operations/test_openai_compat.py` (29 tests): message translation, request building, response parsing, HTTP errors against `httpx.MockTransport`, provider dispatch, and a full `run_tool_loop` run on the self-hosted path in which a failed query is fed back as an error, the model corrects it, and the answer lands.
  - `tests/config/test_operators.py`: off-by-default gating, registry validation and tier overrides.
  - `tests/config/billing/test_ai.py`: rate validation.
- **Suites run:**
  - `tests/config`, `tests/operations/test_ai_client.py` and `tests/operations/operators` pass.
  - `just test-code` passes (ruff, format, basedpyright, cf-lint).
  - The full unit suite, run locally with `-n 2`, finished with failures only in eight database-backed modules this change does not touch. Their errors were connection exhaustion (`too many clients already`, `out of shared memory`) on the local Postgres, which #1435 addresses. All eight pass when rerun serially.
  - `just test-all` at default parallelism hits the same local Postgres limit and was not green locally. CI is the gate.
- **Manual check:** a dev process with no AWS credentials and every tier set to the self-hosted model sent the call to the configured endpoint and never reached Bedrock. The endpoint was unreachable, and the error surfaced as `AIProviderError`.
- **Not done yet:** a live run against a real open-weight model server, and the twelve-question operator eval on that model.

## Certification


jfrench9:
@claude please review this PR

jfrench9:
Review addressed in 088a359f:

- **Negative output cap:** taken. `OPENAI_COMPAT_MAX_OUTPUT_TOKENS < 0` now fails at import, like the other self-hosted settings, and has a test.
- **Bedrock client built outside `dev`:** scoped rather than changed. Every non-`dev` deployment runs on AWS anyway, because graph allocation needs S3 and the DynamoDB registries. Building the client eagerly and checking its credentials at startup is deliberate: it fails at construction rather than on the first billable call. `config/README.md` now says that only a `dev` stack runs with no AWS credentials.
- **Unused `cache_write` rate:** left as is. Every rate-card key carries all four dimensions, and a server that someday reports cache writes will be priced without a code change.

PR #1437, opened 2026-09-19
chore(dev): move local Postgres, Valkey, OpenSearch and LocalStack to named volumes

jfrench9:
## Summary

A local `just reset-local` could leave the stack unable to start. On macOS, the `./data/postgres` bind mount's emulated file ownership made the Postgres entrypoint's temporary server refuse the data directory (`data directory has wrong ownership`). The init script never ran, so the `robosystems` database was never created, and the daemon's platform migrations looped on `database "robosystems" does not exist`. The same failure is why `initdb` had fallen back to its minimum settings (20 connections, 400kB of shared buffers), which #1435 worked around by pinning the settings.

This PR moves the four stateful stores off bind mounts onto named Docker volumes, and adds a single command to wipe them.

## Changes

- **`compose.yaml`:** Postgres, Valkey, OpenSearch and LocalStack now use named volumes with fixed names: `robosystems-pg-data`, `robosystems-valkey-data`, `robosystems-opensearch-data` and `robosystems-localstack-data`. Fixed names keep the wipe command independent of the compose project name. LocalStack's persistence is a Pro feature, so on the community image its volume only holds LocalStack's own state. The pinned Postgres settings from #1435 stay.
- **`justfile`:**
  - new `just clean-volumes`: tears the stack down, removes the four volumes, and clears `.local/config.json`, whose API key belongs to the wiped database;
  - `clean-data`, which `reset-local` calls, removes the volumes in place of the four `rm -rf ./data/...` lines;
  - the volume list is one `_store_volumes` variable.
- LadybugDB, staging, artifacts and lance stay on bind mounts under `./data`; none of them showed the problem.
- **After pulling:** the old `./data/postgres`, `./data/valkey`, `./data/opensearch` and `./data/localstack` folders are no longer read and can be deleted.

## Breaking Changes

None. This is local development only.

## Testing

- `docker compose --profile robosystems config -q` validates, and `just --dry-run clean-volumes` resolves to the expected commands.
- **First boot on a throwaway named volume**, using the same image, command and init script: `initdb` chose `max_connections 100` and `shared_buffers 128MB`, there was no ownership error, and the init script created all five databases.
- **End to end on the dev stack, from this branch:**
  1. `just clean-volumes` with no volumes present exits 0.
  2. `just start` works: Postgres first-boot init ran, the daemon logged `Migrations completed successfully`, platform is at `39c73ea18599`, extensions at `0036`, all nine containers are healthy, and `/v1/status` returns 200.
  3. `just clean-volumes` against the live volumes removed all four.
  4. A second `just start` ran first-boot init again, migrated cleanly, and served 200.
  5. `just demo-user` recreated the user and `.local/config.json`.
- `just test-code` passed in the pre-commit hook. The unit tests were not run, because no Python changed.

## Certification


jfrench9:
@claude please review this PR

PR #1438, opened 2026-09-19
fix(mcp): state that ledger graph amounts are dollars, not cents

jfrench9:
## Summary

Graph ledger amounts are dollars, but nothing on the query surface said so, and `list-period-drafts` told every model that "line amounts are in cents" (true of its own output, since the OLTP ledger stores cents). Weaker models carried that rule into graph queries and reported figures a hundred times too small, fluently and with no error. This states the unit where ledger queries are written and scopes the cents note to the tool it belongs to. It also tidies the self-hosted model block in `.env.example`.

## Changes

- **`middleware/mcp/tools/constants.py`**: new `LEDGER_AMOUNT_GUIDANCE`. `LineItem.debit_amount` / `credit_amount`, `Transaction.amount` and `Event.amount` are dollars; never divide by 100. Only the close and schedule tools work in integer cents.
- **`cypher_tool.py`**: appended to the `read-graph-cypher` description after the anchor and status guidance, on tenant ledger graphs only, which is the same gate as those two, so the `sec` repository is unchanged.
- **`example_queries_tool.py`**: a third ledger info item ("⚠️ READ THIRD — ledger amounts are dollars") ahead of the example queries. The analyst operator prefetches this output into its prompt; MCP clients see it on `get-example-queries`.
- **`schedule_tools.py`**: the `list-period-drafts` note now says its cents apply to its own output and that graph queries return dollars.
- **`.env.example`**: `OPENAI_COMPAT_ENABLED` moves to the feature-flags section (it falls back to SSM `features/`), with a "needs … below" pointer from the connection block, the Mercury/Plaid pattern. The example model is now a tested configuration, and `OPENAI_COMPAT_TIMEOUT_SECONDS` is listed.
- **Tests**: the guidance is present on a ledger graph and absent on a non-ledger graph and on `sec`, for both tools.

## Breaking Changes

None. Tool descriptions and example-query text only; no API, schema or SDK surface changes.

## Testing

- **Before/after on the real operator loop.** Local stack, demo ledger, analyst operator, `standard` mode, gpt-oss-20b through the self-hosted OpenAI-compatible lane. The question was "which account carries the largest absolute net balance, and what is it?" (answer: Paid-in Capital, $2,835,000 credit). Only the served text varied between arms.
  - Before: 5 of 20 runs reported the balance as "2,835,000 cents (≈ $28,350)" or "$28,350.00".
  - After: 0 of 20; all 20 correct. Fisher exact, one-sided p ≈ 0.02.
- Verified that the restarted API serves the new text on all three surfaces before the after arm ran.
- `just test-code`: ruff, format and basedpyright clean (0 errors).
- `pytest tests/middleware/mcp tests/operations/operators tests/scripts`: 1,211 passed. The full `just test-all` was not run.

## Certification


jfrench9:
@claude please review this PR

PR #1440, opened 2026-09-19
fix(mcp): state that RoboInvestor position amounts are dollars in the graph

jfrench9:
## Summary

Follow-up to #1438, from its review. The RoboInvestor API and GraphQL return a position's `cost_basis` and `current_value` as integer **cents** (`PositionResponse`, with `cost_basis_dollars` / `current_value_dollars` beside them), while materialization divides by 100, so the graph's `Position` holds **dollars** under the same property names. A model that reads `get-graphql-schema` ("Cost basis in **cents**") and then queries `Position.cost_basis` in Cypher sees the same name in two units. This states the graph's unit where Cypher is written, as #1438 did for the ledger.

## Changes

- **`constants.py`**: new `INVESTOR_AMOUNT_GUIDANCE`. `Position.cost_basis` and `current_value` are dollars in the position's currency, while the API and GraphQL return the same names in cents. Never divide a graph amount by 100, and convert before comparing across surfaces.
- **`cypher_tool.py`**: appended to the `read-graph-cypher` description on tenant graphs that carry the `roboinvestor` extension. The shared-repository check the ledger gate used moves into `_is_tenant_graph()`, which both gates now call; there is no behavior change for the ledger gate.
- **`example_queries_tool.py`**: an `investor` info item wherever a `Position` table exists. The SEC manifest loads only `roboledger`, so no shared repository has one.
- **Tests**: investor graph, ledger-only graph, a graph with both extensions, and `sec` for the Cypher description; with and without `Position` for example queries.

## Breaking Changes

None. Tool descriptions and example-query text only; no API, schema or SDK surface changes.

## Testing

- Restarted the local stack and read the served text. The RoboInvestor demo graph (both extensions) carries the investor and ledger guidance in `read-graph-cypher` and the investor item in `get-example-queries`. The ledger-only demo graph carries the ledger guidance only.
- `just test-code`: clean. `pytest tests/middleware/mcp`: 954 passed.
- Not measured behaviorally. Unlike #1438 (a 5/20 → 0/20 before/after), no model has been observed making this error on investor data; the change is preventive, on the same pattern.

## Certification


jfrench9:
@claude please review this PR

PR #1444, opened 2026-09-19
fix(docs): quote the sec-filings description so GitHub parses its front matter

jfrench9:
## Summary

GitHub rendered `docs/product/robosystems/sec-filings.md` with `Error in user YAML: mapping values are not allowed in this context` in place of its front matter. The `description` value contains an unquoted `: `, which strict YAML reads as a new key. The docs build never noticed, because `split_front_matter` in `robosystems/scripts/publish_docs.py` splits each line at its first colon and accepts the value, so the published page on robosystems.ai was always correct. This quotes the value and adds a test so the next one fails CI rather than GitHub.

A strict-YAML sweep of every front-matter file under `docs/` found this page as the only failure (1 of 20). The wiki pages carry no front matter.

## Changes

- **`docs/product/robosystems/sec-filings.md`** — the `description` value is double-quoted. The build strips matching quotes, so the published description is byte-identical (checked by running `split_front_matter` on the file before and after).
- **`tests/scripts/test_product_docs.py`** — new `test_front_matter_is_strict_yaml`: loads each product page's front matter with `yaml.safe_load` and fails, naming the page, if it does not parse or is not a mapping. It reads the same files `build_product` does (`*/*.md`, skipping `README.md`) and asserts it found pages, so it cannot pass vacuously. The module docstring says why the build parser and GitHub disagree.

`publish_docs.py` is unchanged: it runs with `--no-project` in the publish workflow, so the strict check lives in the test rather than adding a PyYAML dependency to the build.

## Breaking Changes

None

## Testing

- Every test in `tests/scripts/test_product_docs.py` passes when called directly. The local Docker stack was down, and a pytest run of the file stops at the session's Postgres fixture before reaching these tests, which do not use the database. CI runs them under pytest.
- With the quotes removed from the page, `test_front_matter_is_strict_yaml` fails and the other four still pass. The guard catches exactly this case.
- `ruff check`, `ruff format --check` and `basedpyright` are clean on the test file, and the pre-commit hooks passed. `just test-all` was not run.

## Certification


jfrench9:
@claude please review this PR

PR #1446, opened 2026-09-19
fix(event-block): approving a recorded draft no longer writes a second entry

jfrench9:
## Summary

A journal entry, bill or payment created with its handler applied and `status: draft` writes its entry at creation and lands `classified`. The Inbox offers Approve on classified events, and approving re-ran the handler, writing a second draft that the next close would post beside the first. Approval now fires the handler only when no entry is linked to the event yet.

## Changes

- `operations/event_block/commands.py`: new `_has_linked_entries` counts `Entry` rows of any status by `triggered_by_event_id`; `update_event_block` fires the handler on a transition to `committed` only when there are none. The comment says why drafts must count, unlike `_assert_retractable`'s posted-only check. Docstring updated.
- Paths that still fire as before: captured QuickBooks items and bank lines have no entry yet; restate deletes the event's entries before it re-fires; the loader only re-fires captured events.
- `routers/extensions/roboledger/operations.py`: the `update-event-block` description says the handler does not re-run when it already wrote the entry at creation.
- `python_handlers/journal_entry_recorded.py` and the loader docstring: the missing-account error now says the account isn't in RoboLedger's chart of accounts yet and to sync the source, instead of "map these accounts in the CoA mapping". The error code (`element_unmapped`) is unchanged.
- Tests: `test_commit_after_classify_db.py` runs against a real tenant schema — approving a recorded draft leaves one entry, and approving a captured event still writes its entry. A mocked case in `test_commands_update.py` covers the skip.

For reviewers:
- A `metadata_patch` or `effective_at` sent together with the approval of such an event updates the event but not its existing draft. Before this change the same call double-posted.
- Approving a payment or bill recorded this way no longer retries its discharge link to the invoice or bill.
- Duplicates written before this fix are not removed; that needs a separate data audit.

## Breaking Changes

None. The `update-event-block` description text changes in the OpenAPI spec (generated tier); no schema or response shape changes.

## Testing

- The new database test fails on `main` (`assert 2 == 1`) and passes on this branch.
- `pytest tests/operations/event_block tests/operations/roboledger tests/operations/extensions`: 1542 passed.
- After the wording commits: `tests/routers/extensions`, `tests/middleware/mcp` and `tests/operations/event_block`: 1325 passed; `tests/operations/event_block` and `tests/operations/extensions`: 413 passed.
- `just test-code`: clean.
- Full `just test-all` not run.

## Certification


jfrench9:
@claude please review this PR

PR #1447, opened 2026-09-19
fix(graphql): let RoboInvestor graphs read their entities

jfrench9:
## Summary

The `entity` and `entities` GraphQL reads required the `roboledger` extension, so on a RoboInvestor-only graph Entity Info showed no fund and no graph ID, the header's entity selector was empty, and the issuer picker on the Portfolio page never listed a company that had shared a report. Both reads now accept either extension, as the report reads already do. Writes stay `roboledger`-only.

## Changes

- `graphql/resolvers/ledger.py`: `_ENTITY_EXTENSIONS` (the same pair as `_REPORT_EXTENSIONS`); `entity` and `entities` open their session through `_open_session_for_any`. An investor-only graph has the entities table (the tenant schema is provisioned with every table), its fund's entity row is written at creation, and a share adds the sender as a linked entity.
- Access checks are unchanged: this widens which extension may read the rows, not who may read the graph.
- `tests/graphql/extensions/test_ledger.py`: an investor-only graph reads `entity` and `entities(source: "linked")`; the test that investor-only graphs are refused ledger queries now uses `fiscalCalendar`, which stays ledger-only; the no-extension error message is updated.

## Breaking Changes

None. The GraphQL schema is unchanged; only the runtime extension gate on two reads is widened.

## Testing

- `pytest tests/graphql`: 115 passed.
- Pre-commit (ruff check, ruff format, basedpyright, cf-lint): clean.
- Full `just test-all` not run; not exercised against a running app.

## Certification


jfrench9:
@claude please review this PR

PR #1449, opened 2026-09-19
feat(sec): store the public filing artifacts gzipped at rest

jfrench9:
## Summary

The public filing artifacts — each filing's `holon.jsonld`, `tavi.json` and the document as filed — were stored uncompressed, and the CDN in front of them compresses neither `application/ld+json` nor anything over 10 MB. So every holon (the file the viewer opens) and the largest filers' files reached browsers raw. This PR stores those three representations gzipped, as `Content-Encoding: gzip` under the **same key, media type and URL**, teaches the one server-side reader that talks to the bucket directly to decode them, and adds a one-off job that rewrites the artifacts already in the bucket in place, without a reprocess. On a sample 10-Q holon the stored object goes from 5.35 MB to 331 KB (16×).

Three adjacent defects found in the same review ride along: text blocks were stored with no `Content-Type` (so they too were served uncompressed), the externalizer issued one HEAD per text block, and long narrative sections left their superseded unsplit object behind.

**Order matters on rollout:** the reader fix must be deployed before the job rewrites a single object. Without it the failure is silent rather than loud — see the reader note below.

## Changes

**Writer — `adapters/sec/processors/artifacts.py`**
- `gzip_artifact()` — `gzip.compress(data, compresslevel=6, mtime=0)`. No timestamp in the header, so the module's existing invariant holds: a reprocess rewrites the same bytes unless an emitter moved. The level is pinned because Python defaults to 9, which is ~5% smaller on a holon at half the speed. The backfill imports this same function, so the writer and the backfill cannot produce different bytes for one filing.
- The holon, the Tavi model and the filed document (`.htm` / `.html` / `.xml` / `.txt`; a PDF is left as it is) are uploaded gzipped. `_write_document` now reads the file and goes through `upload_bytes` instead of `upload_file`.
- **`manifest.json` and `tavi.gaps.json` stay plain**, deliberately: the catalog reads manifests with raw boto3, and a gzipped manifest would read as "no manifest" and quietly strip every filing's representations. The manifest's `bytes` remain the decoded sizes, so the catalog and its consumers see no change.

**Reader — `operations/aws/s3.py`**
- `S3Client.download_string` decodes a gzipped body, detected by magic bytes (`gunzip_if_gzipped`, now shared with `S3BackupAdapter._decompress_data`). boto3 returns the stored bytes regardless of `Content-Encoding`. Magic bytes rather than the header or a suffix, because the stored population is mixed while the backfill runs and the key never changes.
- This is the only server-side reader of these objects: `information-block` / `disclosures` on a shared repository load the published holon through it. Previously an undecodable body was swallowed into `None`, which that view reports as *"processed before its filing artifacts existed"* — a wrong answer, not an error. A new test reads a gzipped holon through the real `S3Client` to pin this.
- `upload_string` / `upload_bytes` / `upload_file` take `content_encoding` and `storage_class`; `batch_upload_strings` takes `cache_control` and `storage_class`. All optional; nothing is sent unless given.

**Storage class — every public-bucket writer**
- `PUBLIC_DATA_STORAGE_CLASS = "INTELLIGENT_TIERING"` (`config/storage/shared.py`) is named on the PUT by the artifact writer, the text-block externalizer, the narrative writer and the catalog, instead of leaving it to the bucket's day-0 lifecycle rule (a billed transition per object, and a stay in Standard until it runs). The rule stays as the backstop. The constant's comment records the invariant: Intelligent-Tiering's opt-in Archive tiers need an async restore the CDN cannot wait for and must never be enabled on this bucket.

**One-off job — `adapters/sec/pipeline/public_gzip_backfill.py` (`sec_public_gzip_backfill`)**
- Manually launched; run config `prefixes` (each a filing year or deeper, down to one filing folder — anything else is refused before a request is made), `mode`, `workers`, `dry_run`. 4 vCPU / 8 GB, Spot-preferred; a retry after a reclaim skips what is done. It builds its own boto3 client with a connection pool sized to the workers.
- The bucket is unversioned, so every write is final. Worth a close read:
  - `is_target` selects **by inclusion** — the two fixed names, or a document extension on a file directly in a filing folder that is not a `fact_*` / `narrative_*`. The manifest can never match.
  - `compress` — skips an object already gzip (without reading the body) and never compresses gzip bytes twice; decodes the compressed bytes and compares them to the original before replacing anything; PUTs with the headers it read plus `IfMatch` on the ETag it read, so an object the nightly rewrote in between is left alone.
  - `restore` — the exact inverse, and the only undo there is.
  - `sweep_narratives` — the one mode that deletes, kept separate so it has its own dry run and report. `find_orphan_narratives` removes an unsplit `narrative_{section}.txt` only when all three hold: the name is not itself a `_part{n}`; its `_part1` sibling is in the **same folder listing**; and it is **older** than that part 1. A part is never a candidate, and an unsplit object newer than stale parts is the live one and is kept.
- One failed object does not stop the rest; any failure fails the run at the end, and a re-run is idempotent.

**Text blocks — `adapters/sec/processors/textblock.py`**
- Blocks are uploaded as `text/html; charset=utf-8` / `text/plain; charset=utf-8` with a `Cache-Control`. They were uploaded with no `Content-Type`, which S3 serves as `binary/octet-stream` — not a type the CDN compresses. Only newly written blocks change; existing ones are content-addressed and skipped on reprocess.
- `_check_s3_object_exists` lists the filing's folder once and answers every block from that set, replacing a HEAD per block. A listing that fails falls back to a HEAD per key — it must never read as an empty folder, or every block would upload again.

**Narratives — `adapters/sec/pipeline/text_index.py`**
- `_narrative_keys` names the object a section part is written to and, for part 1 of a split section, the unsplit object it supersedes; the writer deletes that after a successful PUT. An accession is only re-indexed under `force_reindex`, which deletes its index documents first, so nothing references the deleted object. The reverse case (stale parts after a section shrinks) is not handled.

**Docs** — `adapters/sec/pipeline/README.md` describes gzip at rest, the boto3 caveat, and the job.

## Breaking Changes

None to the API, the GraphQL schema, the operations envelope or either SDK — no route, model or response shape changes.

On the public CDN the URLs, keys and media types are unchanged. A browser, `fetch`, `requests`, `httpx` and Arelle's web cache all decode `Content-Encoding` on their own. A client that does not — bare `curl`, `wget`, `urllib` — receives gzip bytes for these three files once they are rewritten (`curl --compressed`). Files under 10 MB that the CDN used to serve as brotli are served as gzip.

## Testing

`just test-all` — **15,202 passed**, 42 skipped, 2 xfailed; ruff, format and basedpyright clean.

New and updated tests cover: byte-identical output across two writes; manifest sizes equal to decoded lengths; the manifest and gaps sidecar staying plain; `download_string` on plain and gzipped bodies; a gzipped holon read end to end through the real `S3Client`; the `is_target` matrix; header preservation and `IfMatch` on rewrite; skip-when-done, the 412 path and dry run; a compress → restore round trip back to the original bytes; the orphan-narrative rule's full matrix and the sweep's dry run; prefix validation; text-block content types; one listing per folder with the HEAD fallback.

**Not run:** the job against a real bucket. It is exercised here only through an in-memory fake of the S3 client. Its first real run is a single filing folder after deploy — `compress`, verify, `restore`, `compress` again — before any wider prefix, and the narrative sweep runs last, after a dry run.

## Certification


jfrench9:
@claude please review this PR

PR #1450, opened 2026-09-19
fix(information-block): bound what the published-filing reader holds

jfrench9:
## Summary

Bounds the memory of the reader behind `disclosures` and `information-block` on a shared repository. It cached a published filing's model with every text-block fragment already read into it, whatever the request asked for. It now caches the model as the holon carries it, and a response reads only the fragments it returns.

## Changes

**`operations/roboledger/views/information_blocks.py`**
- `_published_model` no longer reads the filing's fragments; the cached model keeps each text block as the pointer the holon carries. `MODEL_CACHE_VERSION` moves to `4`, so models cached in the old form are not reused.
- `_inline_text` reads the fragments behind a response's own text entries, then the response is built again with the text in place. `information-block` does this for the block it returns, `disclosures` for the family a `topic` names. The whole-report map shows counts and no text, so it reads none.
- `FRAGMENT_TEXT_BUDGET_CHARS` caps what one response reads into the model. A fragment past it stays a pointer, which the tools already report as `external` with a note on where to read it — the same degradation as a fragment that cannot be fetched.
- `HOLON_BUDGET_CHARS` refuses a holon larger than the reader holds, before it is parsed, as `ReportTooLargeError` with a message pointing at `search-documents` / `get-document-section` and xbrlkit. Sized above every holon in the corpus today (largest measured: 44.6 MB).

**Callers** — `ReportTooLargeError` is exported from the views package; the REST view maps it to `422`, and the MCP tools return it as the tool's `error` string like the reader's other domain errors.

Tenant graphs are unaffected: their model comes from the ledger and carries its text inline, so there is nothing to read per response.

## Breaking Changes

None. Response shapes are unchanged. One behaviour difference: a text block larger than the per-response budget is now reported as `external` rather than previewed.

## Testing

`just test-all` — **15,206 passed**, 42 skipped, 2 xfailed; ruff, format and basedpyright clean.

New tests: the map reads only the holon; a block reads only its own fragments; a fragment past the budget stays marked external; a holon past the budget is refused without being parsed; the cached model holds pointers rather than the text, and a response served from the cache still reads and previews its own text.

Also measured locally on the largest filing in the corpus through this code path: the load and the cache freeze together now peak at roughly a fifth of what they did.

## Certification


jfrench9:
@claude please review this PR

PR #1451, opened 2026-09-20
refactor(extensions): split the RoboLedger surface into per-stage tags

jfrench9:
## Summary

All 57 RoboLedger operations carried a single OpenAPI tag, and the published reference at `robosystems.ai/docs/api` renders one page per tag — so a quarter of the whole API arrived as one undifferentiated list, next to tags like `Graph Health` that hold a single operation each. This splits RoboLedger into eight tags along the stages a ledger actually lives through, and reorganises the operations module to match, because within a tag the reference lists operations in route-registration order.

The generated spec is identical to the previous one except for `tags`: same 181 paths, same 463 component schemas, zero differences in any other operation field.

## Changes

**Tags (`robosystems/config/openapi_tags.py`)**

One tag becomes eight, ordered as a ledger is used — set it up, teach it a taxonomy, author blocks, post events, close the month, build the report, send it, analyse it:

| Tag | Ops |
| --- | --- |
| `RoboLedger: Setup` | 4 |
| `RoboLedger: Taxonomy & Mapping` | 7 |
| `RoboLedger: Information Blocks` | 8 |
| `RoboLedger: Ledger & Events` | 12 |
| `RoboLedger: Fiscal Close` | 7 |
| `RoboLedger: Reports` | 5 |
| `RoboLedger: Report Distribution` | 9 |
| `RoboLedger: Analytical Views` | 5 |

`Extensions: RoboInvestor` → `RoboInvestor` and `Extensions: GraphQL` → `GraphQL`, both keeping one tag. The prefix is dropped because tag ordering already groups the extension surfaces together, and `Extensions: RoboLedger: Taxonomy & Mapping` wraps to three lines in the reference's sidebar.

**Operation modules (`routers/extensions/roboledger/`)**

`operations.py` (3,228 lines) becomes a package with one module per tag; the largest is now 757 lines. Nothing was rewritten — every line of the original was assigned to exactly one module and the split was verified by line accounting before it ran.

- `_common.py` — the context builder, dispatcher, schema-missing 404, write-role dependency, service singletons, and a `make_registrar(router, tag)` factory. `views.py` and `reads.py` now take `_dispatch`/`_ctx` from here instead of from the command module, so a `FACT_GRID_ENABLED`-only deployment no longer imports the whole ledger command surface to serve fact grids.
- `operations/__init__.py` — composes the seven modules and re-exports their handlers and request bodies, so `...roboledger.operations.<name>` resolves as before.

**Worth a close look, reviewers:**

1. `operations/__init__.py` copies its submodules' routes rather than calling `include_router`. As of FastAPI 0.113 an included router stays an `_IncludedRouter` marker until the app mounts it, and two structural tests walk `router.routes` — every hand-written write op carries the write-role gate, and every op writing materialised content marks its graph stale. Both would have gone quietly vacuous. Copying is equivalent here because nothing is layered on at that level; the comment says what to do if a prefix or dependency is ever added.
2. `operations/information_block/schedule.py` defers three `roboledger.commands.schedules` imports to call time. That cycle is pre-existing and mutual — `commands/schedules.py` reaches up into `information_block.rules.engine`, and the handler reaches back down — and the old single-module import order happened to resolve it from the safe side. Deferring is the idiom `commands/schedules.py` already uses for its own back-references.
3. `main.py` mounts the three roboledger routers in three blocks rather than two, so `RoboLedger: Analytical Views` leads with `build-fact-grid` instead of the one OLTP-backed read. The gating is unchanged: commands and `reads.py` on `ROBOLEDGER_ENABLED`, `views.py` on `FACT_GRID_ENABLED`.

**Not affected:** MCP tool publication. `OperationRegistrar._all_instances` already exists so tool adapters can enumerate specs wherever their module lives, which is exactly what splitting one registrar into seven relies on.

**Tests** — patch targets followed the symbols into their new modules; `test_authoring_flag.py` now collects specs through `OperationRegistrar.specs_for_extension("roboledger")` rather than off a single registrar; the schedule-handler tests patch the three commands where they are defined rather than where they are imported.

## Breaking Changes

**None in this repository's own surface.** The GraphQL schema, the operations envelope and every REST request/response shape are untouched — verified by diffing the generated spec before and after, which shows zero differences in any operation field other than `tags`. No endpoint moved, and no request or response schema changed.

**But it forces a client major, which an earlier revision of this description got wrong.** `openapi-python-client` names its module directories after the OpenAPI tag, so retagging relocates 63 generated operation modules in `robosystems-python-client`: `api/extensions_robo_ledger/` becomes eight `api/robo_ledger_*/` packages, and `api/extensions_robo_investor/` becomes `api/robo_investor/`. Three of the six emit operations named in the stable tier — `create_event_block`, `assert_metrics`, `create_taxonomy_block` — are imported by path in `robosystems-integration-template` and `robosystems-demo-integration`, so this reaches the frozen set and costs a major rather than riding a minor.

Both clients therefore publish as **2.0.0**:

- **robosystems-python-client** — RoboFinSystems/robosystems-python-client#216. Function names, signatures and models are unchanged; only the package a function is imported from. The hand-written facades are repointed, so `LedgerClient` / `InvestorClient` consumers need no change.
- **robosystems-typescript-client** — RoboFinSystems/robosystems-typescript-client#231. No surface change at all (identical 1,661 exported symbols; the diff is ordering). It takes the same major to keep the two clients on the version they have shared since 1.15.0.

Shipping 2.0.0 rather than 1.18.0 is what keeps the break from landing silently: anyone pinning `>=1,<2` never resolves into it. The consumer sweep — two integration repos' imports and pins, three frontends' `^2.0.0` — follows the publish.

## Testing

`just test-all` — **15,206 passed, 42 skipped, 2 xfailed**; ruff check, ruff format, basedpyright (0 errors, 0 warnings) and cf-lint all clean.

Beyond the suite, the refactor was verified against the generated OpenAPI document directly: the spec was captured before any edit, and the final spec compared operation-by-operation. Same 181 paths, same 463 schemas, and zero differences in any field except `tags` — 65 operations retagged, matching the table above row for row.

## Certification


jfrench9:
@claude please review this PR

PR #1454, opened 2026-09-20
refactor(auth): carve passkeys, MFA and SSO out of the Auth tag

jfrench9:
## Summary

Carves the passkey, MFA and SSO mechanisms out of the `Auth` tag. Auth was the second-largest tag at 29 operations and the published reference renders one page per tag, so it arrived as a single long list. Part of the same regrouping as #1451.

Unlike RoboLedger, **no code moved** — the auth router is already thirteen modules, so this is tag assignment alone.

## Changes

| Tag | n | Contents |
| --- | --- | --- |
| `Auth` | 15 | register, login, logout, me, refresh, email verify/resend, invitations, password ×5, captcha, providers |
| `Auth: Passkeys` | 7 | WebAuthn registration, passwordless login, re-auth, list, remove |
| `Auth: MFA` | 4 | assertion options, verify, status, recovery-code regeneration |
| `Auth: SSO` | 3 | sso-token, sso-exchange, sso-complete (plus OIDC where enabled, which is `include_in_schema=False`) |

**A carve-out, not a subdivision.** Auth methods are alternatives, not stages — someone asking "how do I authenticate" wants password, passkey and SSO visible together to choose between, which is the opposite of RoboLedger's eight sequential activities. So the sign-in narrative stays whole in `Auth` and only the self-contained mechanisms lift out. Password reset stays with sign-in deliberately: forgot/reset is part of the story a reader is already following.

**One implementation note worth a reviewer's eye.** The tag moved off the `/v1/auth` parent router (`routers/__init__.py`) onto each sub-router include. FastAPI *merges* parent and child tags, and both the reference and the SDK generator read `tags[0]` — so a tag left on the parent would have won the grouping and silently undone the split. The spec is asserted to carry exactly one tag per operation.

## Breaking Changes

None in this repository's surface — same 181 paths, same 463 component schemas, and zero differences in any operation field other than `tags`.

**The SDK regeneration is additive, and the tag names are what make it so.** `openapi-python-client` names `api/<tag>/` after the tag. Keeping the parent named `Auth` leaves `api/auth/` in place holding `login_user` and `register_user` — the only two symbols any consumer imports from that package, verified across every repo. Nothing imports passkeys, MFA or SSO from the SDK. So the regeneration adds `api/auth_passkeys/`, `api/auth_mfa/` and `api/auth_sso/` and moves nothing anyone uses.

That makes it a **client minor**, not a major: the three frontends sit on `^2.0.0` and pick it up with no `package.json` change, and no consumer repo needs an import touched. Renaming the parent instead — to `Auth: Session`, say — would have broken those two imports and owed another major.

Per the two-tier contract this is generated tier, so the relocated modules should be named in the client's release notes.

## Testing

`just test-all` — **15,154 passed, 42 skipped, 2 xfailed**; ruff check, ruff format, basedpyright (0 errors) and cf-lint all clean.

The spec was captured before the change and diffed operation-by-operation afterwards: identical path set, identical schemas, 14 operations retagged (exactly the 7 + 4 + 3 carved out), every operation carrying exactly one tag, and no tag used-but-undeclared or declared-but-unused.


jfrench9:
@claude please review this PR

PR #1455, opened 2026-09-20
feat(graphql): publish resolver docstrings as field descriptions

jfrench9:
## The gap

The extensions GraphQL surface served **66 query fields with zero descriptions**, while the 118 types they return carried **581 docstring blocks**. The index into the schema was blank; the shapes it pointed at were full. On the public API reference the endpoint rendered as **"Handle Http Post"** with no request body at all.

The prose already existed. 64 of the 65 resolvers had a docstring written as API copy — Strawberry only reads an explicit `description=` and ignores `__doc__`, and exactly one field in the codebase passed one. So 64 written sentences sat one keyword argument away from every public surface.

## What this does

**`_describe_from_docstrings`** (`graphql/schema.py`) copies each resolver's `__doc__` onto its GraphQL field description as the schema is composed. A field is documented the moment someone writes an ordinary docstring, so the gap cannot silently reopen; an explicit `description=` still wins where public wording should differ.

Query fields go **0/66 → 66/66**, on five surfaces from one change:

| Surface | How it gets there |
|---|---|
| GraphiQL | the schema, in development |
| Introspection | any client, credentials or not |
| `schema.graphql` | the Python SDK's checked-in snapshot |
| `get-graphql-schema` | the MCP tool an agent reads before querying |
| robosystems.ai/docs/graphql | the published reference (separate PR in robosystems-app) |

**Because that text is now public**, internal notes moved out of six docstrings into comments inside their functions — retired REST endpoints, app-internal call sites, a `types/library.py` pointer. `libraryStructure` gained the one missing docstring. `graphql/README.md` records the convention: a resolver docstring is public API copy.

**The API reference operations.** FastAPI names Strawberry's ASGI handlers after their methods and emits no schema for them, so the reference published `Handle Http Get` / `Handle Http Post` with no request body and a pointer to a repo file path. Both now carry a real summary, a request-body schema with three worked examples (fiscal calendar, paginated-with-variables, introspection), a response schema, and links to the docs site.

## Verification

- 66/66 descriptions confirmed through the live API, not just the host: `{ __type(name:"Query"){ fields { name description } } }`
- `tests/graphql` 115 passed; openapi/graphql/main selection 339 passed
- `just test-code` clean

## Coordination

**Client minor, not a major.** After this deploys, `just refresh-schema` in `robosystems-python-client`. Measured: **+311 lines, 0 deletions** — all description blocks. Additive, no change to symbol resolution, so it rides a client minor and belongs in that release's notes.

**Deploy this before the robosystems-app reference PR.** That page renders from live introspection, so shipping it first would publish 66 fields with no descriptions.

## Known gaps, not closed here

- **Argument descriptions are still empty** — Strawberry takes those from `strawberry.argument(description=...)`, which nothing passes yet.
- **Return shapes are 40% documented** — 393/972 type fields, with 56 object types entirely undocumented (`InformationBlock` among them). Those come from `Field(description=...)` on the Pydantic models, which also feed the REST reference's response schemas, so filling them improves both references from one edit.


jfrench9:
@claude please review this PR

PR #1456, opened 2026-09-20
feat(graphql): document all 141 query arguments

jfrench9:
Follow-up to #1455, before the release cut.

## The gap that was left

#1455 took query fields from 0/66 to 66/66 described. Their **arguments were still 0/141**, so every argument table in the reference rendered an em dash in the Description column — across the 53 fields that take arguments, which is exactly where a caller needs to get it right: which values a filter accepts, what a date bounds, whether null means "both".

## How they get documented

Extends `_describe_from_docstrings` to parse a Google-style `Args:` block out of the docstring, apply each entry to the matching argument **by its Python name** (so `agent_type` reaches the wire's `agentType`), and **strip the block from the field's own description** so it is not published twice — the reference renders arguments as their own table.

Argument prose therefore lives in the same place field prose already does, and a new argument is documented by writing an ordinary docstring:

```python
@strawberry.field
def agents(self, info, agent_type: str | None = None, limit: int | None = None) -> list[Agent]:
    """List counterparty agents with optional filters.

    Args:
      agent_type: Filter by counterparty role, e.g. `customer` or `vendor`.
    """
```

## `limit` and `offset` are shared, deliberately

They appear 31 times between them and mean the same thing every time: `resolvers/_common.py` validates every pair against one shared guard (1-1000, offset >= 0, `INVALID_PAGINATION`), and only the per-field default varies — which is resolved server-side and never sent. So `COMMON_ARGUMENT_DESCRIPTIONS` covers both, rather than 31 copies that would drift apart. An `Args:` entry still wins where a field needs to say something else.

That leaves 110 entries authored per field. Most were **relocated, not invented** — several docstrings already explained their arguments in prose (`eventBlocks`' status and source vocabularies, `informationBlock`'s scenario and series interaction, `libraryElements`' classification axes). Those sentences now sit in the argument table where they are actionable, and the prose keeps the conceptual part.

## Verification

Through the live endpoint, not just the host:

- **arguments 0/141 → 141/141**, camelCase mapping confirmed (`eventType`, `agentId`, `isReconcilingItem`)
- fields stay **66/66**
- **no `Args:` block leaks** into any field description
- `tests/graphql` 115 passed; `just test-code` clean

## What this still does not cover

**Return shapes are 40% documented** — 393/972 type fields, with 83 object types carrying at least one undescribed field and 579 missing in total. The most-visited are `EventBlock` (0/25), `ReportPackage` (0/22), `Agent` (0/20) and `InformationBlock` (0/19). Those come from `Field(description=...)` on the Pydantic models, which also feed the REST reference's response schemas — so that pass improves both references from one edit. Not in this PR.

## Coordination

Same as #1455: after this deploys, `just refresh-schema` in `robosystems-python-client`. Additive descriptions only, so a client **minor**. Worth doing once after both are live rather than twice.


jfrench9:
@claude please review this PR

jfrench9:
All three review findings fixed in dca91c1.

**1. A test now pins the invariant.** This was the one worth having. The PR claimed the gap "cannot silently reopen", but nothing enforced it — a resolver with no docstring compiles, serves, and ships blank, and the only check was me re-running an introspection query by hand. `tests/graphql/extensions/test_schema_descriptions.py` reads the schema the way a client does and fails when any query field or argument reaches the wire undescribed, or when an `Args:` block leaks into a field's prose.

I verified it bites rather than assuming it: removing `taxonomyBlock`'s docstring fails it by name —

```
AssertionError: Query fields reach clients with no description: ['taxonomyBlock'].
Give the resolver a docstring — `_describe_from_docstrings` publishes it.
```

**2. Docstrings now match the documented convention.** Correct catch — the README indents `Args:` entries one level, Google style, and all 50 resolvers wrote them flush. `inspect.getdoc` normalizes both so nothing behaved differently, but a contributor copying the example would have produced something that looked unlike everything around it. Re-indented all 50 to match the example rather than changing the example, since the indented form is the standard.

**3. Dropped the unreachable branch.** `_SECTION_HEADING`'s alternation only ever captures `Args` or `Arguments` into that group, so `"Arg"` and `"Argument"` could never match. Now `("Args", "Arguments")`.

Still **66/66 fields and 141/141 arguments** described, no `Args:` leakage.

On the note about not being able to run the suite: CI is green on this PR — `test / test` passed in 11m, along with CodeQL and the three Analyze jobs. Locally, `tests/graphql` is 119 passed (up from 115 with the four new ones) and `just test-code` is clean.

PR #1457, opened 2026-09-20
fix(docs): publish descriptions written for the wire they reach

jfrench9:
Found by a post-merge review of #1455/#1456 and #413: the descriptions those PRs published are read by people, and two things about them were invisible in Python and obvious on a rendered page.

## reStructuredText nothing renders

` ``foo`` ` is RST. Every surface that publishes this text — GraphiQL, the reference, the SDK snapshot, the `get-graphql-schema` MCP tool, Swagger — treats it as markdown, so it arrived as literal backticks around the word.

| Carrying it | Count |
| --- | --- |
| OpenAPI schemas | 93 |
| OpenAPI properties | 60 |
| GraphQL types | 34 |
| GraphQL type fields | 10 |

Visible on 25 published pages; `create-taxonomy-block` rendered 92 stray backticks, `report-download-url` 26. Three of the affected pages are live in production today.

Swept to single backticks **at source**, because it is wrong on REST and GraphQL alike.

## Python names on a camelCase wire

`auto_camel_case` renames `balance_type` to `balanceType`, but nothing renames the sentence pointing at it. Most of this prose is Pydantic class docstrings that also serve REST, where `balance_type` **is** the property name — so one sentence was correct on `/docs/api` and wrong on `/docs/extensions/graphql`.

36 descriptions named a field the GraphQL schema rejects. `Account` told readers to ask for `balance_type`, `parent_id` and `account_type`; `ReportBundleDownload` for `download_url` and `generation_count`.

15 of the 36 sit in text genuinely shared with REST, so rewriting the source would have broken the REST copy. `_camelize_field_references` instead translates on the surface that did the renaming, which keeps both true. A token is rewritten only when its camelCase form is a real field or argument **of the same type** and the snake_case form is not — so an event-type value like `invoice_issued` is left alone.

## The guard that missed it

`test_schema_descriptions.py` covered the 66 query fields — the index into the schema, not the shapes it points at. That is exactly why this read as clean. It now walks every type, field and argument, and the spec's own prose is checked the same way in `test_main_docs_cutover.py`.

Both new invariants were mutation-tested: disabling the pass fails the name test, reintroducing one ` `` ` span fails the markup test on each surface.

## Reviewing this

Every change outside `schema.py`, the README and the tests is the backtick form and nothing else — verified line by line against that rule, so the 52 model files need only a spot check.

- `just test-all`: **15,165 passed**, ruff + format + basedpyright clean
- REST output confirmed unchanged: `PositionBlock` still documents `cost_basis_dollars` and still serialises snake_case

Pairs with RoboFinSystems/robosystems-app#414, which renders the surviving markdown instead of printing its markers.


jfrench9:
@claude please review this PR

PR #1458, opened 2026-09-20
docs(env): point the self-hosted model block at its own wiki page

jfrench9:
## Summary

The `OPENAI_COMPAT_*` block in `.env.example` sent readers to a section of the Local Development wiki page. That section is now its own page — **Self-Hosted AI Models** — because the lane applies to any deployment someone runs, not only a local stack. This updates the pointer, and widens it from "tested models" to what the page actually carries: endpoints, tested models and caveats.

Comment-only; no variable, default or behavior changes.

## Changes

- `.env.example` — retarget the self-hosted-model comment at the Self-Hosted AI Models page.

The wiki side is a separate repository and lands with its own push; until then the pointer names a page that is not live yet.

## Breaking Changes

None. No configuration surface changes, so nothing reaches either client SDK tier.

## Testing

`just test-all` not run — the diff is two comment lines in `.env.example`, which no test reads. The pre-commit hooks ran on commit (ruff check and format: clean). The wiki build that produced the referenced page was verified separately with `just docs-build`: 66 pages, the new page indexed under Self-Hosting & Development, all internal links resolving.

## Certification


jfrench9:
@claude please review this PR

PR #1460, opened 2026-09-20
chore(graphql): derive the published bounds, and ratchet the doc gap

jfrench9:
The last two findings from the post-merge review. They want different treatment, so they get it.

## The pagination bounds were a claim nothing tied to the validator

`COMMON_ARGUMENT_DESCRIPTIONS` spelled out `"1-1000"` while `_MIN_LIMIT` / `_MAX_LIMIT` live in `resolvers/_common.py`. Accurate on the day — I checked all 16 fields with a `limit` call the shared guard — but it's public copy on **GraphiQL, the published reference, the SDK snapshot and the `get-graphql-schema` MCP tool**, and it would have gone on asserting itself on all 16 fields after someone edited the guard.

Interpolated now, so the published text cannot disagree with what is enforced. The wording is byte-identical today.

The test checks the numbers against the validator's **behaviour**, not its constants: it reads the bounds out of the published text, asserts both ends are accepted and one past each raises `INVALID_PAGINATION`, and asserts every paginated field publishes the same wording.

Mutation-verified the right way round:

| Mutation | Expected | Result |
| --- | --- | --- |
| `_MAX_LIMIT` 1000 → 500 | **passes** — the text follows the guard | ✅ passes |
| Text hardcoded to a bound the guard doesn't enforce | **fails** | ✅ fails |

## The 579 undocumented type fields are not a task

Measured before deciding: **83 types**, and the ten largest are only **31%** of the total (top 20 = 49%). There is no subset worth sweeping — it's a long tail.

Written in bulk, 579 descriptions come out as *"The id of the account"*. **A vacuous description is worse than a blank one**: it reads as done, so nobody returns to it, and it still ships to five surfaces including text a model reads as context. That's editorial work needing domain knowledge per type, not a mechanical pass.

So this ratchets rather than demands. The 83 types are listed as `UNDOCUMENTED_TYPES`; a type **not** on the list must be fully documented. Three guards, all mutation-verified:

| Guard | Fails when |
| --- | --- |
| A type not on the list is fully documented | a new or newly-blank type appears — the gap can't grow |
| No listed type is already documented | progress isn't recorded by deleting its line |
| No listed type is missing from the schema | a rename leaves a hole the ratchet would silently exempt |

The README explains the convention and, for whoever picks it up, **which edit each type needs: 66 of the 83 are Pydantic-derived, where `Field(description=...)` documents the REST schema at the same time**; the other 17 are hand-written Strawberry types.

## Verification

`just test-all`: **15,171 passed**, 42 skipped, ruff + ruff-format + basedpyright clean.

Worth noting: two of the three ratchet mutations initially "passed" because ruff had reformatted `frozenset({` across lines and my replacement string matched nothing. I re-ran them asserting the edit applied, and both then failed correctly — but they'd have been reported as verified on the first run.


jfrench9:
@claude please review this PR

PR #1461, opened 2026-09-21
fix(api): the spec description describes the API, not a page

jfrench9:
Reported by Joey against `robosystems.ai/docs/api`.

`static/description.md` is the spec's `info.description`, and it renders on **two** surfaces: the Swagger console at the API root, and the header of `robosystems.ai/docs/api`. It opened by describing the page it was on — true at the root, wrong on the docs site.

## What the docs page was saying

Verified on a cold render, three problems:

| | |
| --- | --- |
| *"This page is the **playground**: every operation below can be run against this deployment"* | The docs page runs nothing — there is no try-it panel on it. |
| *"The written reference lives on the docs site, linked under each surface"* | Told a reader already on the docs site to go to the docs site. |
| Duplication | The page's own header already says *"To run a call in the browser, `api.robosystems.ai` serves an interactive console"* — so the two blocks contradicted each other about what you were looking at. |

## The fix

The paragraph is **removed rather than reworded**. At the API root it described the console the reader was already looking at; on the docs page that is the header's job, and the header does it better because it names the host. Everything of substance — the two surfaces, auth, MCP, the clients — is untouched.

Cold render of `/docs/api` after the change: "playground" gone, `interactive console` appears exactly once (from the page, with the URL), and the Swagger root still serves the description.

## Guarded

This is the **third instance this release** of one text published on two surfaces and phrased for one — after the reStructuredText nothing renders (#1457) and the `snake_case` field names on a camelCase wire (#1457). So it gets the same treatment:

- a test asserting `info.description` makes no page-relative claim
- a companion asserting the rule can't be satisfied by **deleting** the content

Both mutation-verified, each with an assertion that the mutation actually applied.

## One thing left deliberately

The Platform block still links `→ API reference` → `/docs/api`, which is a self-link when read on that page. Removing it would cost a Swagger reader their only pointer to the written platform reference, which is the more valuable side of the trade. A link to the page you're on is untidy; sending someone nowhere is worse.

`just test-all`: **15,173 passed**, 42 skipped, ruff + ruff-format + basedpyright clean.


jfrench9:
@claude please review this PR

PR #1462, opened 2026-09-21
feat(ci): announce new URLs to IndexNow from a sitemap diff

jfrench9:
## Why

**Bing has crawled nothing under `robosystems.ai/docs`**, three days after the docs launched. Site Explorer's folder tree for the site holds `research`, `pages` and `blog` — no `docs` node — and the newest crawl date on any root page is 15 Sep. Its sitemap was last read 9/19 at 295 URLs against a live file of 375.

A sitemap tells an engine what exists and then waits for it to come looking. IndexNow is the other direction: one POST and Bing — and through it Copilot and DuckDuckGo — plus Yandex, Seznam, Naver and Yep know within minutes. Google ignores the protocol, so there the sitemap and Search Console stay the levers.

This was the one unbuilt item from `seo-foundations`.

## What this adds

A shared composite action, `.github/actions/indexnow-ping`, that consumers call with a host and a key.

**It diffs rather than announces everything.** The previous run's sitemap is kept in the Actions cache; only URLs that appeared since are submitted. Bing throttles repeated submission of unchanged URLs, so submitting the whole sitemap every run would cost us the channel. A missing snapshot — first run, or a cache eviction — falls back to the whole sitemap, which is the one case where that is right.

**It reads the sitemap rather than taking a URL list**, which is what makes it trigger-independent. robosystems.ai's pages come from three places: the app's own deploys, the guides and technical docs published from this repo to S3 and picked up by an hourly revalidate, and the API and GraphQL references that follow the live spec the same way. A deploy hook would miss two of the three.

**It verifies the key is served before submitting.** Without that, a key that was never deployed — or a rotation that changed the variable but not the file — leaves a workflow green for weeks while every submission silently fails validation. The snapshot only advances after a submission is accepted, so a failed run re-announces rather than drops.

## Verified

Run locally against the live `robosystems.ai/sitemap.xml`:

- Parses **375 URLs**, 0 malformed, including all 67 GraphQL reference pages.
- Diff against a snapshot missing the GraphQL pages returns exactly **66 new** — no more, no less.
- The request body validates: `host`, `key`, `keyLocation`, `urlList` as an array.
- `POST https://api.indexnow.org/indexnow` with that body returns **HTTP 202** (accepted, key validation pending), which is the documented response for a key not yet served. The request shape is accepted; it needs the key file live to validate, and that ships with the consumer PR.

Two bugs were caught by running it rather than reading it: `tr -d '[:space:]'` strips the newlines between URLs and collapsed all 375 onto one line, and `sed 's|</\?loc>||g'` uses a GNU-only quantifier that no-ops on BSD sed, leaving the tags in place. Both fixed; the parse is POSIX now.

`actions/cache` is SHA-pinned to v4.2.3 per the convention in this directory. Files live under `RUNNER_TEMP`, not `/tmp`, since self-hosted runners share a host.

## Consumers

robosystems-app is the first, in [robosystems-app#418](https://github.com/RoboFinSystems/robosystems-app/pull/418) — **merge this one first**, since that workflow references the action at `@main`. Porting to roboledger.ai, roboinvestor.ai, xbrlkit.com and harbinger.finance is then a key file plus a workflow per repo; each host needs its own key, since the key file proves ownership of the host serving it.


jfrench9:
@claude please review this PR

PR #1463, opened 2026-09-22
fix(query): bound read cost on shared repositories

jfrench9:
## Summary

A shared repository can now declare text columns it does not serve through Cypher string matching. Statements that apply `CONTAINS`, `STARTS WITH`, `ENDS WITH` or `=~` to a declared column are refused, and the refusal points the caller at the repository's text tools. The SEC repository declares two columns. User graphs are unaffected.

## Changes

**Predicate — `security/cypher_analyzer.py`**
- `find_guarded_string_match(query, guarded_properties)`: a pure function in the same family as `is_write_operation`. It runs on the masked query, so an operator inside a string literal or a comment never matches.
- A declared property counts when it appears in either operand of a string-match operator, including inside a function call (`lower(f.value) CONTAINS …`) and in the function spellings of the same operators. A statement that string-matches one property and only returns a declared one is not matched.
- Declarations are `Label.property` pairs. The variable's label is read from the node patterns, so the same property name on another label is left alone. When the label cannot be read from the statement, the property name decides, and the refusal says how to state the label.

**Declaration — `adapters/base.py`, `adapters/sec/manifest.py`**
- `SharedRepositoryManifest` gains `guarded_string_properties` (default empty) and `guarded_string_guidance`.
- The SEC manifest declares its two columns, and its `cypher_query_guidance` now states the rule where a client reads it before writing a query.

**Enforcement — `middleware/graph/statement_kernel.py`, `middleware/mcp/tools/cypher_tool.py`**
- The kernel applies the rule beside the existing shared-repository write check and raises `SharedRepositoryReadRefused` (HTTP 400). That covers `/query/cypher` and the remote MCP transport.
- `assert_read_only_cypher` takes an optional `graph_id` and applies the same rule, because the Operator path reaches the engine through the tool without the kernel. Worth a close look: this is the second home for the rule, sharing one function (`shared_repository_read_refusal`) with the kernel.
- The rule reads a per-repository declaration with no default, so a tenant graph never reaches the predicate.

**Telemetry — `routers/graphs/query/execute.py`, `routers/graphs/mcp/remote.py`**
- The two recording sites pass the refusal's own signal (`string_match_refused`), so it can be counted apart from other rejected statements.

## Breaking Changes

None. No GraphQL schema, operations envelope or REST shape changes. The only behaviour change is a new `400` for one statement shape on shared repositories. Nothing for the client SDKs to regenerate.

## Testing

- `just test-all`: the test stage passed (15,213 passed, 42 skipped, 2 xfailed). The lint stage then failed on one error in a test added by this PR (`RUF043`), which stopped the run before format and typecheck.
- After fixing it: `just test-code` clean (ruff check, ruff format, basedpyright 0 errors, cf-lint), and the touched test files re-run green (628 passed).
- New tests: the predicate's match, fallback and must-not-match cases; the kernel refusing on the shared repository and its subgraph while serving the identical statement on a tenant graph; a repository that declares nothing refusing nothing; the tool-layer rule; the remote transport recording the distinct signal; and every shipped example query for the shared repository passing the rule.
- Against the running local stack: `/query/cypher` on the shared repository answers `400` with the refusal; the remote MCP transport answers `200` with `isError` and the same text; the identical statement on a tenant graph is served (`200`); an ordinary read on the shared repository passes the rule.
- Not verifiable locally: the telemetry line itself, since audit logging is off in the dev environment. The signal is asserted at the recording site by a test instead.

## Certification


jfrench9:
@claude please review this PR

